PR #73, opened 2021-07-21
Update production-checklist.adoc

### LINKED ISSUE #72: Docs: Feedback for Production Checklist
Hi, I have some feedback about [this page](https://github.com/hazelcast/hz-docs/edit/v/5.0-beta-1/docs/modules/ROOT/pages/production-checklist.adoc)

https://docs.hazelcast.com/hazelcast/latest/production-checklist.html#hardware-recommendations

- `We suggest at least 8 CPU cores or equivalent per member...`

https://docs.hazelcast.com/hazelcast/latest/production-checklist.html#operating-system-recommendations

- Solaris Sparc should be Solaris SPARC in all cases (just like VMWare ESX below it is "ESX" everywhere). See [these](https://en.wikipedia.org/wiki/Oracle_Solaris) [two](https://en.wikipedia.org/wiki/SPARC) Wikipedia articles if confirmation is needed.

https://docs.hazelcast.com/hazelcast/latest/production-checklist.html#vmware-esx

- Add

  - `- Avoid sharing one Network Interface Card (NICs) between multiple virtual machine environments. A Hazelcast cluster is a distributed system and can be very network-intensive. Trying to share one physical NIC between multiple VMs may cause network-related performance problems.`

- Edit

  - `Be careful overcommitting CPU cores. Monitor CPU steal time metrics.`

  - `Do not move guests while Hazelcast is running - for ESX this means disabling vMotion. If you want to use vMotion (live migration), first stop the Hazelcast cluster then restart it after the migration completes.`

  - `Always enable verbose garbage collection (GC) logs in the Java Virtual Machine. When "Real" time is higher than "User" time, this may indicate virtualization issues. The JVM is not using the CPU to execute application code during garbage collection, and is probably waiting on input/output (I/O) operations.`

  - `Use pass-through hard disks/partitions; do not use image files.`

  - `If you want to use automatic snapshots, first stop the Hazelcast cluster then restart it after the snapshot.`

https://docs.hazelcast.com/hazelcast/latest/production-checklist.html#windows

- `The workaround for such cases...`

https://docs.hazelcast.com/hazelcast/latest/production-checklist.html#jvm-recommendations

- `In order to avoid long garbage collection (GC) pauses and latencies from the Java Virtual Machine (JVM), we recommend 16GB or less of maximum JVM heap. If [High-Density Memory]() is enabled, no more than 8GB of maximum JVM heap is recommended. Horizontal scaling of JVM memory is recommended over vertical scaling if you wish to exceed these numbers.`
(Also add a hyperlink on "High-Density Memory", linking to https://docs.hazelcast.com/hazelcast/5.0-beta-1/storage/high-density-memory.html#latest-banner or equivalent)

- Semicolons should be replaced with colons for `General recommendations`, `For Java9+`, and `For Java 8`.

https://docs.hazelcast.com/hazelcast/latest/production-checklist.html#jvm-recommendations

- `Total data size should be calculated based on the combination of primary data and backup data. For example, if you have configured your cluster with a backup count of 2, then total memory consumed is actually 3x larger than the primary data size (primary + backup + backup). Partition sizes of 50MB or less are recommended.`

https://docs.hazelcast.com/hazelcast/latest/production-checklist.html#jvm-recommendations

- Add
  - `The number of internal partitions a Hazelcast member uses can be [configured]().`
  - (Also add link on "configured" pointing to https://docs.hazelcast.com/hazelcast/latest/capacity-planning.html#partition-count)

- Edit
  - `An optimal partition count and size establish a balance between the number of partitions...`

  - `The partition count should be a prime number. This helps to minimize...`

  - `A partition count which is too low constrains the cluster. The count should be...`

  - `A partition size of 50MB or less typically ensures good performance. Larger clusters may be able to use up to 100MB partition sizes, but will likely also require larger JVM heap sizes to accomodate the increase in data flow.`

  - `The partition count cannot be eaily changed after a cluster is created, so if you have a large cluster be sure to test and set an optimum partition count prior to deployment. If you need to change th partition count after a cluster is already running, you will need to schedule a maintenance window to entirely bring the cluster down. If your cluster uses [Persistence]() or [CP Persistence]() features, those persistent files will need to be removed after the cluster is shut down, as they contain references to the previous partition count. Once all member configurations are updated, and any persistent data structure files are removed, the cluster can be safely restarted.`

- Remove "Large Cluster Configuration Requirements" section, as it doesn't add value and doesn't appear to directly correspond with any section from the Deployment & Operations Guide

seii:
Fixes #72 

(Please do check my formatting, as I haven't used AsciiDoc before and wasn't sure if the text had to be manually aligned or not.)


docs/modules/ROOT/pages/production-checklist.adoc
```diff
@@ -84,33 +81,23 @@ For Java 8;
...
-If you need to change the partition count after a cluster is running, you will need to
-schedule a maintenance window to update the partition count and restart the cluster.
-
-== Large Cluster Configuration Recommendations
-
-You may need to increase the partition count relative to the number of Hazelcast members in the cluster and
-the data amount on each partition.
\ No newline at end of file
+The partition count cannot be eaily changed after a cluster is created, so if you have a large cluster be sure to test and set an optimum partition count prior to deployment. If you need to change th partition count after a cluster is already running, you will need to schedule a maintenance window to entirely bring the cluster down. If your cluster uses xref:storage:persistence.adoc[Persistence] or xref:cp-subsystem:persistence.adoc[CP Persistence] features, those persistent files will need to be removed after the cluster is shut down, as they contain references to the previous partition count. Once all member configurations are updated, and any persistent data structure files are removed, the cluster can be safely restarted.
```
Serdaro, line 103: ... cannot be easily changed ...

... If your cluster uses the xref:storage:persistence.adoc[Persistence] or xref:cp-subsystem:persistence.adoc[CP Persistence] feature, those ....

Serdaro (commented):
Minor edits.

docs/modules/ROOT/pages/production-checklist.adoc
seii: Agreed, those should be fixed now.

Serdaro (approved):

Issue #78, opened 2021-07-23
Persistence section is under "Improving cluster performance"


Holmistr:
The section on Persistence (https://docs.hazelcast.com/hazelcast/5.0-beta-1/storage/persistence.html#latest-banner) doesn't really make sense to me under "Improving cluster performance". It actually doesn't have anything to do with performance, most probably it will even make it worse.

I see it somewhere where talking about setting up cluster for reliability.

Issue #35, opened 2021-07-13
Docs: Feedback for Persistence


cangencer:
Hi, I have some feedback about [this page](https://hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/persistence.html)

Persistence is in configuring for Performance section, but not mentioned in the fault tolerance section.

PR #92, opened 2021-08-05
Persistence updates

### LINKED ISSUE #82: Docs: Feedback for Persistence
Hi, I have some feedback about [this page](https://github.com/hazelcast/hz-docs/edit/v/5.0-beta-1/docs/modules/storage/pages/persistence.adoc)

We don't really tell them why we need Merkle trees for map persistence to the new user.  It just has a config description for Merkle Trees.  We should go into more depth as to why we need this.  Explain it.

JakeSCahill:
Fixes #82 #78 and #35


docs/modules/storage/pages/configuring-persistence.adoc
```diff
@@ -0,0 +1,242 @@
...
+
+* `FULL_RECOVERY_ONLY`: Starts the cluster only when all expected members
+are present and correct. Otherwise, it fails.
+* `PARTIAL_RECOVERY_MOST_RECENT`: Starts the cluster with the members which have most up-to-date partition table and successfully restored their data. All other members leave the cluster and force start themselves. If no member restores its data successfully, cluster start fails.
+* `PARTIAL_RECOVERY_MOST_COMPLETE`: Starts the cluster with the largest group of members which have the same partition table version and successfully restored their data. All other members leave the cluster and force start themselves. If no member restores its data successfully, cluster start fails.
+
+=== `persistence.auto-remove-stale-data`
+
+Enables automatic removal of stale persistence stores.
+When a member terminates or crashes and the cluster state is `ACTIVE`, the remaining
```
vbekiaris, line 88: With 5.0 persistence, a member that crashes while state is  `ACTIVE` will generally not consider its data as stale; it will reload data from disk and then sync data with other members of the cluster as needed. There are a few situations that might result in on-disk data being considered stale, but the conditions under which these occur are exceptional and probably not worth listing exhaustively here. So the option still applies, however the circumstances under which data are considered stale are far less probable to occur in 5.0.

docs/modules/storage/pages/configuring-persistence.adoc
```diff
@@ -0,0 +1,242 @@
...
+device when a write operation returns successful response to the caller.
+
+Default: `false`
+
+By default, data is persisted to the disk device
+eventually, instead of on every disk write. This generally provides a better performance.
+
+=== `merkle-tree`
+
+Configures the Merkle tree for faster synchronization of data in the persistence store with the rest of the cluster.
```
vbekiaris, line 125: While this option is not strictly related to persistence (`IMap` merkle trees are also used for WAN sync), it is strongly recommended to enable merkle tree for each persistent `IMap` / `Cache`.

JakeSCahill: Other than WAN sync, are there any other scenarios where map or cache Merkle trees are beneficial?

JakeSCahill: If we have Merkle trees for JCache, can we remove this note from the delta WAN sync page?

```
Currently, Delta WAN Synchronization is implemented only for Hazelcast IMap. It will be implemented also for ICache in the future releases.
```

docs/modules/storage/pages/configuring-persistence.adoc
JakeSCahill: If this option is enabled, do members perform checks for stale data, or does the member just automatically remove its persistence store on restart?

docs/modules/storage/pages/configuring-persistence.adoc
JakeSCahill: We list these prerequisites for using Merkle trees with WAN, do they also apply to Persistence?

- Members/clusters must have the same number of partitions.
- Members/clusters must use the same partitioning strategy.
- Members/clusters must have the Merkle tree structure enabled.

docs/modules/storage/pages/configuring-persistence.adoc
vbekiaris: Regardless of the value of autoRemoveStaleData option, first the rejoining member and cluster's master determine whether data on rejoining member are stale. If stale, then the option's configured value determines whether the rejoining member will automatically delete the data from disk or it will fail startup.

docs/modules/storage/pages/configuring-persistence.adoc
vbekiaris: > Other than WAN sync, are there any other scenarios where map or cache Merkle trees are beneficial?

No, just WAN sync and only for `IMap`. Even though we added Merkle tree support for `ICache` in 5.0, we still lack the integration with WAN sync. So to sum up, merkle trees, as of 5.0, are useful for (1) IMap wan sync and differential partition sync, (2) ICache differential partition sync. Just FYI (2) will be done in 5.0 GA, it's not part of beta 1 or 2.

vbekiaris: About prerequisites, they do apply. However since for persistence we are discussing in the context of a single cluster, it is expected that all members have the same configuration eg you cannot have different partition count configured for members in the same cluster.

Holmistr (commented):
I did a review of the whole persistence chapter and here's my feedback.  It's mostly related to the structure, not the content itself..

* Overview section (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/persistence.html)
    * I like that I can start with the overview. It gives a very brief overview of what I can do with the feature.
    * However, it doesn't give me pointer where to go next. I want to set it up in 3 minutes, how do I do that? I would expect a link to the chapter in the documentation that shows the *minimal* configuration example of how to get it running.
    * In the "Planned cluster-wide shutdown", sub-bullet "Rolling restart" - the rolling restart could be actually a link to the RU section to make it more linked together.

* Persistence Configuration Options (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/configuring-persistence.html)
    * Perhaps the name of the chapter can be just "Persistence configuration"
    * Every developer start with trying to set a feature up and than adds complexity. Therefore, I would like to see right on the top the *minimal but complete* example of how to get persistence up and running just by copy pasting (= without thinking). Thus, there should be code snippets for XML/YAML/Programmatic config with the necessary global persistence configuration and then an example map config with pesistence enabled. Get inspiration from WAN chapter quickstart: https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/wan/quick-start.html
    * After the quickart minimal configuration, the flow would be "if you want additional configuration, see "global persistence configuration" or "data ...". The separation of the configuration options for global persistence, data structure and job snapshot is brilliant. I would like each section to start with a trivial empty code snippet to note which section are we talking about. This is a common problem for me when I'm using some technologies that there are configuration options listed, but I don't know where to put them. So for example, for the global persistence configuration, the code snippet would look like:

```
<hazelcast>
...
<persistence enabled="true">
  <!-- insert global configuration options here -->
</persistence>
...
</hazelcast>
```
* continuing in the same chapter
    * I kind of don't like the formatting of listing all the options. Not sure what are better options though. Maybe a table like <Description, Default, Example> and below that more extensive description if needed. Please note, that I explicitly added "Example". I absolutely love when the documentation shows what the value looks like, it's guiding the developer. It solves another problem that I see - there is not differentiation between `persistence.enabled` and `persistence.base-dir` (and others), but the `enabled` is an attribute on the `<persistence>` element while the rest is child elements. That's a big difference and can be confusing.
    * `persistence.validation-timeout-seconds` and `persistence.data-load-timeout-seconds` don't have defaults documented.
    * Subsection "Data Structure Persistence Configuration", in the sentence "Use the following options to configure Persistence for map and JCache data structures.", the map and JCache could be links to the datastructures.
    * Same thing in "Job Snapshot Configuration", could be a link to the jobs documentation
    * "Persistence Configuration Examples" section - as mentioned above, minimal example on top of the page and additional examples with more details later.
 
* Storing Persistence Data on Intel® Optane™ DC Persistent Memory (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/persistence-on-intel.html)
    * This is a rather minor usecase, nobody is really using it as of now. So the section should be at the end in of the persistence section in the TOC.
    
* Encrypting Persisted Data at Rest (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/encryption-at-rest.html)
    * I suggest following the same structure as I suggested before - minimal config example, than going into details.
    * Same feedback regarding the structure as in the config section, otherwise LGTM
    
* Recovering Persisted Data after Single Member Failures (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/recover-single-member.html)
    * Firstly, the name of the chapter is "Recovering Persisted Data after Single Member Failures". To me that reads as steps that I have to perform when a member crashes to bring it back online. But with persistence, this is pretty much automatical and all the stuff is actually determined during the configuraiton, therefore before anything happens. Not sure now how to "fix" this, maybe it doesn't need a fix, just a thought.
    * The initial part is mixing a few topics together and there's no clear separation. After a first sentence which is an OK introduction, it jumps right into how it internally works without any title. I think this should be separated in "How it work" section. Then it continues directly into listing options of how to solve the situation when partition tables don't match which is fine (if it's in the "How it works" section). But then, we jump to a completely unrelated topic of having lots of data persisted. IMHO this should be somehow separated into sections. 
    * As mentioned above, after we jump from one topic to another, then the order of the sections is reverse - first we're talking about how to solve migration of lots of data and then go back to incosistent partition tables issue. The order should be consistent to keep the flow.
    * Subsection "Delaying Migrations" - I would appreciate a real world example, e.g. when the cluster is running in Kubernetes and the orchestration will bring up the machine back online soon or when doing a planned maitenance and you don't want the migrations to be triggered. The mention of "hazelcast.partition.rebalance.after.member.left.delay.seconds" is kind of hidden. Actually this is a very important configuration that, I'll ask Vassilis why it's not a top level configuration element. But, in case we keep it as system property, it should definitely be listed in the persistence configuration page.
    * Subsection "Synchronizing Persisted Data Faster" - I think a link to the persistent config chapter (and maybe a code snippet) is a better option than duplicating the description.
    * Subsection "Force start", last sentence - the REST API and cluster management scripts should be links to the respective parts of the documentation. Or actually, why aren't the code snippets that trigger force start directly on this page? Is there any reason why to separate them?
    
* Backing Up Persisted Data (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/backing-up-persistence.html)
    * Love the flow - explanation, minimal example. Kudos!
    * Subsection "Triggering a Backup Request" - Java API point should definitely show a specific example/code snippet. Regarding the REST API, I'm just raising this again to think whether the code snippet should be embedded or not (general concern for all REST API calls)
    * Monitoring and Interrupting - can it be done via a client? General question, CC @vbekiaris.
    
    
That's pretty much all I have.

Holmistr (commented):
Hi @JakeSCahill ! Nice round of edits, thanks for that. Definitely a step in the right direction. Let me provide another round of reviews:

* Page "Persistence Configuration Options"  (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/configuring-persistence.html)
    * Since the configuration options are headlines, the right menu TOC now lists them. I don't think it's actually convenient. When you as a developer search for some configuration, you usually skim through all the options to as you don't know exactly what you're looking for. In that sense, listing the config options one by one in the right TOC doesn't help at all. I would rather have a clean right TOC with "Global Persistence Configuration", "Data Structure Persistence Configuration", "Job Snapshot Configuration" and "Full Example of Persistence configuration".
    * I somehow find the page visually hard to read, I can't help it. Maybe it's because the options are headlines (bold) or something. Perhaps it feels to me "too spacy" so it's harder to quickly go skim through it. I don't know, maybe @pawelklasa  could help?
    * In the examples for `base-dir` and `backup-dir`, I suggest putting an example path, not just "persistence" and "backups" in the element to be kind of more illustrative that it's meant to be a path, e.g. `/path/to/persistence` or `/tmp/persistence` or something. 
    * It's a bit tricky to realize that the default for `backup-dir` is empty string. Firstly, I assumed it was a typo. Maybe writing `Default: '' (empty)` would help to clear that up.
    * I absolutely love the structure and ending it with a complete example :)
    
* Page "Encrypting Persisted Data at Rest", section "Enabling Encryption at Rest" (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/encryption-at-rest.html#enabling-encryption-at-rest) - I actually wanted to write that it's not documented anywhere which algorithm you can use. But then I found out that the list is actually a link! It's not obvious due to formatting, so that's a first feedback. Secondly, when I click on it, it takes me to the "Configuration Options" page. That's a weird workflow for me because that's how I got to the encrypt data at rest page in the first place (from https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/configuring-persistence.html#persistence-encryption-at-rest ). So I would say that all the configuration related to encryption at rest should be on one place on https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/encryption-at-rest.html and in the configuration options page (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/configuring-persistence.html) have only a single link like there is: https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/configuring-persistence.html#persistence-encryption-at-rest

* Page "Configuring How Clusters Handle Single-Member Failures" (https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/recover-single-member.html)
    * The separation and the description of ""How Clusters Handle Single-Member Failures" is briliant. I would just appreciate a link to the "Handling Invalid Partitions" section somewhere in there in the text. I know that it's right below it but my point is that it's not that obvious that in the "How Clusters ..." section we're describing what could possibly happen and that the "Handling Invalid Partitions" sections is the answer to "what to do when that happens". Basically to somehow connect the docs there.
    * The link to `auto-remove-stale-data` is not taking me to the specific configuration option.
    * The formatting of examples in https://deploy-preview-92--hardcore-allen-f5257d.netlify.app/hazelcast/5.0-snapshot/storage/recover-single-member.html#synchronizing-persisted-data-faster should follow the same layout as in the persistence configuration options.
    * The link to REST API to trigger the Force Start should take me directly to the command which does that. Not just a general REST API section. I want to copy paste, I don't want to think.
    
* Backing up persisted data - this point wasn't addressed:

> Subsection "Triggering a Backup Request" - Java API point should definitely show a specific example/code snippet. Regarding the REST API, I'm just raising this again to think whether the code snippet should be embedded or not (general concern for all REST API calls)

PR #101, opened 2021-08-10
Remove superfluous backslash

jerrinot:
Backslash is used to escape the dollar sign, but only
when it's outside of the code section. Otherwise the
backslash is rendered in a final page.

Example #1: This line should NOT have the backslash,
because it's inside the [source] section: https://github.com/hazelcast/hz-docs/blame/c34c3869468b2937c6a5c21f736dc5e818af9f36/docs/modules/cp-subsystem/pages/management.adoc#L129

Example #2: This should should have the backslash,
because it's outside of the [source] section thus
the dollar sign has to be escaped: https://github.com/hazelcast/hz-docs/blame/dfc12d120f2944f978d718a981a103aae8a9274f/docs/modules/clients/pages/rest.adoc#L506

### DIFF
--- a/docs/modules/cp-subsystem/pages/management.adoc
+++ b/docs/modules/cp-subsystem/pages/management.adoc
@@ -126,9 +126,9 @@ include::ROOT:example$/cp/CpSubsystemAPI.java[tag=cpgroup]
 .REST API
 [source,sh]
 ----
-> curl http://127.0.0.1:5701/hazelcast/rest/cp-subsystem/groups/$\{CPGROUP_NAME}
+> curl http://127.0.0.1:5701/hazelcast/rest/cp-subsystem/groups/${CPGROUP_NAME}
 OR
-> sh cp-subsystem.sh -o get-group --group $\{CPGROUP_NAME} --address 127.0.0.1 --port 5701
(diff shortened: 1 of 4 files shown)


docs/modules/management/pages/license-info.adoc
```diff
@@ -58,7 +58,7 @@ Its output is similar to the following:
 request through the REST API (if enabled; see the xref:management:rest-endpoint-groups.adoc[Using the REST Endpoint Groups section]) on the `/license` as shown below:
 
 ```
-curl --data "${CLUSTERNAME}&$\{PASSWORD}&${LICENSE}" http://localhost:5001/hazelcast/rest/license
```
jerrinot, line 61: this is not the `[source]` section, but it seems it's should not be there anyway. 
this is how it's currently rendered:
image

jerrinot: probably the triple backticks have the same effect as `source`
(yes, I know nothing about AsciiDoc)

Serdaro (approved):

jerrinot:
I rebased on the latest master. apparently the license file was removed from the repo

PR #122, opened 2021-08-18
Update the Compact serialization config documentation with declarative configuration samples

mdumandag:
Since we have added support for the declarative configuration via XML and YAML,
we have to update the documentation accordingly.

### DIFF
--- a/docs/modules/serialization/pages/compact-serialization.adoc
+++ b/docs/modules/serialization/pages/compact-serialization.adoc
@@ -62,7 +62,7 @@ extraction fails, Hazelcast throws an exception as before.
 Currently, Hazelcast supports extracting schemas out of the classes that have the following
 field types.
 
-* Primitive types: boolean`, `byte`, `short`, `char`, `integer`, `long`, `float`, and `double`.
+* Primitive types: `boolean`, `byte`, `short`, `char`, `integer`, `long`, `float`, and `double`.
 * `String`
 * `java.time.LocalDate`, `java.time.LocalTime`, `java.time.LocalDateTime`, and `java.time.OffsetDateTime`
 * `java.math.BigDecimal`
@@ -150,7 +150,9 @@ class EmployeeSerializer implements CompactSerializer<Employee> {
 ----
 
 The last step is to register the serializer to the `CompactSerializationConfig`.
-Below is the programmatic configuration for this step.
+Below is the programmatic and declarative configurations for this step.
+
+**Programmatic Configuration:**
 
 [source,java]
 ----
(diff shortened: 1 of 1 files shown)


sancar (approved):

docs/modules/serialization/pages/compact-serialization.adoc
```diff
@@ -276,23 +355,101 @@ serializationConfig.
...
+hazelcast:
+  serialization:
+    compact-serialization:
+      enabled: true
+      registered-classes:
+        - class: com.example.Bar
 ----
+====
 
 You may need to register a class for reflective serializer if
```
ihsandemir, line 448: do you mean a case where user class is for example implementing `IdentifiedDataSerializable` but still want 't serialized as compact? If this is the aim, we should say that we do not recommend such uses. Also, then you should explicitly tell this case, and should also tell that it should not be registered to other factory classes, etc.

mdumandag: @sancar wrote that part so he may answer the best https://github.com/hazelcast/hz-docs/pull/120
but, as far as I see, it is just saying that if you have a class that is meant to be serialized with other mechanisms, and you want to use compact for it, you can register a reflective serializer and it will take a priority over it

sancar: @ihsandemir I have added this document specifically for Serializable but to make it complete I have added all the other methods. I haven't thought about what it would mean for our methods. I agree, we should make it clearer as you suggested.

mdumandag: @ihsandemir @sancar See https://github.com/hazelcast/hz-docs/pull/122/commits/8471e2036014fe2cd50249c72d274a2b05febbe9

docs/modules/serialization/pages/compact-serialization.adoc
```diff
@@ -276,24 +355,99 @@ serializationConfig.
...
+        - class: com.example.Bar
+----
+====
 
-* the class implements one of the existing serialization methods such as Serializable,
-Externalizable, Portable, IdentifiedDataSerializable or DataSerializable
-* the class is registered to be used by a custom serializer
-* there is a global serializer registered to serialize every class. 
+If you want to override the serialization mechanism used for `Serializable` or
+`Externalizable` classes and use Compact serialization; you can register
```
ihsandemir, line 449:
```suggestion
`Externalizable` classes and use Compact serialization; you must register
```

mdumandag:
@JakeSCahill @Serdaro Guys, I can this is ready for your reviews

ihsandemir (approved):

Serdaro (approved):

PR #139, opened 2021-08-26
Migration Guide Works

Serdaro:
* Polishing the existing content
* Adding the migration guide for IMDG 3.12.x releases.


docs/modules/ROOT/nav.adoc
```diff
@@ -56,7 +56,7 @@ include::wan:partial$nav.adoc[]
 ** xref:osgi:design.adoc[]
 ** xref:osgi:using-osgi-service.adoc[]
 
-.Migrate to this Version
+.Upgrading From Previous Releases
```
JakeSCahill, line 59: I like the renaming of 'migrate' to 'upgrade' 😄 

I'm not sure we need such a long label. 'Upgrade' implies replacing with a new version.

```suggestion
.Upgrade
```

PR #206, opened 2021-09-22
Improve the Docker deployment docs

JakeSCahill:
For deployments, we should focus on production use cases where we do not recommend running multiple containers on a single host. This PR updates the Docker deployment doc to recommend ways to run a cluster using more than one Docker host.

### DIFF
--- a/docs/modules/deploy/pages/installing-using-docker.adoc
+++ b/docs/modules/deploy/pages/installing-using-docker.adoc
@@ -1,192 +1,135 @@
-= Docker
+= Deploying on Docker
+:description: In this deployment guide, you will learn how to form a cluster from Hazelcast member containers running on more than one Docker host in the same local area network (LAN).
+
 [[installing-using-docker]]
 
-This section shows you how to use Hazelcast with Docker. We explain
-some options and parameters used in the examples, but otherwise assume
-you are familiar with Docker in general. Visit the https://docs.docker.com/get-started/[official documentation^] for a basic introduction.
+{description}
 
-== Starting a Hazelcast Member and Cluster
+== Before you Begin
 
-Paste the following into your command line:
+Make sure you've already installed the official Docker image.
 
-[tabs] 
-==== 
-hazelcast:: 
-+ 
--- 
+You should be familiar with Docker in general. See the https://docs.docker.com/get-started/[official Docker documentation^] for a basic introduction.
 
-[source,shell]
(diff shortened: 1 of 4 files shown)


docs/modules/deploy/pages/installing-using-docker.adoc
```diff
@@ -1,192 +1,135 @@
...
 
-hazelcast-enterprise::
-+
---
-[source,shell]
-----
-docker run -e 5702:5701 HZ_LICENSE_KEY=<Your Enterprise License Key> hazelcast/hazelcast-enterprise:$HAZELCAST_VERSION
-----
---
+TIP: You'll need the public IP address of your Docker host. You can find it on this web page: link:https://www.whatismyip.com/[https://www.whatismyip.com].
```
kwart, line 48: This is only valid if the machine is directly connected to the internet. It'll probably not be the case in companies or home networks, where they usually have LANs with NAT.

docs/modules/deploy/pages/installing-using-docker.adoc
```diff
@@ -233,26 +176,35 @@ Now edit the file and apply it when starting Hazelcast:
 
 [source,bash]
 ----
-docker run -v "$(pwd)"/hazelcast.yaml:/opt/hazelcast/hazelcast.yaml hazelcast/hazelcast
+docker run \
+-v "$(pwd)"/hazelcast.yaml:/opt/hazelcast/hazelcast.yaml \
+-e "JAVA_OPTS=-Dhazelcast.config=/opt//hazelcast/hazelcast.yml" \
```
kwart, line 181:
```suggestion
-v "$(pwd)"/hazelcast.yml:/opt/hazelcast/hazelcast.yml \
-e HAZELCAST_CONFIG=hazelcast.yml \
```

kwart (commented):
One more thing. In the untouched part of the section `Custom Hazelcast Configuration File`, there is the following command:

```bash
docker run --rm hazelcast/hazelcast cat /opt/hazelcast/hazelcast.yaml > hazelcast.yaml
```

It's not correct in the Hazelcast 5 as the listed file doesn't exist in the image. The proper replacement could be the following:

```bash
docker run --rm hazelcast/hazelcast \
  cat /opt/hazelcast/config/examples/hazelcast-docker.yaml \
  > hazelcast.yaml
```

kwart:
We could also unify the approach to image names:
* either use always image name without the version `hazelcast/hazelcast`
* or with the version: `hazelcast/hazelcast:{full-version}`

PR #311, opened 2021-12-08
Restructure getting started

JakeSCahill:
This PR improves the structure of the getting started by pulling out content from the overview that may interest the developer and operator personas as described in this doc by @seii https://docs.google.com/document/d/15nWHNJlxEnJu6wqk6EETzpXqjtaxqJO7A3VeB49mUE8/edit#

It also includes new conditional processing for content that applies only to snapshot docs. These conditional `ifdef` blocks allow us to create new maintenance branches from the `main` branch without having to remove content that was specific to a snapshot. See https://docs.asciidoctor.org/asciidoc/latest/directives/ifdef-ifndef/

It also includes the addition of the `experimental` Asciidoc attribute to allow us to use UI macros. See https://docs.asciidoctor.org/asciidoc/latest/macros/ui-macros/

When this is merged, we will need to update the release pipeline to change the `snapshot` value to `false` when creating a new branch for a release. @olukas for visibility.


seii:
One quick thing I noticed - it appears that "Get Started > Hazelcast Overview > "Hazelcast Editions and Features" remains even though it's been pulled out to "Get Started > Hazelcast Editions and Licenses". Is that intentional?

JakeSCahill:
Seems like that file didn't save before the commit. Thanks. I'll remove those sections.

olukas:
I'll update release pipeline once we merge this. I created https://hazelcast.atlassian.net/browse/HZ-768 for it.

PR #315, opened 2021-12-10
Note added for multimap event support

emreyigit:
The generic event listener class `EntryListener ` supports many events but `MulltiMap` does not support all of them. To clarify that, related note is added.

### DIFF
--- a/docs/modules/events/pages/object-events.adoc
+++ b/docs/modules/events/pages/object-events.adoc
@@ -208,6 +208,8 @@ Its default value is `false`.
 You can listen to entry-based events in the MultiMap using `EntryListener`.
 The following is an example entry listener implementation for MultiMap.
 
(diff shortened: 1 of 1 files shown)


docs/modules/events/pages/object-events.adoc
```diff
@@ -208,6 +208,8 @@ Its default value is `false`.
 You can listen to entry-based events in the MultiMap using `EntryListener`.
 The following is an example entry listener implementation for MultiMap.
 
+NOTE:  MultiMap Events support only `entryAdded`, `entryUpdated` and `mapCleared` events.
```
Serdaro, line 211: May. be rewriting like the following?

The entry listener for MultiMap supports only .......

emreyigit: Sure.

PR #325, opened 2021-12-27
Update obsolete go client `Start a Local Cluster in Docker` example

nevzatseferoglu:
- Typo was fixed.
- Context and sample error handling mechanism were added.
- Obsolete implementation were creating a client for writing to members but did not close it. It even did not use the same client that it has already created it before in writing section and starts a new client in reading section again. Therefore I shutdown the client in both writing and reading sections of the guide. @yuce @utku-caglayan

### DIFF
--- a/docs/modules/getting-started/pages/get-started-docker.adoc
+++ b/docs/modules/getting-started/pages/get-started-docker.adoc
@@ -316,44 +316,44 @@ go get github.com/hazelcast/hazelcast-go-client
 package main
 
 import (
-	"context"
-	"fmt"
-	"log"
+  "context"
+  "fmt"
+  "log"
 
-	"github.com/hazelcast/hazelcast-go-client"
+  "github.com/hazelcast/hazelcast-go-client"
 )
 
 func main() {
-	ctx := context.TODO()
+  ctx := context.TODO()
 
-	cb := hazelcast.NewConfig()
-	cb.ClientName = "hello-world" <1>
+  cb := hazelcast.NewConfig()
+  cb.ClientName = "hello-world" <1>
 
-	hz, err := hazelcast.StartNewClientWithConfig(ctx, cb) <2>
-	if err != nil {
-		log.Fatal(err)
-	}
+  hz, err := hazelcast.StartNewClientWithConfig(ctx, cb) <2>
+  if err != nil {
+    log.Fatal(err)
+  }
 
-	mp, err := hz.GetMap(ctx, "my-distributed-map") <3>
-	if err != nil {
-		log.Fatal(err)
-	}
+  mp, err := hz.GetMap(ctx, "my-distributed-map") <3>
+  if err != nil {
+    log.Fatal(err)
+  }
 
-    <4>
-	_, err = mp.Put(ctx, "1", "John")
-	if err != nil {
-		log.Fatal(err)
-	}
+  <4>
+  _, err = mp.Put(ctx, "1", "John")
+  if err != nil {
+    log.Fatal(err)
+  }
 
-	_, err = mp.Put(ctx, "2", "Mary")
-	if err != nil {
-		log.Fatal(err)
-	}
+  _, err = mp.Put(ctx, "2", "Mary")
+  if err != nil {
+    log.Fatal(err)
+  }
 
-	_, err = mp.Put(ctx, "3", "Jane")
-	if err != nil {
-		log.Fatal(err)
-	}
+  _, err = mp.Put(ctx, "3", "Jane")
+  if err != nil {
+    log.Fatal(err)
+  }
 }
 
 ----
@@ -574,41 +574,40 @@ Go::
 package main
 
 import (
-	"context"
-	"fmt"
-	"log"
+  "context"
+  "fmt"
+  "log"
 
-	"github.com/hazelcast/hazelcast-go-client"
+  "github.com/hazelcast/hazelcast-go-client"
 )
 
 func main() {
-	ctx := context.TODO()
+  ctx := context.TODO()
 
-	cb := hazelcast.NewConfig()
-	cb.ClientName = "hello-world"
+  cb := hazelcast.NewConfig()
+  cb.ClientName = "hello-world"
 
-	hz, err := hazelcast.StartNewClientWithConfig(ctx, cb)
-	if err != nil {
-		log.Fatal(err)
-	}
+  hz, err := hazelcast.StartNewClientWithConfig(ctx, cb)
+  if err != nil {
+    log.Fatal(err)
+  }
 
-	defer hz.Shutdown(ctx) <2>
+  defer hz.Shutdown(ctx) <2>
 
-	mp, err := hz.GetMap(ctx, "my-distributed-map") <1>
-	if err != nil {
-		log.Fatal(err)
-	}
+  mp, err := hz.GetMap(ctx, "my-distributed-map") <1>
+  if err != nil {
+    log.Fatal(err)
+  }
 
-	entries, err := mp.GetEntrySet(ctx) 
-	if err != nil {
-		log.Fatal(err)
-	}
+  entries, err := mp.GetEntrySet(ctx) 
+  if err != nil {
+    log.Fatal(err)
+  }
 
-	for _, entry := range entries {
-		fmt.Printf("key: %s, value: %s\n", entry.Key, entry.Value)
-	}
+  for _, entry := range entries {
+    fmt.Printf("key: %s, value: %s\n", entry.Key, entry.Value)
+  }
 }
-
 
 ----
 <1> Request all data in the map and print it to the console.
--- a/docs/modules/getting-started/pages/get-started-docker.adoc
+++ b/docs/modules/getting-started/pages/get-started-docker.adoc
@@ -302,7 +302,7 @@ if __name__ == "__main__":
 Go::
 +
 --
-. Install the Python client library.
+. Install the Go client library.
 +
 [source,shell]
 ----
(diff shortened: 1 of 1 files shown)


docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -313,28 +313,69 @@ go get github.com/hazelcast/hazelcast-go-client
...
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/hazelcast/hazelcast-go-client"
+)
+
+func main() {
+	ShutdownProperly := func(ctx context.Context, hz *hazelcast.Client, err error) {
```
utku-caglayan, line 327: I believe we can make a "defer" call right after handling returned error values of "hazelcast.StartNewClientWithConfig". Such as:
`defer hz.Shutdown(ctx)`

If we do that, we can handle following errors simply by
`if err != nil {
log.Fatal(err)
}`

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -313,28 +313,69 @@ go get github.com/hazelcast/hazelcast-go-client
...
+			log.Fatal(fmt.Errorf("%v", err))
+		case err != nil && errSd != nil:
+			log.Fatal(fmt.Errorf("%v; %v", err, errSd))
+		}
+	}
 
-func mapSampleRun() {
-  cb := hazelcast.NewConfigBuilder()
+	cb := hazelcast.NewConfig()
+	cb.ClientName = "hello-world" <1>
```
utku-caglayan, line 338: I am not sure if naming client is part of the steps, but if it is not I suggest we stick to default configuration with `StartNewClient` API. It is simpler IMO. What do you think?

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -545,32 +586,59 @@ Go::
...
-  }
+	cb := hazelcast.NewConfig()
+	cb.ClientName = "hello-world"
+
+	ctx := context.TODO()
 
-  hz.Shutdown() <2>
+	hz, err := hazelcast.StartNewClientWithConfig(ctx, cb)
+	if err != nil {
+		log.Fatalln(err)
```
utku-caglayan, line 617: We can use `log.Fatal` here.

utku-caglayan (changes requested):
Thanks for updating these examples! I noted some suggestions

docs/modules/getting-started/pages/get-started-docker.adoc
nevzatseferoglu: It was the part of the example.

utku-caglayan: Okay then, ignore this please 🙏

docs/modules/getting-started/pages/get-started-docker.adoc
nevzatseferoglu: What if `hz.Shutdown(ctx)` return an error ?

utku-caglayan: Clients are generally shutdown at the end of the programs or some scope. Meaning that resources are cleared even if you do not act on the returned error. One use case I can think of is opening/closing multiple clients, which would create some leaks if error is not handled.

But generally handling error on shutdown is not critical for majority of cases as you can see from examples in go client. If you really want to handle it somehow I would suggest
```
defer func() {
        if err := client.Shutdown(ctx); err != nil {
            log.Println("could not shutdown client gracefully")	
        }
}()
```
or some variation of it. How does it sound?

nevzatseferoglu: Sounds good to me!

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -313,28 +313,60 @@ go get github.com/hazelcast/hazelcast-go-client
...
+	cb.ClientName = "hello-world" <1>
 
-  cb.Cluster().SetName("hello-world") <1>
+	hz, err := hazelcast.StartNewClientWithConfig(ctx, cb) <2>
+	if err != nil {
+		log.Fatalf("Client could not connect to the cluster, %v", err)
+	}
 
-  client, err := hazelcast.StartNewClientWithConfig(cb) <2>
+	defer func(ctx context.Context) { <5>
```
utku-caglayan, line 336: I think we can use the context from the outer scope. We do not need to pass it with this parameter. Wdyt?

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -313,28 +313,60 @@ go get github.com/hazelcast/hazelcast-go-client
...
+		if err := hz.Shutdown(ctx); err != nil {
+			log.Fatalf("Could not shutdown gracefully, %v", err)
+		}
+	}(ctx)
 
-	mp, _ := client.GetMap("my-distributed-map") <3>
+	mp, err := hz.GetMap(ctx, "my-distributed-map") <3>
+	if err != nil {
+		log.Fatalf("Could not create a map, %v", err)
```
utku-caglayan, line 344: We can just handle it  with `log.Fatal(err)`. It would be simpler

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -545,32 +577,59 @@ Go::
...
+	"fmt"
+	"log"
 
-  cb.Cluster().SetName("hello-world")
+	"github.com/hazelcast/hazelcast-go-client"
+)
 
-  client, err := hazelcast.StartNewClientWithConfig(cb) 
+func main() {
+	ShutdownProperly := func(ctx context.Context, hz *hazelcast.Client, err error) {
```
utku-caglayan, line 591: It would be great if we can refactor this as we did with the previous one.

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -545,32 +577,59 @@ Go::
...
-  }
+	cb := hazelcast.NewConfig()
+	cb.ClientName = "hello-world"
+
+	ctx := context.TODO()
 
-  hz.Shutdown() <2>
+	hz, err := hazelcast.StartNewClientWithConfig(ctx, cb)
+	if err != nil {
+		log.Fatalln(err)
```
utku-caglayan, line 608: We can use log.Fatal here

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -545,32 +577,59 @@ Go::
...
+
+  <1>
+	mp, err := hz.GetMap(ctx, "my-distributed-map")
+	if err != nil {
+		ShutdownProperly(ctx, hz, err)
+	}
+
+	entries, err := mp.GetEntrySet(ctx)
+	if err != nil {
+		ShutdownProperly(ctx, hz, err)
```
utku-caglayan, line 619: If we can just use log.Fatal(err) here, it would be more similar to the go-client examples

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -545,32 +577,59 @@ Go::
...
+	}
+
+	for key, value := range entries {
```
utku-caglayan, line 622: Variable "key" refers to the index on the "value" here. I think it would be great if we can use the actual key of the entry.
```suggestion
	for _, entry := range entries {
		fmt.Printf("key:%s, value:%s\n", entry.Key, entry.Value)
	}
```

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -545,32 +577,59 @@ Go::
...
+		fmt.Println(key, value)
+	}
+
+  <2>
+	err = hz.Shutdown(ctx)
+	if err != nil {
+		log.Fatalln(err)
```
utku-caglayan, line 629: We can use log.Fatal here to be coherent with examples

utku-caglayan (changes requested):
Thanks a lot for the updates! I provided some suggestions 🙏

docs/modules/getting-started/pages/get-started-docker.adoc
nevzatseferoglu: If I would not send as a parameter it will take main scope's ctx value as a reference I guess. We are also sending context as a parameter as a convention. Maybe, there are something which need to be copied while sending as a parameter.  That's why I sent context in this way. It can be bad practice. I'm not sure.

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -313,23 +313,49 @@ go get github.com/hazelcast/hazelcast-go-client
...
-  cb := hazelcast.NewConfigBuilder()
+  "github.com/hazelcast/hazelcast-go-client"
+)
 
-  cb.Cluster().SetName("hello-world") <1>
+func main() {
+  ctx := context.TODO()
 
-  client, err := hazelcast.StartNewClientWithConfig(cb) <2>
+  cb := hazelcast.NewConfig()
```
yuce, line 329: Can you use "hazelcast.Config{}` instead of `hazelcast.NewConfig()`. And also, can you rename `cb` to something else, maybe `config` or `cfg`?

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -313,23 +313,49 @@ go get github.com/hazelcast/hazelcast-go-client
...
+  cb.ClientName = "hello-world" <1>
```
yuce, line 330: Java client sample seems to set the cluster name. Why do we set the client name here? Maybe the default cluster name  can be used: `dev`.

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -313,23 +313,49 @@ go get github.com/hazelcast/hazelcast-go-client
...
 
-	mp, _ := client.GetMap("my-distributed-map") <3>
+  hz, err := hazelcast.StartNewClientWithConfig(ctx, cb) <2>
+  if err != nil {
+    log.Fatal(err)
```
yuce, line 334: Instead of `log.Fatal`, let's `panic`, with something like: `panic(fmt.Errorf("starting the client: %w", err))`. Similarly for other `log.Fatal` statements.

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -313,23 +313,49 @@ go get github.com/hazelcast/hazelcast-go-client
...
+  mp, err := hz.GetMap(ctx, "my-distributed-map") <3>
+  if err != nil {
+    log.Fatal(err)
+  }
 
   <4>
-	mp.Put("1", "John")
-	mp.Put("2", "Mary")
-	mp.Put("3", "Jane")
+  _, err = mp.Put(ctx, "1", "John")
```
yuce, line 343: It's a bit confusing to use numbers as strings. How about using numbers directly or some other string, like: `"person1"`.

yuce (changes requested):

docs/modules/getting-started/pages/get-started-docker.adoc
nevzatseferoglu: Tutorial is running the hz docker image with `-e HZ_CLUSTERNAME=hello-world \` option. Therefore, as other clients did, setting the cluster's name `hello-world` is a better option than dev I guess.

nevzatseferoglu:
Test

yuce (approved):

utku-caglayan (approved):
Thanks, looks great!

PR #327, opened 2021-12-29
MC-1039 - 'data-access-enabled' MC config

promanenko:


docs/modules/maintain-cluster/pages/monitoring.adoc
```diff
@@ -147,6 +147,36 @@ hazelcast:
 ----
 ====
 
+=== Managing Data Access
+
+Management Center allows you to access contents of Hazelcast data structures (for instance map entries) via SQL Browser or Map Browser.
+Management Center can't access the data if at least one member has the data access disabled.
```
Serdaro, line 153: Does it make sense if we give some brief use cases which may require disabling the data access?

PR #328, opened 2021-12-30
Remove limitation around the schema validation for Compact

mdumandag:
We have decided to add XSD and JSON schema definition for the compact
format to the Hazelcast, even if it is still in BETA.

Due to that, disabling the schema validation is not required anymore.

So, this PR removes that warning.


mdumandag:
@Serdaro @JakeSCahill Hi guys, can we get this merged? Somehow, I cannot ask for reviews, so I had to write it here

PR #351, opened 2022-01-21
Update working-with-json.adoc

dbrimley:
quick typo fix.


JakeSCahill:
Thanks, this is fixed in #352 😄

PR #353, opened 2022-01-26
[HZ-771] Update query cache doc

ahmetmircik:
docs needed after https://github.com/hazelcast/hazelcast/pull/20265


docs/modules/query/pages/continuous-query-cache.adoc
```diff
@@ -155,6 +163,9 @@ hazelcast:
 Continuous query caches have the following configuration elements:
 
 * `name`: Name of your continuous query cache.
+* `serialize-keys`: (since 5.1) Defines if the Query Cache keys should be serialized or not.
```
JakeSCahill, line 166:
```suggestion
* `serialize-keys`: Whether Hazelcast serializes the query cache keys.
```

JakeSCahill (changes requested):
Thanks for documenting this new feature 😄 Just a few comments

docs/modules/query/pages/continuous-query-cache.adoc
```diff
@@ -82,6 +82,16 @@ latter is mostly explained in the previous section. The parent configuration ele
 is `<query-caches>` which should be placed within your `<map>` configuration. You can
 create your query caches using the  `<query-cache>` sub-element under `<query-caches>`.
 
+NOTE: Query Cache returns/uses internal key object reference during
```
JakeSCahill, line 85: I think this is too much information for the user.

How about this?

> If you use mutable keys in maps, consider setting the `serialize-keys` option to `true` to tell Hazelcast to clone the keys rather than use an internal reference. Otherwise, Hazelcast may store references to outdated versions of your keys.

JakeSCahill (approved):

PR #354, opened 2022-01-26
Update capacity planning doc

ahmetmircik:


JakeSCahill (changes requested):
I don't think we should mention how 4.2 behaved. IMDG is a different product. And this behavior is something that should be documented in the IMDG release notes.

Also, since this page is about capacity planning, this property does not need to be mentioned. Setting this property does not have anything to do with planning cluster capacity.

My view is that we can delete this new note as well as the one above.

ahmetmircik:
>I don't think we should mention how 4.2 behaved. IMDG is a different product. And this behavior is something that should be documented in the IMDG release notes.
Then how will a platform user be aware of such a knowledge? Platform is IMDG with new features so they are effectively asme products.

>Also, since this page is about capacity planning, this property does not need to be mentioned. Setting this property does not have anything to do with planning cluster capacity.
Can you please elaborate, i didn't get this? This is related with capacity of memory to be configured and fits under `Memory Considerations` section.

JakeSCahill:
> Then how will a platform user be aware of such a knowledge? Platform is IMDG with new features so they are effectively same products.

The docs are versioned, so any user reading these docs is assumed to be using that version of Platform. The details you added are about IMDG, which isn't relevant to a Platform user. Platform handles it automatically. The user doesn't need to do anything, and therefore doesn't need to know.

> Can you please elaborate, i didn't get this? This is related with capacity of memory to be configured and fits under Memory Considerations section.

Maybe I misunderstood how this feature works. The way I read it:

Hazelcast automatically assigns a percentage of available off-heap memory to the internal memory manager to store metadata for map/cache data structures and the persistence store. If you want to be informed about how much off-heap memory is being used by the memory manager, you can set the `metadata-space-percentage` property to a percentage value and Hazelcast will log warnings when the value is reached.

ahmetmircik:
Your understanding is correct, but effect of setting that property is unknown for a user without a note. 
I'm ok with replacing current note with your suggested one.

JakeSCahill (approved):

PR #358, opened 2022-01-31
Putting Enterprise labels.

Serdaro:
If we decide to go with the labels as in this PR, I am planning to move the Blue-Green Deployment section under Deploy and Manage Cluster chapter. This needs some re-writing in the Java client section where the content resides in currently.


JakeSCahill:
I'd leave the blue/green for now. This is a separate issue that needs a separate body of work.

The labels are great, thanks 😄 

I think we probably should remove labels from the side nav for now. I'll address that in this issue: https://github.com/hazelcast/hazelcast-docs-ui/issues/153

JakeSCahill:
There are lots of unnecessary commits in this PR. Could you try and branch off the current `main` and apply `edbeff4` to that?

PR #374, opened 2022-02-16
DOCS-166 Dynamic config updates

JakeSCahill:
- Added K8s usage procedure

- Removed `persistence-file` option

- Labeled persistence as an Enterprise feature


ramizdundar (approved):
I only have one minor comment. It would be great if we could fit this information somewhere.

When reload is called (not update) we look at the local declarative configurations and update based upon that. In k8 we still look at those files. The problem is when someone updates the config map, these changes will be relayed to the pods with a delay (2 to 3 min). Before this delay finishes, if user calls reload he will see no change. This was the original reason to introduce update endpoint.

This delay is caused by internal k8 settings and it would be unwise/impossible to change them. We also discussed reading directly from the config map but AFAIU that's not the cloud native way to do things.

PR #376, opened 2022-02-16
Change reader methods with new names

srknzl:
We need to change all pages with the changes like I did.


JakeSCahill:
@srknzl what do you mean? Are there more changes to come in this PR?

srknzl:
This changes only one file, I don't know if there are other outdated ones in other files. I did not check it, and most probably won't

mdumandag:
@srknzl Can we close this, as https://github.com/hazelcast/hz-docs/pull/386 contains all of these changes, with the other places where we need to update?

PR #383, opened 2022-02-17
Eviction size of 0 means no size limit

JakeSCahill:
Related to https://github.com/hazelcast/hazelcast-enterprise/issues/4777


ramizdundar (approved):
I have a minor comment, I believe 0 means no limit in map memory, however it behaves as disabled when used elsewhere (cache etc.). I am not 100% sure on this either, however if we explain eviction mechanism elsewhere, we should look into.

PR #384, opened 2022-02-17
HZ-716 Clarify limitations for Tiered Storage

JakeSCahill:


petrpleshachkov (commented):
Approve with minor comment

petrpleshachkov (approved):

docs/modules/configuration/pages/dynamic-config.adoc
```diff
@@ -58,7 +58,7 @@ Although you can configure members to xref:security:native-client-security.adoc#
 
 Dynamic configuration is supported for the following data structures:
 
-- Map
+- Map (apart from xref:tiered-storage:overview.adoc[Tiered Storage])
```
Holmistr, line 61: I think there's a misunderstanding. What you can't do is to turn on Tiered Storage on an existing map. But I believe it's still possible to dynamically add a new map that has TS enabled. Am I wrong @petrpleshachkov ?

docs/modules/tiered-storage/pages/overview.adoc
```diff
@@ -1,35 +1,37 @@
...
+Tiered Storage uses the following:
+
+* An in-memory index to allow faster access to map entries
+* A special memory manager to decide which entries to store in the memory tier or disk tier. If a map entry is not accessed for some time, it gets flushed to the disk. If a map is frequently accessed, it stays in memory.
 
-* an in-memory index to allow faster access to map entries
-* a special memory manager, i.e., HybridLog, to decide which entries to store in the memory tier or disk tier -
-if a map entry is not accessed for some time, it gets flushed to the disk; if it is a frequently accessed data, it stays in the memory.
-Note that an entry cannot be available in both tiers.
+NOTE: An entry is never available in both tiers.
```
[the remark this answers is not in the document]
Holmistr: I would explicitly add a note that this Hazelcast handles the moving between memory and disk transparently and users have (and will never have) no control over it.

Holmistr (commented):
Minor comments

docs/modules/configuration/pages/dynamic-config.adoc
petrpleshachkov: Yes, I meant to turn into tstore existing map is not possible.

Holmistr: @JakeSCahill Therefore, I propose to remove this docs change completely.

JakeSCahill: @Holmistr Wouldn't it be better to change it to 'You cannot add Tiered Storage configuration to an existing map'?

docs/modules/tiered-storage/pages/overview.adoc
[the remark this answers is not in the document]
Holmistr: Agreed. When recovery is implemented, I believe it will have to be described in more details.

docs/modules/configuration/pages/dynamic-config.adoc
Holmistr: Well, since we don't actually support any dynamic configuration changes in existing maps (we support only add and remove), I think it's just noise.

Holmistr: @JakeSCahill I believe this is the only outstanding topic. After that, I'll give a green tick.

JakeSCahill: Done 😄

Holmistr (approved):

PR #385, opened 2022-02-18
Adding content for Integrity Checker.

Serdaro:


docs/modules/management/pages/cluster-utilities.adoc
```diff
@@ -264,3 +264,73 @@ When the TLS mutual authentication is enabled, you have to provide the client ce
...
 NOTE: Currently, this script is not supported on the Windows platforms.
+
+== Integrity Checker
+
+The integrity checker is a utility developed for Java developers who build Hazelcast applications.
+It checks the classpath of your applications built with Hazelcast JAR dependencies (e.g., `hazelcast` or `hazelcast-sql`, using the assembly or other uber-JAR plugins).
+This check involves the verification of all user-provided hooks are loaded during a Hazelcast member startup, that is, it verifies that
+the executable of your application contains all the required `META-INF/services` files.
+For example, there may be cases which you add a new `DataSerializer` hook for your application,
+but do not update the corresponding `/META-INF/services/com.hazelcast.DataSerializerHook` file.
```
ivanthescientist, line 275: Probably a secondary example that we could provide here is: 

```
Or when an application is built with a plugin that merges dependency jars into application jar, in this case the plugin might be misconfigured and system /META-INF/services/com.hazelcast.DataSerializerHook files are not merged correctly, resulting in a corrupted executable. 
```

I think its important to drive the point that even if the user is not adding their own DataSerializerHooks, they might still encounter issues because build plugins were misconfigured etc.

ivanthescientist (approved):
Looks good, very easy to read! Added some very minor suggestions.

PR #440, opened 2022-04-26
Clarify client-side user code deployment

vbekiaris:


docs/modules/clusters/pages/deploying-code-from-clients.adoc
```diff
@@ -96,8 +96,7 @@ will be ignored.
 `LOCAL_AND_CACHED_CLASSES` on all members.
 * No `provider-filter` must be configured.
 
-The client uploads the classes only to one member. If the members don't load
-classes from each other, other members won't see the class.
+If the members don't load classes from each other, new members that join the cluster won't see the class.
```
Serdaro, line 99: @vbekiaris , minor wonderings:
1 - The client does not upload the code to one member?
2 - What can be a situation in which members cannot load classes from each other?

Serdaro (commented):
Minor questions.

docs/modules/clusters/pages/deploying-code-from-clients.adoc
vbekiaris: > @vbekiaris , minor wonderings:
> 1 - The client does not upload the code to one member?

The client sends a message to one member containing the code upload. The receiving member executes the client message which broadcasts the uploaded code to all members in the cluster. The fact that the client message is sent to one member is just an implementation detail, the code is actually uploaded to all members in the cluster.

> 2 - What can be a situation in which members cannot load classes from each other?

When members don't have user code deployment enabled (which is the default member config)

JakeSCahill: I think we can remove this sentence altogether. What do you think? We already say that members must be configured with certain settings for this feature to work. I would rather rewrite the sentence before the bulleted list to say: `To enable all existing and new members in a cluster to receive classes from clients, make sure that all member configurations follow these rules:`

PR #455, opened 2022-05-31
Clarify eviction

### LINKED ISSUE #219: Eviction documentation is hard to understand and maybe has small problems
I didn't get this part in the docs: https://docs.hazelcast.com/hazelcast/5.0/data-structures/managing-map-memory#understanding-map-eviction. If only one entry is going to be evicted then shouldn't we have 20000 entries? Because we had 20000, then we add one, then we evict one. If we want to tell them when eviction happens a number of entries going to be removed because we don't want to evict in every put we should say it like that. I don't have much knowledge about eviction but WDYT?

JakeSCahill:
Fixes #219

### DIFF
--- a/docs/modules/data-structures/pages/managing-map-memory.adoc
+++ b/docs/modules/data-structures/pages/managing-map-memory.adoc
@@ -128,8 +128,7 @@ partition-maximum-size = max-size * member-count / partition-count
 
 NOTE: If the `partition-maximum-size` is less than 1, it will be set to 1. Otherwise, all partitions would be emptied immediately.
 
-The eviction process starts when you try
-to add an entry to a map. If the entry count in the entry's partition exceeds the `partition-maximum-size`, eviction starts on that partition.
+Entries are evicted when you try to add an entry to a map only if the entry count in the entry's partition exceeds the value in the `partition-maximum-size` setting.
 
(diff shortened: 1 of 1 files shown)


docs/modules/data-structures/pages/managing-map-memory.adoc
```diff
@@ -120,40 +120,17 @@ memory size percentage for each Hazelcast instance.
...
-NOTE: If the `partition-maximum-size` is less than 1 in the equation above, it will be set to 1
-(otherwise, the partitions would be emptied immediately by eviction due to the exceedance of
-`max-size` being less than 1).
+NOTE: If the `partition-maximum-size` is less than 1, it will be set to 1. Otherwise, all partitions would be emptied immediately.
 
-The eviction process starts according to this calculated partition maximum size when you try
-to put an entry. When entry count in that partition exceeds partition maximum size, eviction
-starts on that partition.
+The eviction process starts when you try
+to add an entry to a map. If the entry count in the entry's partition exceeds the `partition-maximum-size`, eviction starts on that partition.
```
ramizdundar, line 132: Maybe something like:

Entries are evicted when you try to add an entry to a map only if the entry count in the entry's partition exceeds the `partition-maximum-size`.

ramizdundar (approved):
LGTM, thanks for the fix.

PR #459, opened 2022-06-02
Updated broken link for .Net

emreyigit:
Updated the broken link, and removed sentence which is not valid anymore.


Serdaro:
@emreyigit , I think it is safe to remove that sentence from all versions of Platform docs, right?

PR #495, opened 2022-08-08
Added ListenerWithPredicate sample for Go. [API-1485]

mehmettokgoz:


docs/modules/data-structures/pages/listening-for-map-entries.adoc
```diff
@@ -133,45 +139,30 @@ func main() {
...
-		case hazelcast.EntryLoaded:
-			fmt.Println("Entry Loaded:", event)
-		}
-	})
-
-
+	m, _ := client.GetMap(ctx, "map")
+	subscriptionID, _ := m.AddListenerWithPredicate(ctx, hazelcast.MapListener{
+		EntryAdded: func(event *hazelcast.EntryNotified) {
+			fmt.Println("Entry Added:", event.Key)
```
srknzl, line 145: Java code prints the event itself here, can we achieve the same somehow? 

Can we print an output like:
```
value=Person{name= smith }, event=ADDED, by Member [192.168.178.10]:5702
```

if it's not possible easily let's skip.

docs/modules/data-structures/pages/listening-for-map-entries.adoc
```diff
@@ -71,7 +71,13 @@ Go::
 -- 
 [source,go]
 ----
-
+func employee(name string, surname string) serialization.JSON {
```
srknzl, line 74: Alternative way to do this I think is the following:

```go
type Employee struct {
	Name string `json:"name"`
	Surname  string  `json:"surname"`
}
// To get a JSON value: 
b, err := json.Marshal(Employee{Name: "a", Surname: "b"})
if err != nil {
     jv := serialization.JSON(b)
     // use JSON value 
}
```

and this may be easier for larger structs. WDYT? Should we switch? cc @yuce

docs/modules/data-structures/pages/listening-for-map-entries.adoc
mehmettokgoz: Directly printing the event gives a little different output in Go.
```
Entry Added: &{<nil> 10 {
                "name":"Jack",
                "surname": "Smith"
        } <nil> map 127.0.0.1:5701:56df295c-370a-42b2-ab77-7d27812c7875 1 1}
```
So , we have to write a format function to get the same output as Java. Maybe we can leave it as above instead of `event.Key`.

docs/modules/data-structures/pages/listening-for-map-entries.adoc
mehmettokgoz: We discussed this alternative with @yuce but decided not to make things complex and use a simple function. We thought that we don't have to strictly follow the same structure with Java, wdyt?

docs/modules/data-structures/pages/listening-for-map-entries.adoc
utku-caglayan: @mehmettokgoz I think that's a good idea, can you implement the Stringer interface for EntryNotified type? We can try to make it similar to Java's output that Serkan shared

docs/modules/data-structures/pages/listening-for-map-entries.adoc
```diff
@@ -133,45 +139,30 @@ func main() {
...
 
 NOTE: See xref:query:how-distributed-query-works.adoc[How Distributed Query Works] for more information about creating predicates. 
 
-When the listener is running, a change to any record with the name `smith` will display output similar to the one below. 
+When the listener is running, a change to any record with the surname `smith` will display output similar to the one below.
 
 ```
-value=Person{name= smith }, event=ADDED, by Member [192.168.178.10]:5702
+value=Person{surname= smith }, event=ADDED, by Member [192.168.178.10]:5702
```
utku-caglayan, line 165: TODO let's change this one we update the output of the example

docs/modules/data-structures/pages/listening-for-map-entries.adoc
mehmettokgoz: Implemented `Stringer()` interface that reflects the Java client output, see https://github.com/hazelcast/hazelcast-go-client/pull/864 for more details.

yuce: We don't aim to have the same output with the reference implementation in these cases. In case we cannot skip the output in this case and the output have to look like reference implementation's, we can write a small utility function.

docs/modules/data-structures/pages/listening-for-map-entries.adoc
mehmettokgoz: We decided to keep the current output.

yuce (approved):

PR #497, opened 2022-08-09
Cpp client code added [API-1478]

hakanaktas0:
Cpp client code added

### DIFF
--- a/docs/modules/data-structures/pages/listening-for-map-entries.adoc
+++ b/docs/modules/data-structures/pages/listening-for-map-entries.adoc
@@ -109,7 +109,6 @@ C++::
 
 int main() {
     hazelcast::client::client_config config;
-    config.set_property("hazelcast.map.entry.filtering.natural.event.types", "true");
     auto hz = hazelcast::new_client(std::move(config)).get();
     auto map = hz.get_map("map").get();
     hazelcast::client::query::sql_predicate sqlPredicate(hz,"surname=smith");
--- a/docs/modules/data-structures/pages/listening-for-map-entries.adoc
+++ b/docs/modules/data-structures/pages/listening-for-map-entries.adoc
@@ -36,7 +36,18 @@ C++::
 --
 [source,cpp]
 ----
+#include <hazelcast/client/hazelcast_client.h>
 
+class Employee {
+public:
+    std::string surname;
+    Employee(std::string surname) {
+        this->surname = std::move(surname);
+    }
+    hazelcast::client::hazelcast_json_value to_json(){
+        return hazelcast::client::hazelcast_json_value("{ \"surname\": \"" + surname + "\" }");
+    }
+};
 
 ----
 --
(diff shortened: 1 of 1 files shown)


docs/modules/data-structures/pages/listening-for-map-entries.adoc
```diff
@@ -36,7 +36,18 @@ C++::
 --
 [source,cpp]
 ----
-
+#include <hazelcast/client/hazelcast_client.h>
+
+class Employee {
+public:
+    std::string surname;
```
ihsandemir, line 43: make it private or make this class a struct and you do not even need the constructor.

docs/modules/data-structures/pages/listening-for-map-entries.adoc
```diff
@@ -94,8 +105,24 @@ C++::
 --
 [source,cpp]
 ----
-
-
+#include "Employee.h"
```
ihsandemir, line 108: why don't we use include file statement similar to java version at line [89](https://github.com/hazelcast/hz-docs/pull/497/files#diff-3a0f447206aa7115457ee5e29aea9409826343642e7e1fec90ee94ae0af9554aL89) above.

docs/modules/data-structures/pages/listening-for-map-entries.adoc
```diff
@@ -94,8 +105,24 @@ C++::
...
+
+int main() {
+    hazelcast::client::client_config config;
+    config.set_property("hazelcast.map.entry.filtering.natural.event.types", "true");
```
ihsandemir, line 112: there is no such property. also, we should also remove that for Java, it should not be needed.

docs/modules/data-structures/pages/listening-for-map-entries.adoc
```diff
@@ -171,7 +198,7 @@ When the listener is running, a change to any record with the name `smith` will
 
 ```
-value=Person{name= smith }, event=ADDED, by Member [192.168.178.10]:5702
+value={surname=smith}, event=ADDED, by Member [192.168.178.10]:5702
```
ihsandemir, line 201: why is it changed?

docs/modules/data-structures/pages/listening-for-map-entries.adoc
hakanaktas0: this is how they implemented in java. I guess it is done that way to imitate a real world application

docs/modules/data-structures/pages/listening-for-map-entries.adoc
hakanaktas0: there is no object called Person in the code, I think it is a mistale

docs/modules/data-structures/pages/listening-for-map-entries.adoc
hakanaktas0: there is a ROOT folder to keep all the java example codes, there isn't for any other client. Maybe it can be restructured for other clients as well, but I think it should be discussed.

ihsandemir (approved):

ihsandemir (approved):

PR #508, opened 2022-08-18
Adding MapStore Offloading content.

Serdaro:


docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -250,6 +250,57 @@ include::ROOT:example$/dds/map/PersonEntryStore.java[tag=personms]
...
+----
+Config config = new Config();
+MapConfig mapConfig = config.getMapConfig();
+MapStoreConfig mapStoreConfig = mapConfig.getMapStoreConfig();
+mapStoreConfig.setOffload("true");
+...
+----
+====
+
+NOTE: Offloading cannot be used for xref:tiered-storage:overview.adoc[Tiered Storage], and for `MapLoaderLifecycleSupport`, since it is a one time operation per member along the lifecycle of a map.
```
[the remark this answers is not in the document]
JakeSCahill: I also see this in the design doc: `No changes in cluster wide map bulk loading mechanism.` Does this just mean the `loadAllKeys()` method?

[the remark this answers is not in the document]
JakeSCahill: I think we should move any exceptions up to the bulleted list that I suggested above.

JakeSCahill (changes requested):

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -250,6 +250,57 @@ include::ROOT:example$/dds/map/PersonEntryStore.java[tag=personms]
 
 TIP: link:https://github.com/hazelcast/hazelcast-code-samples/tree/master/distributed-map/mapstore/src/main/java[See more MapStore/MapLoader code samples].
 
+== Offloading MapStore and MapLoader Operations
+
+Offloading `MapStore` and `MapLoader` operations allow these operations to work asynchronously, meaning they do not block each other.
```
ahmetmircik, line 255:
```suggestion
Offloading `MapStore` and `MapLoader` operations allow these operations to work asynchronously, meaning they do not block partition threads.
```

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -250,6 +250,57 @@ include::ROOT:example$/dds/map/PersonEntryStore.java[tag=personms]
...
+This provides faster executions of these operations, resulting in higher throughput of the whole cluster.
```
ahmetmircik, line 256:
```suggestion
This provides faster executions of other operations in cluster, since partition threads are not blocked by MapStore/MapLoader calls, resulting in higher throughput of the whole cluster.
```

docs/modules/data-structures/pages/working-with-external-data.adoc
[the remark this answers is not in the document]
ahmetmircik: @JakeSCahill currently you cannot have both map-store and tiered-storage enabled for a map. This is by design.

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -250,6 +250,57 @@ include::ROOT:example$/dds/map/PersonEntryStore.java[tag=personms]
 
 TIP: link:https://github.com/hazelcast/hazelcast-code-samples/tree/master/distributed-map/mapstore/src/main/java[See more MapStore/MapLoader code samples].
 
+== Offloading MapStore and MapLoader Operations
```
ahmetmircik, line 253: @Serdaro you can have up-to-date info for the design here: https://github.com/hazelcast/hazelcast/blob/master/docs/design/map/map-store-operation-offloading.md

docs/modules/data-structures/pages/working-with-external-data.adoc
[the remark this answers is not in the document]
JakeSCahill: So maybe it's worth having a 'Limitations' section higher up this page? So users are aware of what doesn't work with MapStore?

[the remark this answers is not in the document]
JakeSCahill: This location doesn't seem like the place to mention a limitation that affects all MapStores.

docs/modules/data-structures/pages/working-with-external-data.adoc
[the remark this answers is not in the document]
JakeSCahill: @ahmetmircik WDYT of this?

docs/modules/data-structures/pages/working-with-external-data.adoc
ahmetmircik: @Serdaro also worth to mention the name of the executor for the offloaded calls and how oen can configure it: https://github.com/hazelcast/hazelcast/blob/master/docs/design/map/map-store-operation-offloading.md#configuring-offloadable-executor

docs/modules/data-structures/pages/working-with-external-data.adoc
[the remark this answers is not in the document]
ahmetmircik: >To improve cluster throughput, most MapStore operations do not block partition threads.

Not most but all MapStore operations are async by default. Only `MapLoaderLifecycleSupport` is sync but it runs only one time during the creation of IMap.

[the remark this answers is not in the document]
JakeSCahill: > > MapStore operations that communicate with the external data source are handled asynchronously, using an entry processor.
> 
> I didn't get the relation between EP and async map-store calls. Async map-store calls are done independently, EP is not required.

It's an executor instead?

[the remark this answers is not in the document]
JakeSCahill: > > To improve cluster throughput, most MapStore operations do not block partition threads.
> 
> Not most but all MapStore operations are async by default. Only `MapLoaderLifecycleSupport` is sync but it runs only one time during the creation of IMap.

Since `MapLoaderLifecycleSupport` is sync, isn't it incorrect to say all?

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -250,6 +250,57 @@ include::ROOT:example$/dds/map/PersonEntryStore.java[tag=personms]
...
+[tabs] 
+==== 
+XML:: 
++ 
+-- 
+[source,xml]
+----
+<hazelcast>
+    <map name="default">
+        <map-store enabled="true" initial-mode="LAZY">
```
ahmetmircik, line 268:
```suggestion
        <map-store enabled="true">
```
Having LAZY initial mode is not required.

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -250,6 +250,57 @@ include::ROOT:example$/dds/map/PersonEntryStore.java[tag=personms]
...
+----
+
+Java::
++
+[source,java]
+----
+Config config = new Config();
+MapConfig mapConfig = config.getMapConfig();
+MapStoreConfig mapStoreConfig = mapConfig.getMapStoreConfig();
+mapStoreConfig.setOffload("true");
```
ahmetmircik, line 298:
```suggestion
mapStoreConfig.setOffload(true);
```

docs/modules/data-structures/pages/working-with-external-data.adoc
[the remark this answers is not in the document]
ahmetmircik: >Since MapLoaderLifecycleSupport is sync, isn't it incorrect to say all?

No, it is a separate interface and implementing it is not required, it is up-to-user. If user is implemented it, it is blocking.

docs/modules/data-structures/pages/working-with-external-data.adoc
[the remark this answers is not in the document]
ahmetmircik: makes sense.

docs/modules/data-structures/pages/working-with-external-data.adoc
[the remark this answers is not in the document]
JakeSCahill: > > It's an executor instead?
> 
> It is an executor behind the scenes, see this please: [#508 (comment)](https://github.com/hazelcast/hz-docs/pull/508#discussion_r951240867)

Got it. I think we should mention it here in that case.

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -295,12 +306,50 @@ Java::
...
+XML:: 
++ 
+-- 
+[source,xml]
+----
+<hazelcast>
+    ...
+    <executor-service name="hz:map-store-offloadable">
+        <pool-size>1</pool-size>
+        <queue-capacity>10</queue-capacity>
```
ahmetmircik, line 331: better to have default settings here:
```
pool-size: 16
queue-capacity: 0
```

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -295,12 +306,50 @@ Java::
...
+YAML::
++
+[source,yaml]
+----
+hazelcast:
+  ..
+  executor-service:
+    "hz:map-store-offloadable":
+      pool-size: 1
+      queue-capacity: 10
```
ahmetmircik, line 347: same here, better to have default settings here:
```
pool-size: 16
queue-capacity: 0
```

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -250,6 +250,106 @@ include::ROOT:example$/dds/map/PersonEntryStore.java[tag=personms]
...
+
+To improve cluster throughput, all MapStore operations do not block partition threads.
+Each partition in a member is managed by a single thread, ensuring that the write operations to a given partition are handled
+one at a time in first-in-first-out order. However, MapStore operations may take a long time to connect to the external data source and complete, blocking the affected partitions. 
+To avoid blocking partition threads, MapStore operations that communicate with the external data source are handled asynchronously, using an <<configuring-the-offload-executor, executor>>.
+
+=== Limitations
+
+* The `init()` and `destroy()` methods of the `MapLoaderLifecycleSupport` interface cannot be offloaded. These methods are triggered only once in the whole lifecycle of a map.
+* MapStore methods `loadAll()` and `loadAllKeys()` cannot be offloaded.
```
ahmetmircik, line 263: >MapStore methods `loadAll()` and `loadAllKeys()` cannot be offloaded.

These methods are also offloaded independent of the work done in 5.2. In 5.1, they are also working on offloaded manner.

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -250,6 +250,106 @@ include::ROOT:example$/dds/map/PersonEntryStore.java[tag=personms]
 
 TIP: link:https://github.com/hazelcast/hazelcast-code-samples/tree/master/distributed-map/mapstore/src/main/java[See more MapStore/MapLoader code samples].
 
+== Offloading MapStore and MapLoader Operations
+
+To improve cluster throughput, all MapStore operations do not block partition threads.
```
ahmetmircik, line 255:
```suggestion
To improve cluster throughput, MapStore/MapLoader api calls are offloaded. This way partition threads are not blocked by them.
```

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -250,6 +250,106 @@ include::ROOT:example$/dds/map/PersonEntryStore.java[tag=personms]
...
+=== Limitations
+
+* The `init()` and `destroy()` methods of the `MapLoaderLifecycleSupport` interface cannot be offloaded. These methods are triggered only once in the whole lifecycle of a map.
+* MapStore methods `loadAll()` and `loadAllKeys()` cannot be offloaded.
+* Offloading cannot be used for xref:tiered-storage:overview.adoc[tiered storage], since you cannot enable both MapStore and tiered storage for a map.
+
+=== Blocking Partition Threads
+
+If you want to block partition threads until MapStore operations are complete, 
+you can configure the `offload` setting for the MapStore. It is enabled by default (`true`) and can be configured for each map in your cluster.
```
ahmetmircik, line 269: >If you want to block partition threads until MapStore operations are complete, 
you can configure the `offload` setting for the MapStore. It is enabled by default (`true`) and can be configured for each map in your cluster.

To restore the behavior in previous releases than 5.2, you can set `offload` setting for the `MapStoreConfig` to `false` which is `true` by default. This way partition threads will be blocked during MapStore/MapLoader api calls.

Serdaro: We try to avoid mentioning the product release numbers, since the documentation is versioned.
Instead, I can simply give the idea of "set to "false" to block".

ahmetmircik: My point is we should emphasis that the point of making it false is only bringing old behavior back. Giving version number is not necessary.

ahmetmircik (approved):

PR #518, opened 2022-08-31
Add docs for GenericMapStore

### LINKED ISSUE #271: Uplifting MapStore / MapLoader
Currently, MapStore/MapLoader is located in the documentation as a subsection of the Map data structure, which is not so visible.

_Develop Solutions > Distributed Data Structures > Map > [Working with an External Data Store](https://docs.hazelcast.com/hazelcast/5.0/data-structures/working-with-external-data)_

Since this capability addresses one of the major use cases, I believe we should uplift this section, i.e. add it directly to the _Develop Solutions_ section.

An idea without moving the current section:
We could consider adding a subsection directly under the Develop Solutions named **_“Working with External Databases / Data Stores“_** which has a very short introduction with a link to the actual part that is under the Map section.

JakeSCahill:
Fixes #271


Serdaro:
Nice work!

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -170,12 +177,23 @@ class ProcessingStore implements MapStore<Integer, Employee>, PostProcessingMapS
...
 
-Once you've created your `MapStore` implementation, you need to configure your map to use it. 
+[[existing-connection]]
+=== Using an Existing Connection
+
+If you have an xref:external-data-stores:external-data-stores.adoc[external data store] already configured, you can configure your MapStore to use that existing connection.
+
+When you configure your MapStore to use that existing connection, Hazelcast creates a xref:sql:mapping-to-jdbc.adoc[SQL mapping with the JDBC connector]. The name of the mapping is the same name as your map prefixed with `__map-store.`. This mapping is then used to read data from the store, or write data to it. This mapping is removed when the map is removed.
+
+This is a basic example configuration that shows you the minimal configuration that is required. This example is for an external data store configuration called `my-mysql-database`.
```
frant-hartm, line 196: I think this mixes 2 things:
- configuring the MapStore on a map
- the GenericMapStore implementation

They should be separate (either a separate section or even a page). Also, the GenericMapStore should be more prominent (basically at the top). It should be the first option the user should try to use and only if it doesn't work they should resort to own implementation. However it's currently in beta for 5.2, so needs to be marked as such.

docs/modules/data-structures/pages/working-with-external-data.adoc
```diff
@@ -208,25 +284,628 @@ hazelcast:
...
+MapConfig mapConfig = new MapConfig("default");
+
+MapStoreConfig mapStoreConfig = new MapStoreConfig();
+mapStoreConfig.setClassName(GenericMapStore.class.getName());
+mapStoreConfig.setProperty("external-data-store-ref", "my-mysql-database");
+mapStoreConfig.setProperty("table-name", "test");
+
+mapConfig.setMapStoreConfig(mapStoreConfig);
+----
+--
```
frant-hartm, line 730: This again doesn't fit well, we have configuration options under the `map-store` element. They apply to all implementations. Then there is couple of properties that apply only to `GenericMapStore` and go into the `properties` section.

These should be separate.

frant-hartm (changes requested):

docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
```diff
@@ -0,0 +1,383 @@
...
+
+== Related Resources
+
+To monitor MapStores for each loaded entry, use the `EntryLoadedListener` interface. See the xref:events:object-events.adoc#listening-for-map-events[Listening for Map Events section] to learn how you can catch entry-based events.
+
+xref:mapstore-triggers.adoc[].
+
+== Next Steps
+
+xref:configuration-guide.adoc[Configure a map to use the generic MapStore]. The cluster does not load the MapStore until you do this.
```
frant-hartm, line 383: I find this a bit confusing, didn't we already configure a map store earlier in this section?

frant-hartm (approved):
Thanks! I like this new structure a lot better. One minor comment.

docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
JakeSCahill: Ah yes, thanks. I adjusted that sentence so that the next step is to see a general configuration guide for MapStores rather than as a must-do.

docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
```diff
@@ -0,0 +1,383 @@
+= Configuring a Map with the Generic MapStore
+:description: You can configure a map with the generic MapStore to use it as a cache with an external data store.
```
frant-hartm, line 2: This is actually a beta feature in 5.2. Should be marked as such. (depends on JDBC SQL Connector which is also beta).

frant-hartm (changes requested):
Realized one more thing.

frant-hartm (approved):

PR #523, opened 2022-09-13
Add Windows Installation Steps

utku-caglayan:


JakeSCahill:
I think we should focus any new content on the dedicated docs: https://github.com/hazelcast/hazelcast-commandline-client/pull/94

PR #524, opened 2022-09-15
Fix #485 and add behavior of 0 auto removal setting

### LINKED ISSUE #485: Wrong link for manually removing the CP members
Hi, 
The link for manually removing CP members in https://docs.hazelcast.com/hazelcast/5.1/cp-subsystem/configuration#persistence-enabled is wrong (currently, it refers to its own section). I think the correct link should be https://docs.hazelcast.com/hazelcast/5.1/cp-subsystem/management#removing-a-cp-member.

JakeSCahill:


arodionov (approved):
Thanks for updating the docs!

PR #572, opened 2022-10-20
#421 Workaround for no automatic reconnection to primary cluster

### LINKED ISSUE #421: Docs: Feedback for Deploy Blue-Green Clusters
Hi, I have some feedback about [this page](https://github.com/hazelcast/hz-docs/edit/main/docs/modules/getting-started/pages/blue-green.adoc)

Some users assume that once the primary cluster becomes available again, clients would reconnect to that cluster instead of the backup one.  I think we should document that it is not the case (i.e. it does not happen automatically. Needs user config). Considering the number of confusions, we should manage customers' expectation here.

cc: @sancar, @ihsandemir, @mdumandag

angelasimms:
Closes #421.

A separate PR is in progress to update the Management Center docs.


JakeSCahill (approved):

JakeSCahill:
I addressed James' point in this PR to improve the Management Center docs about client filtering: https://github.com/hazelcast/management-center-docs/pull/206

It's waiting for a reviewer 😄

PR #584, opened 2022-11-03
Unify serialization mechanisms in the entry listener with predicate documentation API-1572

srknzl:
This should backport to 5.0, 5.1 and 5.2

### DIFF
--- a/docs/modules/ROOT/examples/dds/map/Employee.java
+++ b/docs/modules/ROOT/examples/dds/map/Employee.java
@@ -7,7 +7,7 @@ public class Employee {
         this.surname = surname;
     }
 
-    public HazelcastJsonValue toHazelcastJsonValue() {
+    public static HazelcastJsonValue toHazelcastJsonValue(String surname) {
         return new HazelcastJsonValue("{ \"surname\": \"" + surname + "\" }");
     }
 }
--- a/docs/modules/ROOT/examples/dds/map/Employee.java
+++ b/docs/modules/ROOT/examples/dds/map/Employee.java
@@ -1,7 +1,5 @@
-import java.io.Serializable;
-
 //tag::emp[]
-public class Employee implements Serializable {
+public class Employee {
 
     private final String surname;
 
@@ -9,11 +7,8 @@ public class Employee implements Serializable {
         this.surname = surname;
     }
 
-    @Override
-    public String toString() {
-        return "Employee{" +
-                "surname='" + surname + '\'' +
-                '}';
+    public HazelcastJsonValue toHazelcastJsonValue() {
+        return new HazelcastJsonValue("{ \"surname\": \"" + surname + "\" }");
     }
 }
 //end::emp[]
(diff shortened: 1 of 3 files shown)


docs/modules/ROOT/examples/dds/map/Employee.java
```diff
@@ -1,19 +1,14 @@
...
         this.surname = surname;
     }
 
-    @Override
-    public String toString() {
-        return "Employee{" +
-                "surname='" + surname + '\'' +
-                '}';
+    public HazelcastJsonValue toHazelcastJsonValue() {
+        return new HazelcastJsonValue("{ \"surname\": \"" + surname + "\" }");
```
mdumandag, line 11: I think this can just be a static method and that way we can eliminate one unnecessary object creation while adding some entries to map. I guess we don't show this step, but if we change the text around this class, I think we can just say we are going to use HJV instead of a class

docs/modules/data-structures/pages/listening-for-map-entries.adoc
```diff
@@ -58,8 +58,8 @@ class Employee {
         this.surname = surname;
     }
 
-    toString() {
-        return "Employee: " + this.surname;
+    toHazelcastJsonValue() {
```
mdumandag, line 61: I can't add comment to that line so, putting it here: 

In a bit below the code samples, we show a log like 
```
value=Person{surname= smith }, event=ADDED, by Member [192.168.178.10]:5702
```
which is not in line with the current status

docs/modules/ROOT/examples/dds/map/Employee.java
srknzl: This method uses surname field of Employee. How can it be static?

docs/modules/data-structures/pages/listening-for-map-entries.adoc
srknzl: Updated in 27094ff09f1fe121a6d845c0c40fc96177664343

docs/modules/ROOT/examples/dds/map/Employee.java
srknzl: Updated in b4807f58c20f5ef99d3cf2e9bdb595aaa24c69f9

docs/modules/data-structures/pages/listening-for-map-entries.adoc
```diff
@@ -39,8 +40,7 @@ C++::
 #include <hazelcast/client/hazelcast_client.h>
 
 struct Employee {
-    std::string surname;
-    hazelcast::client::hazelcast_json_value to_json(){
+    hazelcast::client::hazelcast_json_value as_json(){
```
mdumandag, line 43: missing parameter to the method. also, don't we have to make it static?

mdumandag (approved):

srknzl:
Status: Awaiting review from one of documentation team members

Serdaro (approved):

PR #585, opened 2022-11-03
Unify code samples to use JSON serialization API-1572

srknzl:


srknzl:
HI documentation team, any idea why check link workflow fails?

PR #587, opened 2022-11-03
Unify code samples to use JSON serialization API-1572

srknzl:


srknzl:
Hi doc team, any idea why link checking fails ?

Serdaro:
@srknzl , It is related to the [check links](https://github.com/srknzl/hz-docs/blob/api-1572-v5.2/check-links-playbook.yml) action, which tries to find your content in the "main" branch (which does not exist since the content is for 5.2 branch). Is this content only specific for 5.2 or is it also valid for the future releases? If latter, could you please change the PR branch to "main" (in this case we will put the "backport to 5.2" label)?

srknzl:
Ok did not know that. Closing other PRs than the main PR.

PR #588, opened 2022-11-07
Make Stream merge example more realistic

k-jamroz:
Use proper (and different) topic names for exchanges to make the example more realistic.


Serdaro:
@k-jamroz , Do we need to reference these new names in other related codes in this section of the documentation?

Serdaro (approved):

PR #619, opened 2022-12-06
Increase width for mapping parameter examples

k-jamroz:
Mapping examples are not readable in narrow column, narrow column makes them also needlessly long.


k-jamroz:
These widths are still not perfect (when viewed on Mac screen) but better than before. Lines in examples are still wrapped. On the other hand, some descriptions also are long.

Serdaro (approved):

Serdaro:
Thank you @k-jamroz

PR #664, opened 2023-02-23
[CN-720] Restructure Kubernetes section

cagric0:
Move helm chart readmes to the doc.
Update kubernetes overview page.


docs/modules/kubernetes/pages/helm-configuring-ssl.adoc
```diff
@@ -0,0 +1,67 @@
...
+:page-enterprise: true
+
+To enable SSL-protected communication between members and clients, you need first to generate `keystore`/`truststore` and import them as secrets into your Kubernetes environment.
+
+[source,shell]
+----
+kubectl create secret generic keystore --from-file=./keystore --from-file=./truststore
+----
+
+Then, since Kubernetes liveness/readiness probes cannot use SSL, we need to prepare Hazelcast configuration with a separate non-secured port opened for health checks. Create a file `hazelcast.yaml`.
```
Serdaro, line 11: ... cannot use SSL, you need to prepare a Hazelcast ...

... Create `hazelcast.yaml` with the following content.

(not sure if it is meaningful, but maybe we can also tell the user where on the system to create this file. When you use the --from-file argument, it takes the file from the location where the command is run?)

Serdaro (commented):
Minor comments.

Serdaro (approved):

PR #668, opened 2023-02-27
Fix XML config example for custom ByteArraySerializer

kwart:


Serdaro (approved):
LGTM</br>_// approved in Slack with WorkerB_

PR #628, opened 2022-12-20
Add compact binary specification API-1759

srknzl:
Publication of https://hazelcast.atlassian.net/wiki/spaces/IMDG/pages/3507487040/Compact+Wire-level+Format+Design+Document with some changes. The purpose is to describe our binary compact format to public. 

I refactored the document so that we don't use `we` and `will` words. I also removed the sections about discussion and brain storming. The document should be format and describe what is the binary format as of now. I used grammarly to do a grammar check after that.

PR #671, opened 2023-02-28
Fix header styling in compact binary specification document

srknzl:
According to https://docs.asciidoctor.org/asciidoc/latest/tables/build-a-basic-table/#add-a-header-row-to-the-table there should be a newline after the header line. 

Should be backported to all other branches that it is backported in the main [PR](https://github.com/hazelcast/hz-docs/pull/628).


srknzl:
@yuce  The preview has the format correct now. https://deploy-preview-671--hardcore-allen-f5257d.netlify.app/hazelcast/5.3-snapshot/compact-binary-specification

PR #674, opened 2023-03-08
Fix best-practices.adoc - recommended heap

Holmistr:
The recommendation is probably "maximum" heap size, rather than "minimum". Especially with the extra info about HD memory.


Holmistr:
@Serdaro Actually brainstorming it here with @jerrinot - maybe it's correct actually. It talks about RAM, not heap. Can you actually double check?

PR #679, opened 2023-03-10
Adding TS metrics.

Serdaro:


k-jamroz:
@blazember could you have a look if the descriptions make sense?

k-jamroz (approved):

docs/modules/ROOT/pages/list-of-metrics.adoc
```diff
@@ -1570,6 +1571,207 @@ with the master.
 |count
 |Number of started transactions
 
+|`tstore.device.freeSpace`
+|bytes
+|Amount of free space in the device directory
+
+|`tstore.device.maxSpace`
+|bytes
+|Amount of total space in the device directory
```
blazember, line 1580: Isn't this the maximum usage of the directory?

blazember (approved):
Looks good, just one comment on what we may want to clarify.

docs/modules/ROOT/pages/list-of-metrics.adoc
k-jamroz: no, it is calculated as `File.getTotalSpace()` for `baseDir` which in turn
> Returns the size of the partition named by this abstract pathname

So it is size of the partition/disk not a max space that was ever used by the Hybrid Log files in given device.

Serdaro: I also thought (from the metric name) it'd be the maximum. I am guessing changing the metric name is a bit complicated right (like `tstore.device.totalSpace`)?

PR #692, opened 2023-03-24
Update map-config.adoc: fixed link to yaml example

ExtraClock:
A link to a yaml example referenced an xml one.


Serdaro:
Thanks a lot @ExtraClock

Serdaro:
Fixed with recent commits.

PR #699, opened 2023-03-30
Add some necessary information to tutorial [API-1927]

fatihozer0:


docs/modules/pipelines/pages/python.adoc
```diff
@@ -17,6 +17,8 @@ To complete this tutorial, you need the following:
 
 |A Hazelcast cluster running in client/server mode
 |xref:getting-started:install-hazelcast.adoc#use-the-binary[Install Hazelcast]
+
+|A Hazelcast Full Distribution |https://hazelcast.com/open-source-projects/downloads/[Hazelcast Open Source Downloads]
```
fatihozer0, line 21: Slim distribution does not provide Jet, this is why I added it like this. 
`A Hazelcast Full Distribution` does not sound very well. Any better idea?

Serdaro: @fatihozer0 , we mention it as stated at https://docs.hazelcast.com/hazelcast/5.2/getting-started/editions#full-and-slim-distributions. So I think it is the correct text.

Serdaro (approved):

PR #717, opened 2023-04-18
CN-794 Remove outdated helm chart values

dzeromski-hazelcast:
We want to keep one soruce of truth values in https://github.com/hazelcast/charts/blob/master/stable/hazelcast/values.yaml. Right now is hard to keep all different places with values docs in sync.


hasancelik (approved):

cagric0 (approved):
Did you also check if the values.yaml file has all the descriptions from values table as comments?

PR #739, opened 2023-05-24
DOCS-523 SQL Catalog Persistence draft

angelasimms:
Early draft. Catalog persistence changes.

### DIFF
--- a/docs/modules/ROOT/examples/storage/SampleHotRestartConfiguration.java
+++ b/docs/modules/ROOT/examples/storage/SampleHotRestartConfiguration.java
@@ -30,6 +30,11 @@ public class SamplePersistenceConfiguration {
         cacheConfig.getMerkleTreeConfig().setDepth(12);
         config.addCacheConfig(cacheConfig);
 
+        MapConfig mapConfig = config.getMapConfig("__sql.catalog");
+        mapConfig.getDataPersistenceConfig().setEnabled(true);
+        mapConfig.getSqlConfig().setEnabled(true);
+        config.addMapConfig(mapConfig);
+
         config.getJetConfig().setLosslessRestartEnabled(true);
         //end::hrconf[]
     }
--- a/docs/modules/ROOT/examples/storage/SampleHotRestartConfiguration.java
+++ b/docs/modules/ROOT/examples/storage/SampleHotRestartConfiguration.java
@@ -30,10 +30,7 @@ public class SamplePersistenceConfiguration {
         cacheConfig.getMerkleTreeConfig().setDepth(12);
         config.addCacheConfig(cacheConfig);
 
-        MapConfig mapConfig = config.getMapConfig("__sql.catalog");
-        mapConfig.getDataPersistenceConfig().setEnabled(true);
-        mapConfig.getSqlConfig().setEnabled(true);
-        config.addMapConfig(mapConfig);
+        config.getSqlConfig().setCatalogPersistenceEnabled(true);
 
         config.getJetConfig().setLosslessRestartEnabled(true);
         //end::hrconf[]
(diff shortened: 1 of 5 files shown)


docs/modules/ROOT/examples/storage/SampleHotRestartConfiguration.java
```diff
@@ -30,6 +30,11 @@ public static void main(String[] args) throws Exception{
         cacheConfig.getMerkleTreeConfig().setDepth(12);
         config.addCacheConfig(cacheConfig);
 
+        MapConfig mapConfig = config.getMapConfig("__sql.catalog");
```
ivanthescientist, line 33: Sorry I probably miscommunicated here, but users aren't supposed to configure this IMap at all. They're supposed to use the `CatalogPersistenceEnabled` option in the `SqlConfig`
```
config.getSqlConfig().setCatalogPersistenceEnabled(true);
```

In general `__sql.catalog` is an internal object and might be replaced with something else in the future (similar to `__jet.*` internal IMaps), so we definitely don't want users accessing it directly.

docs/modules/storage/pages/configuring-persistence.adoc
```diff
@@ -1261,6 +1261,55 @@ config.addMapConfig(mapConfig);
...
+
+[tabs] 
+==== 
+XML:: 
++ 
+--
+[source,xml]
+----
+<hazelcast>
+  <map name="__sql.catalog">
```
ivanthescientist, line 1277: Same as the Java configuration, there is an option in the SQL Config specifically to enable SQL Catalog persistence in a safe way: 

```
<hazelcast>
    <sql>
        <catalog-persistence-enabled>true</catalog-persistence-enabled>
    </sql>
</hazelcast>
```

docs/modules/storage/pages/configuring-persistence.adoc
```diff
@@ -1261,6 +1261,55 @@ config.addMapConfig(mapConfig);
...
+MapConfig mapConfig = config.getMapConfig("__sql.catalog");
+mapConfig.getDataPersistenceConfig().setEnabled(true);
+mapConfig.getSqlConfig().setEnabled(true);
+
+config.addMapConfig(mapConfig);
+----
+--
+====
+
+NOTE: When implementing permissions for the SQL catalog IMap (`__sql.catalog`), we do not recommend using wildcards (*).
```
ivanthescientist, line 1311: We probably want to get rid of this line altogether given that we don't want users to configure anything about `__sql.catalog` directly (I'm not sure it includes security specifically though).

docs/modules/storage/pages/configuring-persistence.adoc
```diff
@@ -1261,6 +1261,55 @@ config.addMapConfig(mapConfig);
 ====
 |===
 
+[[sql]]
+== SQL Metadata Options
+
+Use the following configuration to enable the cluster-wide persistence of SQL metadata to disk. With persistence enabled, SQL mappings, data connections, and views are retained after the cluster restarts.
+
```
ivanthescientist, line 1268: Something I forgot to mention, but we probably want to add a note here: 

```
NOTE: A known issue is that for clusters larger than 6 members, it might take some time for the SQL Catalog to become available on all nodes after a restart, which may impact Hot Restart of jobs that use catalog entities (e.g. jobs that use Data Connections).
```

Its a known bug that unfortunately could not be fixed in 5.3.0.

docs/modules/ROOT/examples/storage/SampleHotRestartConfiguration.java
angelasimms: Updated with correct config.

docs/modules/storage/pages/configuring-persistence.adoc
angelasimms: Updated with correct config.

docs/modules/storage/pages/configuring-persistence.adoc
angelasimms: Removed.

docs/modules/storage/pages/configuring-persistence.adoc
angelasimms: Added note.

ivanthescientist (approved):

Serdaro (approved):
LGTM

Issue #744, opened 2023-05-25
Docs: Feedback for Start a Local Cluster in Docker


nishaatr:
Hi, I have some feedback about [this page](https://github.com/hazelcast/hz-docs/edit/main/docs/modules/getting-started/pages/get-started-docker.adoc)

The 'Cluster states' link end of page navigates to 'Cluster Utilities' https://docs.hazelcast.com/hazelcast/5.3/management/cluster-utilities#cluster-states

![image](https://github.com/hazelcast/hz-docs/assets/12186256/4759baf0-ee71-4804-b26b-e8af7e7fda0c)

I think it should link to https://docs.hazelcast.com/hazelcast/5.3/maintain-cluster/cluster-member-states

Serdaro:
Fixed with above commits to all 5.x branches.

Issue #741, opened 2023-05-24
Docs: Feedback for Locking Maps


nishaatr:
Hi, I have some feedback about [this page](https://github.com/hazelcast/hz-docs/edit/main/docs/modules/data-structures/pages/locking-maps.adoc)

Should the below read as '...values are **not** equal...' ? Otherwise why replace a value which is equal. May be I am reading this wrong!

![image](https://github.com/hazelcast/hz-docs/assets/12186256/7af0bdb5-ff1c-4ec6-91cc-672d6605f35a)

Serdaro:
Fixed with above commits to all 5.x branches.

Issue #731, opened 2023-05-12
Docs: Feedback for Deploying User Code on the Member


jedikcz:
Hi, I have some feedback about [this page](https://github.com/hazelcast/hz-docs/edit/main/docs/modules/clusters/pages/deploying-code-on-member.adoc).

The link `generic object interface` in Note section is invalid. I'd say it should point to https://docs.hazelcast.com/hazelcast/5.2/clusters/accessing-domain-objects.

![image](https://github.com/hazelcast/hz-docs/assets/45359017/6d9ddf2f-6605-4e08-83a5-19494fb01e06)

Serdaro:
Fixed with above commits to all 5.x branches.

PR #745, opened 2023-05-25
Fixes for #744, #741, #731

Serdaro:


Serdaro:
Fixes #744  , #741  and #731

PR #769, opened 2023-06-12
DOCS-551_Deprecation notice for transactions

angelasimms:
Added the following note to transaction sections and added a equivalent line of text to the release notes. Future 5.4 release notes will also need to be updated.

**Deprecation Notice for Transactions**
Hazelcast will deprecate transactions in the upcoming release (5.4). An improved version of this feature is under consideration. If you are already using transactions, get in touch and share your use case. Your feedback will help us to develop a solution that meets your needs.

Closes [DOCS-551](https://hazelcast.atlassian.net/browse/DOCS-551)

[DOCS-551]: https://hazelcast.atlassian.net/browse/DOCS-551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/modules/clients/pages/java.adoc
+++ b/docs/modules/clients/pages/java.adoc
@@ -389,7 +389,16 @@ each distributed data structure in this Reference Manual.
 === Client Transactions
 
 Transactional distributed objects are supported on the client side.
-See the xref:transactions:providing-xa-transactions.adoc[Transactions chapter] on how to use them.
+See xref:transactions:providing-xa-transactions.adoc[Transactions] for more details.
+
+[NOTE]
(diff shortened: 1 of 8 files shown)


nandankidambi:
@angelasimms : May be we should remove the 5.5 reference. Say something like "We are considering an improved version of transactions. If you are using transactions get in touch with us to talk about your use case."

nandankidambi:
Looks good

Serdaro (approved):
LGTM

PR #782, opened 2023-06-28
Update TCP tuning section

vbekiaris:
- correct kernel parameter names
- provide examples how to set kernel parameters

@Serdaro I think it would be nice to backport this to supported versions' documentation branches because currently they mention incorrect kernel parameter names


Serdaro (approved):

Serdaro:
Thank you very much @vbekiaris

PR #800, opened 2023-07-17
Clarify SQL type assignment for numeric literals [HZ-1748]

burakgok:
Closes hazelcast/hazelcast#22785


Serdaro:
Is this update valid for all 5.x doc series?

burakgok:
It applies to all versions since 4.2.

Issue #805, opened 2023-07-20
Docs: Feedback for Start a Local Cluster in Docker


JackPGreen:
There is an inconsistency in the `docker run` command between the [getting started documentation](https://docs.hazelcast.com/hazelcast/5.3/getting-started/get-started-docker#step-2-start-a-local-member) and the information on [Docker hub](https://hub.docker.com/r/hazelcast/hazelcast).

The former is missing the `-e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5701`, and the latter explicitly states:
`supplying a custom HZ_NETWORK_PUBLICADDRESS is critical for autodiscovery. Otherwise, Hazelcast will bind to Docker's internal ports.`

Discovered here - https://github.com/hazelcast/hazelcast/issues/24900 - Further discussion - https://hazelcast.slack.com/archives/C01JU7ZJYGP/p1689853817396409

PR #806, opened 2023-07-20
Update getting started Docker run command

JackPGreen:
Add `HZ_NETWORK_PUBLICADDRESS` to Docker run command

Fixes https://github.com/hazelcast/hz-docs/issues/805

### DIFF
--- a/docs/modules/getting-started/pages/get-started-docker.adoc
+++ b/docs/modules/getting-started/pages/get-started-docker.adoc
@@ -32,6 +32,7 @@ docker run \
     -it \
     --network hazelcast-network \
     --rm \
+    -e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5701 \
     -e HZ_CLUSTERNAME=hello-world \
     -p 5701:5701 hazelcast/hazelcast:{full-version}
 ----
--- a/docs/modules/getting-started/pages/get-started-docker.adoc
+++ b/docs/modules/getting-started/pages/get-started-docker.adoc
@@ -26,6 +26,8 @@ Docker networks make it easier for you to connect to your cluster and add other 
 
 . To start a member, execute the Docker `run` command.
 +
+include::deploy:partial$docker-ip-warning.adoc[]
++
 Replace the `<host_ip>` placeholder with the IP address of your Docker host.
 +
 [source,shell,subs="attributes+"]
--- a/docs/modules/getting-started/pages/get-started-docker.adoc
+++ b/docs/modules/getting-started/pages/get-started-docker.adoc
@@ -656,7 +656,7 @@ NOTE: Each container must expose port 5701 from a different port on the host mac
 ----
 docker run \
     --name my-second-member --network hazelcast-network \
-    -e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5701 \
+    -e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5702 \
     -e HZ_CLUSTERNAME=hello-world \
     -p 5702:5701 hazelcast/hazelcast:{full-version}
 ----
@@ -669,7 +669,7 @@ Replace the `<host_ip>` placeholder with the IP address of your member's Docker 
 ----
 docker run \
     --name my-third-member --network hazelcast-network \
-    -e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5701 \
+    -e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5703 \
     -e HZ_CLUSTERNAME=hello-world \
     -p 5703:5701 hazelcast/hazelcast:{full-version}
 ----
--- a/docs/modules/getting-started/pages/get-started-docker.adoc
+++ b/docs/modules/getting-started/pages/get-started-docker.adoc
@@ -26,6 +26,8 @@ Docker networks make it easier for you to connect to your cluster and add other 
 
 . To start a member, execute the Docker `run` command.
 +
+Replace the `<host_ip>` placeholder with the IP address of your member's Docker container.
++
 [source,shell,subs="attributes+"]
 ----
 docker run \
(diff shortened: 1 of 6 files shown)


docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -644,22 +648,28 @@ In this step, you add two more members to your cluster.
...
 +
+Replace the `<host_ip>` placeholder with the IP address of your member's Docker container.
++
 NOTE: Each container must expose port 5701 from a different port on the host machine. On this member, port 5701 is exposed to port 5702 on the host machine.
 +
 [source,shell,subs="attributes+"]
 ----
 docker run \
     --name my-second-member --network hazelcast-network \
+    -e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5701 \
```
frant-hartm, line 659: Shouldn't this match the `5702` in the `-p` parameter below?

frant-hartm (changes requested):
Let's discuss how to approach this.

Ideally, I would like to see
- same recommendation in all places where we document how to start a docker image
- when we use the `-p` parameters we need to add the `HZ_NETWORK_PUBLICADDRESS` env variable, but the `-p` parameter is not needed on Linux, user just needs to use the container address instead of localhost, it is probably enough to have the parameter everywhere and have a note for the Linux users, they will work out the rest
- using `-p 5701:5701` is actually bad advice - it exposes the port on all interfaces, e.g. making it available from wifi, very risky e.g. on a conference wifi setup, this should bind to localhost only, cc @kwart

JackPGreen:
> Ideally, I would like to see

Thanks for your feedback :)

> same recommendation in all places where we document how to start a docker image

I'm not sure what you mean by this? I do agree the ~~duplication~~ triplication is the cause of this issue.

> when we use the -p parameters we need to add the HZ_NETWORK_PUBLICADDRESS env variable, but the -p parameter is not needed on Linux, user just needs to use the container address instead of localhost, it is probably enough to have the parameter everywhere and have a note for the Linux users, they will work out the rest

I *personally* think that it's easier to have one command that works on any platform - especially in a getting started document - but I was trying to address the inconsistency rather than improve the readability of the document.

> using -p 5701:5701 is actually bad advice - it exposes the port on all interfaces, e.g. making it available from wifi, very risky e.g. on a conference wifi setup, this should bind to localhost only

If this is the case, should the [Docker hub page](https://hub.docker.com/r/hazelcast/hazelcast) also be updated?

angelasimms (changes requested):
Hi Jack, Could you also make sure that you add the appropriate release labels to the PR. It looks like this content is in all versions back to version 5.0.

frant-hartm:
We have multiple places in docs with `docker run hazelcast/hazelcast....`. We should have the same command in all places. I agree it's simpler to have the same command everywhere. I would just add a note to the getting started section that on Linux you can instead do something simpler.

So each `docker run ` should have 
- `-e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5701`
- `-p 5701:5701` or maybe `-p 127.0.0.1:5701:5701`

unless there is a reason for a different setup (e.g. the deploying with docker page is different).

The Docker hub page (hazelcast-docker readme) should be updated as well.

docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -26,18 +26,22 @@ Docker networks make it easier for you to connect to your cluster and add other
...
 +
+Replace the `<host_ip>` placeholder with the IP address of your member's Docker container.
++
 [source,shell,subs="attributes+"]
 ----
 docker run \
     -it \
     --network hazelcast-network \
     --rm \
+    -e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5701 \
```
ldziedziul, line 37: I would mention why we need this setting or point the documentation

ldziedziul:
> * using `-p 5701:5701` is actually bad advice - it exposes the port on all interfaces, e.g. making it available from wifi, very risky e.g. on a conference wifi setup, this should bind to localhost only, cc @kwart

but it's just a tutorial I would keep it simple, instead we can mention it's a production config

kwart (changes requested):
Both, @frant-hartm and @ldziedziul are right. Exposing the port to the public has security implications, but we should keep the commands as simple as possible.

I suggest adding a warning Notice at the beginning of the tutorial describing the implications. I.e. Other computers on the network can connect to the exposed port and
* manipulate data in the cluster,
* execute jobs in the running member containers;
* manage the cluster (e.g. shut down members)

JackPGreen:
> Hi Jack, Could you also make sure that you add the appropriate release labels to the PR? It looks like this content is in all versions back to version 5.0. Thanks again.

Done, thanks :)

JackPGreen:
> We have multiple places in docs with `docker run hazelcast/hazelcast....`. We should have the same command in all places. I agree it's simpler to have the same command everywhere. 

I did try changing everywhere but it got a bit silly - e.g. the [logging examples](https://docs.hazelcast.com/hazelcast/5.3/maintain-cluster/logging#logging-pattern) don't need it.
So I've just added it everywhere a `HZ_CLUSTERNAME` is used as that seems somewhere logical another member might be used as well.

JackPGreen:
> I would just add a note to the getting started section that on Linux you can instead do something simpler.

I've not done this - I think for consistency you'd need to add it to every `docker run` reference, there's no guarantee a user will start at `Getting started` - and I think it's easier to have simpler common instructions.

Maybe a compromise can be some slightly more technical description in in the `Deploying` section that talks about what `-p` actually does, the security implications and also that it's not needed on Linux? But I think that's probably beyond the scope of this change (and also I don't know enough to write it...)

> Both, @frant-hartm and @ldziedziul are right. Exposing the port to the public has security implications, but we should keep the commands as simple as possible.
> 
> I suggest adding a warning Notice at the beginning of the tutorial describing the implications. I.e. Other computers on the network can connect to the exposed port and
> 
> * manipulate data in the cluster,
> * execute jobs in the running member containers;
> * manage the cluster (e.g. shut down members)

Thanks - I've added something, not sure if it's what you had in mind.

kwart (approved):

frant-hartm (approved):

PR #810, opened 2023-07-25
DOCS-601 Caution message for SQL catalog persistence

angelasimms:
Added Caution admonition to warn users that SQL catalog persistence is permanently enabled.

### DIFF
--- a/docs/modules/storage/pages/configuring-persistence.adoc
+++ b/docs/modules/storage/pages/configuring-persistence.adoc
@@ -1266,6 +1266,8 @@ config.addMapConfig(mapConfig);
 
 Use the following configuration to enable the cluster-wide persistence of SQL metadata to disk. With persistence enabled, SQL mappings, data connections, and views are retained after cluster restarts.
 
+CAUTION: Once persistence of SQL metadata is enabled on a cluster, you cannot disable it.
+
 [tabs] 
 ==== 
 XML:: 


docs/modules/storage/pages/configuring-persistence.adoc
```diff
@@ -1266,6 +1266,8 @@ config.addMapConfig(mapConfig);
 
 Use the following configuration to enable the cluster-wide persistence of SQL metadata to disk. With persistence enabled, SQL mappings, data connections, and views are retained after cluster restarts.
 
+CAUTION: Once persistence of SQL metadata is enabled on a cluster it is _permanently enabled_. You cannot disable it.
```
k-jamroz, line 1269: do we have a similar warning for IMaps somewhere?

angelasimms: Not at the moment. Can we raise this under a separate PR? @k-jamroz?

k-jamroz: SQL persistence is available from 5.3, but IMap persistence is much older

angelasimms: Ok. We can agree the messaging here then and I'll raise another PR for that change. In the meantime, @vbekiaris, is the CAUTION message above accurate.

vbekiaris: I 'll describe what users cannot do with a persistent `IMap` / `ICache` and let you @angelasimms decide the appropriate wording :) 

I have a 3 cluster member + two persistent `IMap`s. At some point I decide that I want to change configuration and switch off persistence for one of the two persistent `IMap`s. I shut down my cluster, change configuration and restart the cluster. At the point, member startup will fail due to discrepancy between configuration stored by persistence engine vs configuration provided to member.

In order to change a persistent `IMap` to non-persistent, I need to shutdown my cluster, change configuration and clean all persistent data from disk (so effectively erase all my persistent data).

angelasimms: Thanks @vbekiaris. I will raise another PR for this change.

k-jamroz: >so effectively erase all my persistent data

In practice this means that you cannot disable persistence for IMap. It is effectively the same as recreating the cluster with altered config.

PR #853, opened 2023-09-22
Drafting the ExtendedMapEntry Content.

Serdaro:
As per https://hazelcast.atlassian.net/browse/DOCS-645.

### DIFF
--- a/docs/modules/computing/pages/entry-processor.adoc
+++ b/docs/modules/computing/pages/entry-processor.adoc
@@ -105,10 +105,8 @@ The following is an example snippet where the entries are incremented by one and
 ExtendedMapEntry<Integer, Integer> entry = (ExtendedMapEntry<Integer, Integer>) e;
 int newValue = entry.getValue() + 1;
 if (ttlSeconds > 0) {
-    entry.setValue(newValue, ttlSeconds, TimeUnit.SECONDS);
-    } else {
-    entry.setValue(newValue);
-    }
+// set TTL to 10 seconds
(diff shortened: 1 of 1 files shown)


docs/modules/computing/pages/entry-processor.adoc
```diff
@@ -95,6 +95,25 @@ You may also return `null` from the `EntryProcessor#getBackupProcessor` method.
...
+=== Updating Entry TTLs
+
+You can update the time-to-live (TTL) duration and value of existing map entries in your entry processor application, using the `ExtendedMapEntry` interface.
+The following is an example snippet where the entries are incremented by one and provided with a new TTL.
+
+[source,java]
+----
+ExtendedMapEntry<Integer, Integer> entry = (ExtendedMapEntry<Integer, Integer>) e;
+int newValue = entry.getValue() + 1;
+if (ttlSeconds > 0) {
```
vbekiaris, line 107: since there is no `ttlSeconds` field mentioned anywhere before we can:
- either present the full class definition from `EntryProcessor`'s [javadoc](https://github.com/hazelcast/hazelcast/blob/7c3894b1474dba56648cf10ad8441615486fcdea/hazelcast/src/main/java/com/hazelcast/map/EntryProcessor.java#L66-L88) 
- or skip the `ttlSeconds` condition and just demonstrate set-with-ttl variant. 
```
...
int newValue = entry.getValue() + 1;
// set TTL to 10 seconds
entry.setValue(newValue, 10, TimeUnit.SECONDS); 
```

PR #863, opened 2023-10-02
Clarify that external connectors should use some other means to protect access [HZ-2991]

frant-hartm:


kwart (changes requested):
I suggest rephrasing the warnings slightly, to make them more focused on external systems protection.

kwart (approved):

PR #867, opened 2023-10-11
Simplifying supported JDK content.

Serdaro:
* No need to mention "latest and LTS" since we are saying "we support the following JDK versions".
* Removed the JDK 8 warning since this version (`main`) of content belongs to 5.4.
* Removed "may not be compatible" information. If there is a clear replacement for "some JVMs", I will we happy to re-add it with specific information.


kwart (approved):
:+1: Simple and clear.

PR #869, opened 2023-10-11
Adding missing configuration options for Scheduled Executor Service.

Serdaro:
@vbekiaris , @JamesHazelcast .
I've added the missing configuration entries (capacity-policy and merge-policy). However, I need your input of:
* a more clarifying description for the capacity-policy (couldn't understand what exactly the policy is)
* the recommended values for pool-size
* the equivalent configuration with Java member API in https://github.com/hazelcast/hz-docs/blob/main/docs/modules/ROOT/examples/distributedcomputing/ScheduledExecutorConfiguration.java especially for the merge-policy.

### DIFF
--- a/docs/modules/ROOT/examples/distributedcomputing/ScheduledExecutorConfiguration.java
+++ b/docs/modules/ROOT/examples/distributedcomputing/ScheduledExecutorConfiguration.java
@@ -8,10 +8,12 @@ public class ScheduledExecutorConfiguration {
         //tag::sec[]
         Config config = new Config();
         config.getScheduledExecutorConfig( "myScheduledExecSvc" )
-                .setPoolSize ( 16 )
-                .setCapacity( 100 )
-                .setDurability( 1 )
-                .setSplitBrainProtectionName( "splitbrainprotectionname" );
+              .setPoolSize ( 16 )
+              .setCapacityPolicy( ScheduledExecutorConfig.CapacityPolicy.PER_PARTITION )
+              .setCapacity( 100 )
+              .setDurability( 1 )
+              .setMergePolicyConfig( new MergePolicyConfig("com.hazelcast.spi.merge.PassThroughMergePolicy", 110) )
+              .setSplitBrainProtectionName( "splitbrainprotectionname" );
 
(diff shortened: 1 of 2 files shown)


JamesHazelcast:
Hi @Serdaro, thanks for taking care of this - to respond to your requests:

> * a more clarifying description for the capacity-policy (couldn't understand what exactly the policy is)
The `CapacityPolicy` essentially defines how the defined `capacity` integer value is applied (either per-node, or per-partition) - previously it was only per-partition (and as a result, we also need to update the `capacity` attribute description). I think [the JavaDocs](https://github.com/hazelcast/hazelcast/blob/7c3894b1474dba56648cf10ad8441615486fcdea/hazelcast/src/main/java/com/hazelcast/config/ScheduledExecutorConfig.java#L386C1-L402) does a pretty good job of summarising the policies and recommendations on which to pick and I'd recommend essentially copying this information. The [JavaDocs for capacity](https://github.com/hazelcast/hazelcast/blob/7c3894b1474dba56648cf10ad8441615486fcdea/hazelcast/src/main/java/com/hazelcast/config/ScheduledExecutorConfig.java#L186-L197) are also useful for referencing.

> * the recommended values for pool-size
The default value is 16 if not specified, but for actual recommendations I'll leave that to @vbekiaris.

> * the equivalent configuration with Java member API in https://github.com/hazelcast/hz-docs/blob/main/docs/modules/ROOT/examples/distributedcomputing/ScheduledExecutorConfiguration.java especially for the merge-policy.
Apologies, but to clarify - do you want me to add the Java code for supplying a merge-policy to that example linked, or create a new one specifically for merge-policy? If the latter, should I also include the other config options?

Serdaro:
@JamesHazelcast , thanks a lot. For the code sample, the former would be awesome.

JamesHazelcast:
I've updated the example in https://github.com/hazelcast/hz-docs/pull/869/commits/fb4d7f82dc0b42d28e738609ac858be5976672cb @Serdaro - I also added the `CapacityPolicy` definition, please feel free to remove this line if you think it's not necessary 👍

Serdaro:
Awesomeness, thank you.

JamesHazelcast (approved):
Thanks for the improvements @Serdaro!

PR #886, opened 2023-10-27
Fix WAN advanced network YML example indentation

JamesHazelcast:
The example is currently incorrect as `wan-replication` and `map` are listed as child nodes of `advanced-network` when they should be child nodes of `hazelcast`.

The `endpoint` reference is also incorrect in the current example, and fixed with this PR.


rebekah-lawrence (approved):
All looks good to me

PR #891, opened 2023-11-07
Adding RN for Platform 5.3.6

Serdaro:


docs/modules/release-notes/pages/5-3-6.adoc
```diff
@@ -0,0 +1,6 @@
+= 5.3.6 Release Notes
+
+== Fixes
+
+* Fixed a WAN replication failure - when using maps with WAN replication enabled, and `BINARY` or `NATIVE` in-memory formats for the entries, WAN replication was failing when an entry is updated with the same key/value pair.
```
vbekiaris, line 5: there are 2 additional requirements for the issue to manifest:
- cluster members lack the serialization information to deserialize map values
- entries have TTL set

but I 'm not sure how to word this in a concise way. @JamesHazelcast any ideas?

JamesHazelcast: I think adding this on to the end will help clarify: `and cluster members lack serialization information` resulting in:

> Fixed a WAN replication failure - when using maps with WAN replication enabled, and `BINARY` or `NATIVE` in-memory formats for the entries, WAN replication was failing when an entry is updated with the same key/value pair and cluster members lack serialization information.

Entries do not actually need to have TTL enabled to trigger the offending code block, so I think this additional detail is sufficient 👍

frant-hartm (approved):

PR #897, opened 2023-11-17
Improve RingbufferStore documentation

JackPGreen:
- Reword the configuration heading to match that of `List`, `Set`, `Multimap` etc
- Fix formatting of property reference

### DIFF
--- a/docs/modules/data-structures/pages/ringbuffer.adoc
+++ b/docs/modules/data-structures/pages/ringbuffer.adoc
@@ -295,9 +295,9 @@ hazelcast:
 ----
 ====
 
-The following are the explanations for the Ringbuffer store configuration elements:
+The following are the configuration elements and their descriptions:
 
-* `class-name: Name of the Ringbuffer store factory class.
+* `class-name`: Name of the Ringbuffer store factory class.
 
 [[configuring-ringbuffer-in-memory-format]]
 == Configuring Ringbuffer In-Memory Format


Serdaro (approved):

docs/modules/data-structures/pages/ringbuffer.adoc
```diff
@@ -295,9 +295,9 @@ hazelcast:
 ----
 ====
 
-The following are the explanations for the Ringbuffer store configuration elements:
+The following are the configuration elements and their descriptions:
```
rebekah-lawrence, line 298: The configuration elements are as follows:

rebekah-lawrence (changes requested):

docs/modules/data-structures/pages/ringbuffer.adoc
JackPGreen: @rebekah-lawrence I don't mind this change, but I used the same wording as in [List](https://docs.hazelcast.com/hazelcast/latest/data-structures/list#configuring-list), [Set](https://docs.hazelcast.com/hazelcast/latest/data-structures/set#configuring-set) & [MultiMap](https://docs.hazelcast.com/hazelcast/latest/data-structures/multimap#configuring-multimap) - so maybe they should all be changed separately?

rebekah-lawrence: If it matches the other sections, just leave it as-is. Consistency is king ;-)
We can re-visit later when reassessing the docs as a whole.
Thanks!

rebekah-lawrence (approved):

PR #914, opened 2023-11-28
Add note about compact serialization not supported

srknzl:
There is a precondition about compact serialization that should hold all the time in a cluster: before putting any compact data into the cluster, the schema of it has to be replicated in the cluster. Otherwise, some errors like SchemaNotFoundException can happen. In phase 1, (the current version) we don't migrate compact schemas. The user can migrate their compact data but it can lead to errors. 

I mention that compact serialization data migration is not supported due to this in our phase 1 documentation

I think we should backport this to 5.3, help me do that pls.


rebekah-lawrence (approved):
LGTM

PR #926, opened 2023-12-11
Update WAN publisher description

ethurma:


docs/modules/wan/pages/wan.adoc
```diff
@@ -38,9 +38,9 @@ events and replicates them actively to other clusters.
 * Passive cluster: The user updates performed on this cluster are not replicated to other clusters. In other words,
 this cluster can be seen as the "target" cluster which is capable of receiving, applying and possibly forwarding
 WAN events from other ("active") clusters. It does not generate any WAN update events because of user interaction.
-* WAN publisher: A publisher is a sink for WAN events and an implementation of `WanReplicationPublisher`.
-Most often, this is a single, entire target Hazelcast cluster but you can also define custom publishers which
-may transmit WAN events to other systems such as messaging queues, Kafka or even persist events on disk.
+* WAN publisher: A publisher acts as a sink for local WAN events and pushes them to a target. The target is usually
+another Hazelcast cluster, but can be customized to be another system, such as a messaging queue, Kafka or even persisted
```
rebekah-lawrence, line 42: The target is usually another Hazelcast cluster; however, you could also use a different target such as a message queue, Kafka, or saving data to persistent storage.

rebekah-lawrence: Questions not related to the update, but as I looked and saw it....

1. Can we link to any docs on WanReplicationPublisher? If not, what does it add to tell me that this is an implementation of it?
2. Can we link to any information on using different targets? Possibly the [Configuring WAN Replication ](https://docs.hazelcast.com/hazelcast/5.3/wan/defining-wan-replication)page, and/or [Advanced Features](https://docs.hazelcast.com/hazelcast/5.3/wan/advanced-features)?

docs/modules/wan/pages/wan.adoc
```diff
@@ -50,4 +50,4 @@ clusters or multiple external systems.
 identifier can then be used to control the behavior of a WAN replication publisher while the source/active cluster is
 running. For instance, you can use this identifier in combination with the WAN replication scheme name to pause, stop or
 resume WAN replication for that specific publisher. Or, you can trigger synchronization with a specific target cluster
-and so on.
\ No newline at end of file
+and so on.
```
rebekah-lawrence, line 53: I'd suggest recasting the previous two sentences (the second is really only a phrase and the last three words that were deleted and re-added don't make sense in it); something like:

For example, you could use the WAN Publisher ID as follows: 
+
* In combination with the WAN replication scheme name to pause, stop, or resume WAN replication for a specific publisher
* To trigger synchronization with a specific target cluster

rebekah-lawrence (changes requested):

docs/modules/wan/pages/wan.adoc
ethurma: Thanks @rebekah-lawrence, your tweaks read better. 

@JamesHazelcast or @vbekiaris may have some suggestions about informative links.

vbekiaris: "different targets" like JMS etc are examples of custom `WanPublisher` implementations a user can implement, it's not something we support out of the box. To be honest, I have never seen a user/customer implement their own `WanPublisher`.
The `WanReplicationPublisher` name is probably outdated - the interface is called `WanPublisher` and it's what users should implement to provide their custom WAN publisher implementation. Probably the audience here would be a java developer, so I think it would be most appropriate to link to `WanPublisher` [javadoc](https://docs.hazelcast.org/docs/latest-dev/javadoc/com/hazelcast/wan/WanPublisher.html)

PR #936, opened 2023-12-18
DMT Phase 2 Doc Updates

rebekah-lawrence:
After the conversations following the previous release, I have restructured the DMT section.

Instead of getting everyone to review this restructure AND the v2 updates, I have separated out this stage to ensure we end up with useful, useable information once the updates are included

Feel free to correct any incorrect/unclear information, but for this round, please concentrate on the structure.


frant-hartm (changes requested):
This is a huge improvement, thanks!

docs/modules/migrate/pages/dmt-connection-config.adoc
```diff
@@ -0,0 +1,128 @@
...
+* For an on-premise target cluster, update the `cluster-members` field to match the addresses of the cluster members
+* For a cloud target cluster, including a {hazelcast-cloud} cluster, update the network information 
++
+For a public cloud cluster, refer to the documentation for the cloud provider for the required network details. 
++
+For {hazelcast-cloud}, you must update the network section as follows:
++
+----
+hazelcast-client:
+  :
```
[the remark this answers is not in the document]
rebekah-lawrence: The `:` on a line generally communicates further content that are/might be present in the code; it has been standard doc practice for a long time

However, if it s going to cause confusion, we'll need a different approach. We could: 

* Include the omitted lines (which may also muddy the water of what is actually being shown)
* Only include the changed lines with no additional context (I'd recommend keeping context if possible)
* Use a different notation (ideas welcome)
* Explicitly state at the start that lines containing `:` identify places where other lines might exist

docs/modules/migrate/pages/dmt-migration.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+Once downloaded, extract the DMT package to a location in your folder structure. The DMT package includes the following:
+
+* The DMT
++
+You must use the DMT binary that is suitable for your machine. The DMT binaries are in the _bin_ folder of the extracted DMT package with filenames in `dmt_[platform]_[arch]` format.  Use the ``arm64`` binary for ARM, and the `amd64` binary for Intel.
+
+* A custom Hazelcast distribution that creates a migration cluster and runs the data migration service
+* Example configuration files
+* Example connection configuration file, which is used to connect to the migration client
```
srknzl, line 49:
```suggestion
* Example connection configuration file, which is used to connect to the migration cluster
```

docs/modules/migrate/pages/dmt-migration.adoc
```diff
@@ -0,0 +1,55 @@
...
+
+== Limited Migration Cluster License
```
srknzl, line 51: no longer true we removed this

docs/modules/migrate/pages/dmt-view-details.adoc
```diff
@@ -0,0 +1,30 @@
+= View Migration Details
+:description: Once the migration completes, you can view the details of the migration in a report, DMT log file, or the results IMap created on the target cluster.
+
+{description}
+
+The details of the migration are recorded as follows:
+
+* Migration report
++
+This is written to the _migration$$_$$report$$_$$[migration_id].txt_ file in the directory used when running the `dmt` command.
```
srknzl, line 10: I don't know these details by my heart. I think @yuce should review this PR as well to be sure that frontend related stuff. If nothing changed from the old content though, we should be fine. cc @rebekah-lawrence

docs/modules/migrate/pages/dmt-run.adoc
```diff
@@ -0,0 +1,53 @@
...
+. The DMT will attempt to connect to the migration cluster indefinitely. This means that it can appear to hang if unable to connect. To avoid this, you can set a timeout for the connection attempt using the `--timeout` flag.  For further information on the `--timeout` flag, refer to the https://docs.hazelcast.com/clc/5.3.5/configuration#clc-configuration-with-command-line-parameters[CLC Configuration with Command-line Parameters^] section of the Hazelcast CLC documentation. 
+. On MacOS, you might need to allow the `dmt*` binary to run. If the command is rejected, go to the *Privacy & Security* settings on your device and update them to allow the binary. After updating the settings, retry the command, and select *Open* when prompted
+====
+
+== Track the Migration
+
+You can use the DMT `status` command to track the migration. 
+
+For further information on the available DMT commands, see the xref:migrate:dmt-command-reference.adoc[DMT Command Reference].
```
srknzl, line 53: I think it's not in this PR but we need to update the command reference. --config flag for dmt migration cluster config to connect to migration cluster is missing, also --yes flag etc.(not sure if we should add it but good to do )

docs/modules/migrate/pages/dmt-connection-config.adoc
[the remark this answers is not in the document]
rebekah-lawrence: After discussion with the team, standard notation will be three periods (`...`) on the line instead of the colon (`:`)

docs/modules/migrate/pages/dmt-view-details.adoc
rebekah-lawrence: Invited Yuce to review; this will form part of the updates for v2 work (which I am starting now)

docs/modules/migrate/pages/dmt-run.adoc
rebekah-lawrence: I didn't do the Command Reference last time, so it didn't occur to me!

@srknzl do you want me to pick this up, or would you prefer that the team continues to maintain this page?

frant-hartm (approved):

docs/modules/migrate/pages/dmt-start-migration-cluster.adoc
```diff
@@ -0,0 +1,27 @@
...
+HZ_NETWORK_PORT_PORT=5702 HZ_CLUSTERNAME=migration ./bin/hz start 
+----
+
+If the specified port is available, the cluster starts on that port. Otherwise, Hazelcast tries to find a free port as described in the xref:clusters:network-configuration.adoc#port[Port] section of the Networking topic. You can confirm the port used by the cluster in the logs displayed in your terminal.
+
+You can find the _migration.yaml_ file in the root folder of the DMT download package. If your logs show that the cluster starts on a different port to that specified in this file, you must update the `address` field to match the port number used.
+
+DMT uses this configuration file to connect to the migration cluster when running the migration.
+
+NOTE: The _migration.yaml_ file uses the same configuration options as the Hazelcast CLC. For further information on the options, refer to the https://docs.hazelcast.com/clc/latest/clc-config[Hazelcast CLC documentation^].
```
yuce, line 23: This section may be updated in case named CLC configuration is to be used.

docs/modules/migrate/pages/dmt-start-migration-cluster.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+== Disconnect Client Applications
+
+Don't forget to disconnect all client applications from your source cluster before you continue.
```
yuce, line 27: Is this necessary?

docs/modules/migrate/pages/dmt-run.adoc
yuce: IMO it would be the best if the user created a named configuration using `clc config add` instead of providing the configuration path directly. In that case they don't need to provide the configuration with each command. CLC will display a configuration selector if necessary.

docs/modules/migrate/pages/dmt-view-data.adoc
```diff
@@ -0,0 +1,30 @@
...
+clc -c target.yaml map size --name my-map
+----
++
+The output is similar to the following
++
+[source,shell]
+----
+1000
+OK
+----
```
yuce, line 30: The new output is:
```
------
 Size 
------
 1000 
------
    OK Returned 1 row(s).
```

yuce (changes requested):
Great update, thanks Rebekah!
Just a few ideas...

docs/modules/migrate/pages/dmt-run.adoc
srknzl: > IMO it would be the best if the user created a named configuration using clc config add instead of providing the configuration path directly. In that case they don't need to provide the configuration with each command. CLC will display a configuration selector if necessary.

yeah that'll be very good. What if they don't have clc though? In any case, I think we should have --config flag in the command reference because it's available for use and I don't see a point in hiding it. 

@rebekah-lawrence @yuce and I can do that, thank you

srknzl: @rebekah-lawrence note: one possible update I see is that adding a note about what @yuce said, basically they can create a named clc config so that they don't have to specify "migration.json" config path all the time via --config flag. 

@yuce is it possible to call "dmt config add"?

docs/modules/migrate/pages/dmt-start-migration-cluster.adoc
rebekah-lawrence: Debatable ;-)

I chose to include it for the following reasons:

* It is mentioned as a step in the High-level Process section, so it ought to be reflected in the How-to topics for completeness and to reduce cognitive load for people while reading
* When people use docs, they might not be thinking as they follow the steps and could inadvertently miss a step that would otherwise be in their mind
* It was raised by Hendy when she commented on the original single-page version of this information

Bottom line is, it is short enough, and clear enough, that people can easily skip it if they don't need the reminder/full picture; but, it provides enough support to cover the first two reasons given above for including it

docs/modules/migrate/pages/dmt-run.adoc
rebekah-lawrence: I have added a note, called it `dmt config add` for now along with a x-ref to the DMT Command Reference assuming the `dmt config` command will be added there. Can easily change if this isn't implemented as dmt-specific.

docs/modules/migrate/pages/dmt-start-migration-cluster.adoc
rebekah-lawrence: Not changed for the moment, will revisit when approach is confirmed

docs/modules/migrate/pages/dmt-run.adoc
yuce: @srknzl 
DMT has the same `config` commands with CLC.

srknzl: Rebekah gave a link to dmt command reference. We should update DMT command reference and state that clc commands like "clc config" are also usable with dmt

srknzl: https://github.com/hazelcast/hz-docs/pull/981

docs/modules/migrate/pages/data-migration-tool.adoc
```diff
@@ -17,6 +17,7 @@ When using the DMT, bear the following in mind:
 ** The source cluster must be running and populated with required data
 ** The application must be using the source cluster successfully before migration
 ** Before starting the migration, remember to shutdown any applications that are using the source cluster
+** If migrating Replicated map metadata, you must add the custom JAR to your class path and restart the cluster. If this is not done, the Replicated map metadata is not migrated. For further information on adding the custom JAR to your classpath, refer to the https://docs.oracle.com/javase/6/docs/technotes/tools/windows/classpath.html[Oracle documentation,window=_blank].
```
srknzl, line 20: We'll add an option to enable/disable replicated map metadata migration via a flag, will let you know when that happens

docs/modules/migrate/pages/dmt-connection-config.adoc
```diff
@@ -38,6 +38,33 @@ If you have multiple data structures use a new line for each map name, or use wi
...
+
+You can migrate the following CP data structures:
+
+* xref:data-structures:iatomiclong.adoc[]
+* xref:data-structures:iatomicreference.adoc[]
+* xref:data-structures:icountdownlatch.adoc[]
+
+If a CP data structure has no data, it is not migrated. For example, where the `set()` method has not been called on an existing `AtomicReference`. This is because such data structures are created lazily and do not exist in the cluster until used. This avoids the creation of a server-side resource with null data.
+
+To configure the  migration of CP data structures, you must update the _config.json_ file located in the _migration_config_ folder.
```
srknzl, line 53:
```suggestion
To configure the migration of CP data structures, you must update the _config.json_ file located in the _migration_config_ folder.
```

we'll not have an option about cp structures I think.  so we can move this sentence to a different place and say:

To configure the data migration, you must update the _config.json_ file located in the _migration_config_ folder.

srknzl: It might be reasonable to keep this PR open until we finish the config.json addition then. Not to have to open a new pr

docs/modules/migrate/pages/dmt-migration.adoc
```diff
@@ -1,8 +1,10 @@
 = Migrate Your Data
-:description: The DMT migrates your data for maps and replicated maps only, and can be run on Mac, Linux, and Windows Operating Systems.
+:description: The DMT migrates your data for maps and replicated maps, compact serialization schemas, and- if configured - CP data structures. It can be run on Mac, Linux, and Windows Operating Systems.
```
srknzl, line 2: for now, the cp configuration is only txt files where you list their names. we can try to allow them not requiring a cluster restart if they give us a list of names instead of a wildcard. then there will be a config.json option for that. 

but as I know, we'll not do it. am i right? @frant-hartm

srknzl: this means for now, we can delete "if configured" part above. if there wont be a extra config for cp structures in config.json

docs/modules/migrate/pages/dmt-run.adoc
```diff
@@ -51,3 +51,15 @@ To run the migration, complete the following steps:
...
+
+The DMT migrates the data structures one-by-one.
+
+If the migration of a data structure is not successful, the DMT does the following:
+
+* Retains any data structures that have been successfully migrated already
+* Reverts the unsuccessful data structure migration
+* Cancels the migration of all remaining data structures
+
+In this situation, you can restart the migration [information to be confirmed].
```
srknzl, line 65: yes they can restart the migration. currently the way to do that is just restarting the migration with the same command. existing data structures in the target cluster will be skipped.

docs/modules/migrate/pages/dmt-connection-config.adoc
```diff
@@ -0,0 +1,169 @@
...
+
+. Navigate to the folder containing the extracted DMT package
+. Open the _migration_config/data/imap_names.txt_ and/or the _migration_config/data/replicated_map_names.txt_ file in your favorite editor
+. Update the file content to match the names of your maps
++
+If you have multiple data structures use a new line for each map name, or use wildcards to select multiple data structures using a single entry. For further information on using wildcards, see the xref:configuration:using-wildcards.adoc[Using Wildcards] topic.
+
+. Save the files
+
+By default, Replicated Map metadata is migrated. However, you must add the custom JAR to your class path and restart the cluster before starting the migration. If this is not done, the Replicated map metadata is not migrated. For further information on adding the custom JAR to your classpath, refer to the https://docs.oracle.com/javase/6/docs/technotes/tools/windows/classpath.html[Oracle documentation,window=_blank].
```
srknzl, line 41: For ReplicatedMap with metadata, we should add a note. "It's not possible to migrate replicated map metadata to target versions below 5.3.7." This kind of limitation does not exist for CP. They can be migrated to a target cluster with version 5.2.z for instance.

srknzl: Another thing, we should clarify that the custom jar is for the source cluster and not the target cluster. Also we can give a link to the new page I added in this PR https://github.com/hazelcast/hz-docs/pull/1042

docs/modules/migrate/pages/data-migration-tool.adoc
```diff
@@ -17,18 +17,26 @@ When using the DMT, bear the following in mind:
...
 ** The application must be using the source cluster successfully before migration
 ** Before starting the migration, remember to shutdown any applications that are using the source cluster
-** If migrating Replicated map metadata, you must add the custom JAR to your class path and restart the cluster. If this is not done, the Replicated map metadata is not migrated. For further information on adding the custom JAR to your classpath, refer to the https://docs.oracle.com/javase/6/docs/technotes/tools/windows/classpath.html[Oracle documentation,window=_blank].
+** For certain versions, you must add the provided extension JAR to your class path and restart the cluster if you want to migrate Replicated map metadata and/or CP data structures. For further information on the affected versions and use of the extension JAR, see the xref:migrate:dmt-source-cluster-extension.adoc[] topic
++
+If needed, but not done, Replicated map metadata and/or CP data structure are not migrated. For further information on adding the extension JAR to your classpath, refer to the https://docs.oracle.com/javase/6/docs/technotes/tools/windows/classpath.html[Oracle documentation,window=_blank].
```
srknzl, line 22: almost true. 

If the user does not set the "migrateReplicatedMapMetadata" option to false explicitly, the migration cluster will try to migrate replicated map metadata by default and there will be an error & migration fail about the missing API.

The CP migration is explicit by the user, we only migrate CP migrations if there are entries in the CP files in  migration_config/data folder, used to select which CP structures to migrate. If they select CP data structures but the API is absent on the source side, we'll again fail the migration.

docs/modules/migrate/pages/data-migration-tool.adoc
```diff
@@ -17,18 +17,26 @@ When using the DMT, bear the following in mind:
...
 
 * *Target Cluster:*
 
 ** The target cluster must be set up
 +
 Any populated data structures that already exist on the target cluster are not migrated; however, if the existing data structure is empty, it is migrated.
 
-** The target cluster must be running on either Enterprise Edition, version 5.3.2 or later, or {hazelcast-cloud}
+** The target cluster must be running on oneof the following:
```
srknzl, line 30:
```suggestion
** The target cluster must be running on one of the following:
```

docs/modules/migrate/pages/data-migration-tool.adoc
```diff
@@ -17,18 +17,26 @@ When using the DMT, bear the following in mind:
...
+
+*** If you don't want to migrate ReplicatedMap metadata, Enterprise Edition version 5.3.2 or later. You can disable migration of Replicated map metadata as described in xref:migrate:dmt-connection-config.adoc#replicatedmap-metadata-migration[ReplicatedMap Metadata Migration]
```
srknzl, line 32: 5.3.2 is for the IMap metadata migration, (which can't be disabled by an option), not for ReplicatedMap

docs/modules/migrate/pages/data-migration-tool.adoc
rebekah-lawrence: Great spot, thank you!
Amended in [https://github.com/hazelcast/hz-docs/pull/936/commits/58c391dd98438c1981eac72bf13fe637ebb02c3d](url)

docs/modules/migrate/pages/data-migration-tool.adoc
rebekah-lawrence: I have amended and expanded this description in the latest update.

docs/modules/migrate/pages/data-migration-tool.adoc
```diff
@@ -17,17 +17,25 @@ When using the DMT, bear the following in mind:
...
 +
-If needed, but not done, Replicated map metadata and/or CP data structure are not migrated. For further information on adding the extension JAR to your classpath, refer to the https://docs.oracle.com/javase/6/docs/technotes/tools/windows/classpath.html[Oracle documentation,window=_blank].
+For further information on the affected versions and use of the extension JAR, see the xref:migrate:dmt-source-cluster-extension.adoc[] topic. For further information on adding the extension JAR to your classpath, refer to the https://docs.oracle.com/javase/6/docs/technotes/tools/windows/classpath.html[Oracle documentation,window=_blank].
++
+If needed, but not done, the following occurs:
+
+*** The migration cluster attempts to migrate Replicated map metadata unless you have disabled the migration of Replicated map metadata in your configuration, as described in xref:migrate:dmt-connection-config.adoc#replicatedmap-metadata-migration[ReplicatedMap Metadata Migration]. 
++
+If you have not added the extension JAR or disabled the migration of Replicated map metadata, the migration fails with an error about the missing API contained in the extension JAR.
```
srknzl, line 28: If you have not added the extension JAR or disabled the migration of Replicated map metadata, the migration fails with an error about the missing API contained in the extension JAR.

--> 

If you have not added the extension JAR and have not disabled the migration of Replicated map metadata, the migration fails with an error about the missing API contained in the extension JAR.

docs/modules/migrate/pages/data-migration-tool.adoc
```diff
@@ -17,17 +17,25 @@ When using the DMT, bear the following in mind:
...
+
+*** CP data structure are not migrated.
```
srknzl, line 30: since we say if needed, lets change here as well. an extension jar is needed if they want to migrate cp structures. 

So, "if needed" means they selected cp structures to migrate.

if they selected cp structures to migrate, but haven't added the extension jar to the classpath the behaviour is not "CP data structure are not migrated."

it's failing the migration with an error about the missing CP APIs that are in the extension jar or 5.3.7+ cluster version.

docs/modules/migrate/pages/dmt-dev-test.adoc
```diff
@@ -0,0 +1,115 @@
...
+To complete the migration, follow the xref:migrate:dmt-migration.adoc#high-level-process[high-level process], omitting steps 3 and 5. 
+
+NOTE: You can skip steps 3, shutting down connected applications, and 5, updating your client configuration, as the test on a Development cluster only demonstrates how the data is migrated.
+
+== Start the Source Cluster
+
+You can start your source cluster in either of the following ways:
+
+* Using Docker, as described below. This is the recommended method
+* Downloading a version package and following the IMDG https://docs.hazelcast.com/imdg/4.2/getting-started[Quickstart]. For example, we used https://hazelcast.com/open-source-projects/downloads/archives/#hazelcast-imdg[Hazelcast IMDG version 4.2.7] for the samples provided in this section
```
srknzl, line 22: do we still use the word IMDG? Is it because the page is only available on 4.2? Is there a platform version of it? newer, for 5.x

docs/modules/migrate/pages/dmt-migration.adoc
```diff
@@ -0,0 +1,49 @@
+= Migrate Your Data
+:description: The DMT migrates your data for maps and replicated maps, compact serialization schemas, and CP data structures. It can be run on Mac, Linux, and Windows Operating Systems.
+
+{description}
+
+NOTE: To migrate Replicated metadata, you must include the provided JAR in your class path. The source cluster must be restarted after the addition of the JAR.
```
srknzl, line 6:
```suggestion
NOTE: To migrate Replicated metadata and/or CP structures, you must include the provided JAR in your class path. The source cluster must be restarted after the addition of the JAR.
```

srknzl: Also, maybe remove this note as it may be confusing.


or we can give a link to the related place without removing it.

docs/modules/migrate/pages/dmt-connection-config.adoc
```diff
@@ -0,0 +1,174 @@
...
+. Update the `cluster-name` field to match the name of your source cluster
+. Update the `cluster-members` field to match the addresses of the cluster members
+. Save the file
+
+=== Update Data Structures
+
+To update the data structure information, complete the following steps:
+
+. Navigate to the folder containing the extracted DMT package
+. Open the _migration_config/data/imap_names.txt_ and/or the _migration_config/data/replicated_map_names.txt_ file in your favorite editor
```
srknzl, line 34: also there are new files for cp structures. see https://github.com/hazelcast/hazelcast-mono/tree/0778d11cc407c3677c3a166308aecf5c01a5c418/distribution/src/root/migration_config/data

srknzl: I would just remove the file names from here as there are 5 of them now

docs/modules/migrate/pages/dmt-connection-config.adoc
```diff
@@ -0,0 +1,174 @@
...
+To configure the migration of CP data structures, you must update the _migration_config.yaml_ file located in the _migration_config_ folder.
+
+The CP data structure configuration options are described in the following table:
+
+[%header,cols="1,2"]
+|===
+|Option |Description
+
+|`"matchHidden"`
+|Defines whether to include maps with names that start with a double underscore ($$__$$) in the migration when a wildcard (`*`) is included in the `run` command. For example, Hazelcast uses some internal maps with names starting with $$__$$ such as $$__$$sql.catalog.
```
srknzl, line 77:
```suggestion
|Defines whether to include data structures with names that start with a double underscore ($$__$$) in the migration when a wildcard (`*`) is included in the `run` command. For example, Hazelcast uses some internal maps with names starting with $$__$$ such as $$__$$sql.catalog.
```

This option is a generic option that can be used with any data structure migration. If there are other places, with this description, we should also update them

srknzl: Also, I believe we should mention "migration config" instead of "run" command to be more clear with our description.

srknzl: Plus, since this option is generic, I don't think we should mention in the CP structures section, maybe in "Update Data Structures" section above.

docs/modules/migrate/pages/dmt-view-data.adoc
```diff
@@ -0,0 +1,32 @@
+= Verify the Migrated Data
+:description: To ensure that the migration includes all data, you can verify the size of the map in the target cluster.
```
srknzl, line 2: with the addition of CP migration support, we may want to update this page with more generic "data structure" migration validation. it can be either

- map
- replicated map 
- cp structures

for map and replicated map, size check should be enough. for cp structures, checking the existence of such a cp structure with the provided name should be good enough

srknzl: As I know, clc does not support replicated map, atomic reference, count down latch.

We can direct users to management center for those.

docs/modules/migrate/pages/dmt-connection-config.adoc
srknzl: For reference, this is my description of the config option in the dmt package: https://github.com/hazelcast/hazelcast-mono/blob/d12b08f726de9dff17f2e06ac5188e9f52109d0b/distribution/src/root/migration_config/migration_config.yaml#L3 

I added this description.

srknzl (commented):
looks good in general, however still can be improved imo. ty for the changes in advance

frant-hartm (approved):

docs/modules/migrate/pages/dmt-dev-test.adoc
rebekah-lawrence: The example was to show the migration path - people can substitute their product/version and still follow the steps. I chose an older version because (a) it seems to be a common migration use case and (b) to highlight the steps required to update multiple versions.

As you say, it is fine, so I won't amend this.

docs/modules/migrate/pages/dmt-view-data.adoc
```diff
@@ -1,20 +1,22 @@
...
++
 To use the Hazelcast Management Center, you can use either of the following methods:
 +
 ** Check the target map size, as described in the https://docs.hazelcast.com/management-center/5.3/data-structures/map[Maps^] section of the Hazelcast Management Center documentation
 ** Check the map entries, as described in the https://docs.hazelcast.com/management-center/5.3/data-structures/map#map-browser[Exploring Map Entries^] section of the Hazelcast Management Center documentation
 
 * Use Hazelcast CLC 
 +
-To use Hazelcast CLC to verify the migrated map size, enter the following command in your terminal:
+To use Hazelcast CLC to verify the migrated map or replicated map size, enter the following command in your terminal:
```
srknzl, line 19: not sure if this is correct. "--map" should be for the imaps. let me confirm it with Yuce

srknzl: confirmed that clc does not support replicated map

rebekah-lawrence: [e3b6c3b](https://github.com/hazelcast/hz-docs/pull/936/commits/e3b6c3b17d71b48118a3cf9c9e2e607c215ed112)

docs/modules/migrate/pages/dmt-view-data.adoc
rebekah-lawrence: [e3b6c3b](https://github.com/hazelcast/hz-docs/pull/936/commits/e3b6c3b17d71b48118a3cf9c9e2e607c215ed112)

srknzl (approved):

srknzl:
Sorry I realized now. CLC has atomicreference command now it seems: https://github.com/hazelcast/clc/pull/69

Can we update the related "verify your data migration" part accordingly? @rebekah-lawrence

srknzl:
@yuce what's the proper way of checking if a atomic reference exists using clc? `clc atomicreference is-null` ? However, if an atomic reference does not exist, we don't want to create it via this command. Is it available to do that?

yuce:
@srknzl Any `atomicreference` operation would create a proxy if it doesn't already exist, just like the client libraries.
I would assume creating the proxy also creates an atomic reference on the server-side.
Currently, there's no way to check whether an atomic reference exists on the server, using CLC.
To my knowledge, it's not possible to easily do that without CLC as well.

docs/modules/migrate/pages/data-migration-tool.adoc
```diff
@@ -1,428 +1,86 @@
...
+** Indicative timings from our testing:
 +
-[source,shell]
-----
-for /l %x in (1, 1, 1000) do clc -c source.yaml map --name my-map set key-%x value-%x --quiet
-----
---
+[NOTE]
 ====
+These timings will be affected by entry and data structure counts as follows:
```
sumnerib, line 18: Maybe an additional disclaimer around this part that mentions that these timings are just examples, and that individual migration times will depend on various factors including environmental constraints (i.e. network, hardware resources, etc.). Otherwise LGTM.

sumnerib (changes requested):

srknzl:
> @srknzl Any atomicreference operation would create a proxy if it doesn't already exist, just like the client libraries.
I would assume creating the proxy also creates an atomic reference on the server-side.
Currently, there's no way to check whether an atomic reference exists on the server, using CLC.
To my knowledge, it's not possible to easily do that without CLC as well.

Hmm, as far as I recall, there was a note in our documentation (or somewhere in confluence) about CP structures not being created until some value is written to them. So suppose if I do: "return getAtomicReference("foo").get() == null". Will this create atomic reference on the cluster @arodionov ?

arodionov:
Yes, the proxy on the client side will not automatically create the object on the server side.
So,
        `IAtomicReference<String> ref = cpSubsystem.getAtomicReference("iRef");`
will not create a server-side object.
But 
        `String val = ref.get();`
will create it with a `null` value.

arodionov:
Yes, the source code for this operation is the following:
```
        V atomicValue = atomicValues.get(key);
        if (atomicValue == null) {
            atomicValue = newAtomicValue(groupId, name, null);
            atomicValues.put(key, atomicValue);
        }
        return atomicValue;
```

rebekah-lawrence:
I have updated the 5.3.7 requirement to 5.3.8 - @frant-hartm could you take another final look and make sure I have correctly understood

srknzl (approved):
the version change is lgtm

rebekah-lawrence:
Adding a comment to ensure this is not automatically Closed before it is confirmed that this can be merged.

PR #940, opened 2023-12-23
Fix a typo

neshkeev:
There is a similar command in the Read Data from Memory section which lacks this dollar sign, so I removed it to make the instructions consistent


Serdaro:
Thanks a lot @neshkeev

Serdaro (approved):

PR #945, opened 2024-01-02
Add Documentation for terminate cp session from MC [MC-1719][MC-2553]

aberatcan:
docs update for terminate cp session from MC


rebekah-lawrence (approved):
LGTM

PR #949, opened 2024-01-10
Additional docs for `EXTERNAL NAME` in mapping fields [HZ-4026]

Fly-Style:


docs/modules/sql/pages/mapping-to-maps.adoc
```diff
@@ -41,6 +41,39 @@ OPTIONS (
 )
 ```
 
+If key or value is a complex type, you need to enumerate all required columns in the
```
rebekah-lawrence, line 44:
```suggestion
If key or value is a complex type, you must enumerate all required columns in the
```

docs/modules/sql/pages/mapping-to-maps.adoc
```diff
@@ -41,6 +41,39 @@ OPTIONS (
...
+    this INT
+)
+TYPE IMap
+OPTIONS (
+    'keyFormat'='java',
+    'valueFormat'='int'
+)
+```
+
+NOTE: You cannot rename `this` field, it is a technical limitation.
```
rebekah-lawrence, line 75:
```suggestion
NOTE: You must not rename the `this` field.
```

rebekah-lawrence: I suggest removing the statement about it being a technical limitation as that level of detail is not required or useful. If you prefer to keep it, you'll need more information on the limitation.

Cannot means they are not able to - if that is the case, you can leave it as-is; otherwise, we are telling them not to do it, so must not makes more sense.

Fly-Style: > Cannot means they are not able to - if that is the case

Exactly. It is technically impossible.

rebekah-lawrence: Thanks. In that case, you are quite right to stick with cannot.

rebekah-lawrence (approved):

k-jamroz (changes requested):
This is not true for Java mapping - see the jira issue.
The only limitation is that if you want to explicitly map columns from complex key, then external name with `__key` prefix must be used. Without prefix very confusing error message is produced.

Fly-Style:
I still think, that mentioned approach is more preferrable and natural to SQL.

k-jamroz:
> I still think, that the mentioned approach (explicit mapping of each field to column) is more preferrable and natural to SQL.

It might be preferable or not, but definietly is not required. I don't think we should strongly recommend one way or another (explicit or implicit) - both have pros and cons and both are correct.

docs/modules/sql/pages/mapping-to-maps.adoc
```diff
@@ -41,6 +41,42 @@ OPTIONS (
 )
 ```
 
+If key or value is a Portable or Compact complex type, you must enumerate all required columns
```
k-jamroz, line 44: > you must enumerate all required columns

Not true for Java, did not test Compact but also should work.

For Portable (and maybe for Compact?) explicit enumeration is only required when the schema of the Portable (maybe also Compact?) object cannot be determined, eg. because it has not yet been registered.

k-jamroz: And, as discussed today, for Portable it may be insufficient to only register PortableFactory and this only does not register `ClassDefinitions`

PR #953, opened 2024-01-14
DMT ReplicatedMap note update [dont review yet]

srknzl:
https://hazelcast.atlassian.net/browse/HZ-3948

I am not sure if extra changes will be required or not. The implementation PRs are not merged. Also https://github.com/hazelcast/hz-docs/pull/936/files should be merged before this PR. Therefore, let's keep the PR open for now.

There'll be a new option added for selecting if replicated map metadata should be migrated or not (false will mean current approach, no metadata migration). After the option is added, we should add the usage of it here.


srknzl:
closing because @rebekah-lawrence started to work on this

PR #954, opened 2024-01-15
Removing Java 11 and updating related references [HZ-4062][HZ-4168]

Serdaro:


docs/modules/ROOT/pages/capacity-planning.adoc
```diff
@@ -585,8 +585,8 @@ of your Hazelcast data pipeline.
 
 === Upgrade Your JDK
 
-Hazelcast does not support JDK 8, so if you are still on JDK 8, please upgrade it.
-The garbage collectors of JDK 11+ have been improved to achieve much higher throughputs. 
+Hazelcast does not support JDK 8 and 11, so if you are still on these, please upgrade it.
```
rebekah-lawrence, line 588:
```suggestion
Hazelcast does not support JDK 8 or 11. If you are using either of these version, upgrade to version 17 or later.
```

rebekah-lawrence (approved):
One minor suggestion for your consideration

docs/modules/ROOT/pages/capacity-planning.adoc
JackPGreen:
```suggestion
Newer JDK releases often contain performance improvements, consider upgrading to 
xref:deploy:versioning-compatibility.adoc#supported-java-virtual-machines[a newer supported version].
```

nit - shouldn't it just be <=11? We don't list 7, 9, 10 etc.
Does it make sense to list the supported JDKs on this page anyway? We have a dedicated page for that.
_(not sure the syntax of my suggestion is right)_

JackPGreen (changes requested):
If Java 17 is the minimum, should we have any references to "if your on Java 17+" in the documentation? Or are we catering to non-`5.4` users with the latest documentation as well?

docs/modules/ROOT/pages/capacity-planning.adoc
Serdaro: I don't think it is necessary to list them here since, as you said, we have a dedicated page (linked from here)?

docs/modules/ROOT/pages/capacity-planning.adoc
```diff
@@ -585,8 +585,9 @@ of your Hazelcast data pipeline.
 
 === Upgrade Your JDK
 
-Hazelcast does not support JDK 8, so if you are still on JDK 8, please upgrade it.
-The garbage collectors of JDK 11+ have been improved to achieve much higher throughputs. 
+Hazelcast does not support JDK 11 and older versions. Newer JDK releases often contain performance improvements, consider upgrading to
```
frant-hartm, line 588: Maybe instead of mentioning a specific version we could mention our policy of supporting last 2 LTS versions, and list what exactly we support in `versioning-compatibility.adoc`

frant-hartm (approved):

docs/modules/ROOT/pages/capacity-planning.adoc
```diff
@@ -585,8 +585,9 @@ of your Hazelcast data pipeline.
 
 === Upgrade Your JDK
 
-Hazelcast does not support JDK 8, so if you are still on JDK 8, please upgrade it.
-The garbage collectors of JDK 11+ have been improved to achieve much higher throughputs. 
+Hazelcast supports the **last two Java LTS versions** available at the time of its release. Newer JDK releases often contain performance improvements, consider upgrading to
```
oliverhowell, line 588:
```suggestion
Hazelcast supports the **current** Java LTS version and the **previous** Java LTS version, as available at the time of the Hazelcast release. Newer JDK releases often contain performance improvements, consider upgrading to 
```

docs/modules/ROOT/pages/capacity-planning.adoc
JackPGreen: I'm not sure it's a good idea to put our policy buried on one of these pages. If Java 23 LTS comes out tomorrow, we won't be supporting it until the next major release of Hazelcast - whereas if it says we support the latest 2 LTS releases, it would be implied that 23 was supported today.

docs/modules/ROOT/pages/capacity-planning.adoc
oliverhowell: Alternate version:
```suggestion
Hazelcast supports the last two Java LTS versions (i.e. the current Java LTS version **and** the previous Java LTS version) available at the time of the Hazelcast release. Newer JDK releases often contain performance improvements, consider upgrading to 
```

Serdaro: > I'm not sure it's a good idea to put our policy buried on one of these pages. If Java 23 LTS comes out tomorrow, we won't be supporting it until the next major release of Hazelcast - whereas if it says we support the latest 2 LTS releases, it would be implied that 23 was supported today.

You're right @JackPGreen.

@frant-hartm , @oliverhowell , I'd remove the first sentence in this paragraph, and stick with a phrase something like ```Newer JDK releases often contain performance improvements, so we strongly recommend you to upgrade to a newer JDK version. See the xref:deploy:versioning-compatibility.adoc#supported-java-virtual-machines[supported JDKs]. Note that the garbage collectors of the newer JDK versions have been improved to achieve much higher throughputs.```, where in the given link we list the exact supported JDK versions for that versioned documentation content.

Screenshot 2024-02-04 at 12 01 59 PM

oliverhowell: > If Java 23 LTS comes out tomorrow, we won't be supporting it until the next major release of Hazelcast

Yeah, if that's the case we need to be explicit about versions we support and agree with the general approach of linking to https://docs.hazelcast.com/hazelcast/latest/deploy/versioning-compatibility

docs/modules/ROOT/pages/production-checklist.adoc
```diff
@@ -70,7 +70,7 @@ General recommendations:
 * GC logs should be enabled
 * Minimum and maximum heap size should be equal
 
-For Java 11+:
+For Java 17+:
```
JackPGreen, line 73: Nit - not sure what value this adds? There's no non-Java 17 option for the user, or in the documentation

docs/modules/cluster-performance/pages/performance-tuning.adoc
```diff
@@ -147,7 +147,7 @@ We xref:ROOT:production-checklist.adoc#jvm-recommendations[recommend] enabling
 GC logs to allow troubleshooting if performance problems occur. To enable GC
 logging, use the following JVM arguments:
 
-*Java 11 (and above)*
+*Java 17 (and above)*
```
JackPGreen, line 150: Nit - not sure what value this adds? There's no non-Java 17 option for the user, or in the documentation

Also we said "Java 17+" earlier. Is it "+" or "above"?

docs/modules/clusters/pages/network-configuration.adoc
```diff
@@ -1021,7 +1021,7 @@ You can specify TCP keep-alive options for each socket for use cases such as pre
 
 These options:
 
-* work with JDK 11 or newer, and recent JDK 8 distributions based on OpenJDK
+* work with JDK 17 or newer, and JDK 17 distributions based on OpenJDK
```
JackPGreen, line 1024: I'm not sure this makes sense any more - originally it was saying there were non-OpenJDK, JDK8 distributions where these options won't work.
But if they all work on JDK11, and the minimum supported is 17, then I think these lines can just go?

docs/modules/getting-started/pages/get-started-binary.adoc
```diff
@@ -11,11 +11,11 @@ To complete this tutorial, you need the following:
 |===
 |Prerequisites|Useful resources
 
-|JDK 11+
+|JDK 17+
```
JackPGreen, line 14: To save upgrading this page in future - could we **just** put the supported JDK link instead? And skip versions here alltogether?

docs/modules/getting-started/pages/install-hazelcast.adoc
```diff
@@ -277,7 +277,7 @@ endif::[]
 === Using Modular Java
 
 // tag::modular[]
-If you're using JDK 11 or later, you can use Hazelcast as a module in the http://openjdk.java.net/projects/jigsaw/[Java Platform Module System] (JPMS).
+If you're using JDK 17 or later, you can use Hazelcast as a module in the http://openjdk.java.net/projects/jigsaw/[Java Platform Module System] (JPMS).
```
JackPGreen, line 280:
```suggestion
You can use Hazelcast as a module in the http://openjdk.java.net/projects/jigsaw/[Java Platform Module System] (JPMS).
```

Can assume user **must** be using >11 and 17 is the minimum requirement.

docs/modules/getting-started/pages/get-started-java.adoc
```diff
@@ -11,7 +11,7 @@ To complete this tutorial, you need the following:
 |===
 |Prerequisites|Useful resources
 
-|JDK 11+
```
JackPGreen, line 14: As before:

> To save upgrading this page in future - could we **just** put the supported JDK link instead? And skip versions here alltogether?

docs/modules/migrate/pages/upgrading-from-imdg-3.adoc
```diff
@@ -618,7 +618,7 @@ Hazelcast defaults to use OpenSSL when:
 
 * when you use xref:security:tls-ssl.adoc[TLS/SSL] and Hazelcast detects some
 xref:security:integrating-openssl.adoc[OpenSSL] capabilities
-* the Java version is less than 11
+* the Java version is less than 17
```
JackPGreen, line 621: This page is about migrating from IMDG 3 to Hazelcast 5.0.

- if it's really about 5.0, then it shouldn't change with the new 5.4 docs
- if it's actually about 5.x (which is _also_ used on this page) then I'm a bit confused, as you can't use <17 with 5.4?

docs/modules/pipelines/pages/stream-processing-client.adoc
```diff
@@ -14,11 +14,11 @@ To complete this tutorial, you need the following:
 |A full Hazelcast distribution (Docker or Binary)
 |xref:getting-started:install-hazelcast.adoc[Install Hazelcast]
 
-|JDK 11+
+|JDK 17+
 |
 xref:deploy:supported-jvms.adoc[Supported Java Virtual Machines]
 
-xref:deploy:running-in-modular-java.adoc[Using JDK 11 with Modular Java]
+xref:deploy:running-in-modular-java.adoc[Using JDK 17 with Modular Java]
```
JackPGreen, line 21: As before:

> To save upgrading this page in future - could we **just** put the supported JDK link instead? And skip versions here alltogether?

docs/modules/security/pages/integrating-openssl.adoc
```diff
@@ -27,7 +27,7 @@ Check the https://netty.io/wiki/forked-tomcat-native.html[netty-tcnative^] page
 for installation details.
 
 NOTE: If the Java version is
-less than 11 and OpenSSL capabilities are detected
+less than 17 and OpenSSL capabilities are detected
```
JackPGreen, line 30: As before - how can a user use Java <17?

JackPGreen (changes requested):
I'm sorry I'm being picky.

Serdaro:
@JackPGreen , first of all, thank you for the pickiness. I addressed the comments.
For the OpenSSL ones, I think we need to do some rewording on the 5.x series, until 5.4.

JackPGreen (approved):

PR #955, opened 2024-01-15
Sql reflection config

shultseva:
[(https://hazelcast.atlassian.net/browse/HZ-3574)](https://hazelcast.atlassian.net/browse/HZ-3574)

new parameters have been introduced that limit the ability to create objects through reflection.

### DIFF
--- a/docs/modules/sql/pages/sql-reflection-configuration.adoc
+++ b/docs/modules/sql/pages/sql-reflection-configuration.adoc
@@ -19,7 +19,9 @@ When objects are constructed through reflection, the following filtering rules a
 
 When reflection fails, a `SecurityException` is thrown.
 
-By default, class names or package names that have these prefixed are allowed:
+By default, reflection restriction filter is empty and all class names or package names are allowed.
+
+If the reflection restriction filter is not empty, class names or package names with the specified prefixes are automatically added to the whitelist by default:
 
 * `java`
 * `com.hazelcast.`
--- a/docs/modules/sql/pages/sql-reflection-configuration.adoc
+++ b/docs/modules/sql/pages/sql-reflection-configuration.adoc
@@ -7,12 +7,12 @@ Clusters can be configured to restrict the capability of creating objects throug
 
 To prevent Hazelcast from constructing objects of arbitrary classes through reflection, you have the option to permit or deny the reflection for Java classes based on class name, package name, or prefix using the java-reflection-filter parameter.
 
-The following filtering rules are used when the objects are constructed through reflection:
+When objects are constructed through reflection, the following filtering rules apply:
 
 * When the `whitelist` option is empty:
 ** If the object's class name or package name is blacklisted, reflection fails.
 ** Otherwise, reflection is allowed.
-* When the `whitelist` option is provided:
+* When the `whitelist` option is populated:
 ** If the object's class name or package name is blacklisted, reflection fails.
 ** If the object's class name or package name is whitelisted, reflection is allowed
 ** Otherwise, reflection fails.
--- a/docs/modules/sql/pages/sql-reflection-configuration.adoc
+++ b/docs/modules/sql/pages/sql-reflection-configuration.adoc
@@ -5,7 +5,7 @@ When you have set object format equals to `java`, you can configure clusters to 
 
 == Disallowing Reflection of Untrusted Java Classes
 
-To prevent Hazelcast from constructing objects of arbitrary classes through reflection, you have the option to permit or deny the reflection for Java classes based on class name, package name, or prefix using the `java-reflection-filter` parameter.
+To prevent Hazelcast from constructing objects of arbitrary classes through reflection, you can allow or deny reflection for Java classes based on class name, package name, or prefix using the `java-reflection-filter` parameter.
 
 When objects are constructed through reflection, the following filtering rules apply:
 
--- a/docs/modules/sql/pages/sql-reflection-configuration.adoc
+++ b/docs/modules/sql/pages/sql-reflection-configuration.adoc
@@ -25,7 +25,7 @@ By default, class names or package names that have these prefixed are allowed:
 * `com.hazelcast.`
 * `[` (for primitives and arrays)
 
-If you do not want to allow these default prefixes, set the `defaults-disabled` attribute to true.
+If you do not want to allow these default prefixes, set the `defaults-disabled` attribute to `true`.
 
 [tabs]
 ====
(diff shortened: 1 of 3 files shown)


k-jamroz:
IMO worth mentioning the new setting in 
https://docs.hazelcast.com/hazelcast/5.3/secure-cluster/hardening-recommendations

docs/modules/sql/pages/sql-reflection-configuration.adoc
```diff
@@ -0,0 +1,96 @@
+= Configuring Reflection
+:description: Clusters can be configured to restrict the capability of creating objects through reflection during SQL execution.
```
k-jamroz, line 2: we should be clear here that this applies only to `format=java`. Eg. zero-config compact serialization reflection will not be affected (secured)

docs/modules/sql/pages/sql-reflection-configuration.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+Clusters can be configured to restrict the capability of creating objects through reflection during SQL execution.
+
+== Disallowing Reflection of Untrusted Java Classes
+
+To prevent Hazelcast from constructing objects of arbitrary classes through reflection, you have the option to permit or deny the reflection for Java classes based on class name, package name, or prefix using the java-reflection-filter parameter.
```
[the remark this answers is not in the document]
rebekah-lawrence: I'd amend Krzysztof's comment slightly; instead of "you have the option to permit...", use "you can allow...."

docs/modules/sql/pages/sql-reflection-configuration.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+The following filtering rules are used when the objects are constructed through reflection:
```
rebekah-lawrence, line 10: When objects are constructed through reflection, the following filtering rules apply:

docs/modules/sql/pages/sql-reflection-configuration.adoc
```diff
@@ -0,0 +1,96 @@
...
+
+* When the `whitelist` option is empty:
+** If the object's class name or package name is blacklisted, reflection fails.
+** Otherwise, reflection is allowed.
+* When the `whitelist` option is provided:
```
rebekah-lawrence, line 15: consider "populated" instead of "provided"

rebekah-lawrence (changes requested):

docs/modules/sql/pages/sql-reflection-configuration.adoc
shultseva: dc36063bc06516632afd7329dbc04854ec1e3174

docs/modules/sql/pages/sql-reflection-configuration.adoc
shultseva: b78116e8d8a8fd65fb1046577fd0c5dd0fa0e5a2

docs/modules/sql/pages/sql-reflection-configuration.adoc
shultseva: b78116e8d8a8fd65fb1046577fd0c5dd0fa0e5a2

docs/modules/sql/pages/sql-reflection-configuration.adoc
k-jamroz: it is unclear that this applies to SQL mappings and UDTs. IMO writing only `format=java` is too generic.

shultseva: a376deb8e8a6aefeef679dd5796b743893931703

shultseva:
> IMO worth mentioning the new setting in https://docs.hazelcast.com/hazelcast/5.3/secure-cluster/hardening-recommendations

a376deb8e8a6aefeef679dd5796b743893931703

docs/modules/sql/pages/sql-reflection-configuration.adoc
shultseva, line 2:
```suggestion
:description: Clusters can be configured to restrict the capability of creating objects through reflection during SQL execution (applicable to INSERT/SINK/UPDATE operations).
```

shultseva: https://github.com/hazelcast/hz-docs/pull/955#discussion_r1453461783 ?

k-jamroz: this is tricky. Currently you cannot create UDT instance in SELECT projection (currently projections produce only `ROW` type), but in theory is it possible  and one day might be supported.
I think we can leave the description as is.

rebekah-lawrence (changes requested):

docs/modules/sql/pages/sql-reflection-configuration.adoc
```diff
@@ -0,0 +1,96 @@
...
+** If the object's class name or package name is blacklisted, reflection fails.
+** Otherwise, reflection is allowed.
+* When the `whitelist` option is populated:
+** If the object's class name or package name is blacklisted, reflection fails.
+** If the object's class name or package name is whitelisted, reflection is allowed
+** Otherwise, reflection fails.
+
+When reflection fails, a `SecurityException` is thrown.
+
+By default, class names or package names that have these prefixed are allowed:
```
k-jamroz, line 22: there are actually 2 "defaults":
- `SqlConfig.javaReflectionFilterConfig = null` -> all classes allowed
- `SqlConfig.javaReflectionFilterConfig = new JavaSerializationFilterConfig()` -> whitelist as described here

we should make this clear somehow or maybe change the implementation, but without enabling any filtering by default because it would be a breaking change

shultseva: The Java Reflection filter operates in a manner analogous to the Java serialization filter configuration.
I don't think having a different logic is a good idea.

I ll add more info to the doc.

shultseva: no, i am wrong

rebekah-lawrence (approved):

docs/modules/sql/pages/sql-reflection-configuration.adoc
k-jamroz: this is fine for java config but how does it map to XML and yaml?

shultseva: 5532bd50a327cbb9b384c33d1cbc6332c2192147

k-jamroz (approved):

PR #969, opened 2024-01-24
Document UDT improvements and breaking changes [HZ-3497] [HZ-3686]

burakgok:
Documents hazelcast/hazelcast#25269
Documents hazelcast/hazelcast#25935
Documents hazelcast/hazelcast#25954
Documents hazelcast/hazelcast#25977

Closes [HZ-3497](https://hazelcast.atlassian.net/browse/HZ-3497)
Closes [HZ-3686](https://hazelcast.atlassian.net/browse/HZ-3686)

[HZ-3497]: https://hazelcast.atlassian.net/browse/HZ-3497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HZ-3686]: https://hazelcast.atlassian.net/browse/HZ-3686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/modules/sql/pages/user-defined-types.adoc
+++ b/docs/modules/sql/pages/user-defined-types.adoc
@@ -8,6 +8,7 @@ User-Defined Types (also known as UDTs, nested types, or nested fields) allow yo
 - Type options override mapping options. Where both the mapping and type define a schema (Portable class definition, Compact schema, Java class or Avro schema) and the corresponding mapping field is `__key` or `this`, the schema defined by the type is used. This means that the mapping does not need to define a schema if the type defines one.
 - If the type does not specify a schema, it is resolved from the parent structure, which may be a mapping or another type.
 - Type fields are optional and if not specified, they are resolved from the class/schema when the type is used in a `CREATE MAPPING` statement. You can check the resolved fields using a `GET_DDL` query. The field resolution feature has the following limitations.
+ ** Once the fields are resolved, they are not updated when the underlying class/schema changes or when the type is used in another mapping.
  ** For Portable, Compact, or Avro formats, schema fields cannot be complex (`PORTABLE`, `COMPACT`, or `RECORD`). Otherwise, an exception will be thrown once the type is used in a mapping.
  ** For Portable format, portable IDs (`portableFactoryId`, `portableClassId` and `portableVersion`) must be unique within the cluster. Otherwise, deserialization issues can occur.
  ** For Portable format, the field resolution relies on the Portable class being already registered (by internal means, such as adding a configuration for Portable serialization) on the member executing the `CREATE MAPPING` command.
@@ -116,7 +117,7 @@ CREATE MAPPING users (
 ----
 
 == Support for Cycles
-When creating a UDT, the existence of referenced types are only verified when the type is used in a `CREATE MAPPING` statement. This makes it possible to create cyclic types.
+When creating a UDT, the existence of referenced types is only verified when the type is used in a `CREATE MAPPING` statement. This makes it possible to create cyclic types.
 
 NOTE: Cyclic types are only supported for Java format. However, the support is limited only to querying. Inserting or updating with cyclic types is currently not supported.
 
--- a/docs/modules/sql/pages/user-defined-types.adoc
+++ b/docs/modules/sql/pages/user-defined-types.adoc
@@ -1,44 +1,31 @@
 = User-Defined Types in SQL
 
-User-Defined Types (also known as UDTs, nested types, or nested fields) is an experimental feature that allows you to create
- custom data types that can be referenced in the xref:sql:create-mapping.adoc[CREATE MAPPING statement].
+User-Defined Types (also known as UDTs, nested types, or nested fields) allows you to create custom data types that can be referenced in the xref:sql:create-mapping.adoc[CREATE MAPPING statement].
 
-UDTs are useful for creating and accessing hierarchical data structures, including simple cases of nested objects and
-more complex cases of fully/partially connected graphs of objects. See <<queryingExamples,examples>> for more information.
-
-WARNING: Support for UDTs is an experimental feature, and disabled by default.
-See <<enabling-udt-support, Enabling UDT support>> to learn how to enable it.
+UDTs are useful for creating and accessing hierarchical data structures, which can even be cyclic. See <<queryingExamples,examples>> for more information.
 
 == Feature Overview
 
-Due to the experimental nature of UDTs, the feature set is limited to the following:
+- All serialization formats (namely `portable`, `compact`, `java` and `avro`) are supported.
+- Cyclic definitions are only supported for Java-serialized structures. (Portable and Compact formats do not support cyclic schemas at all. Avro support is currently limited to acyclic schemas.) Note that `INSERT`/`UPDATE` queries are disabled for mappings that use cyclic structures.
+- Type options _always_ override mapping options. Specifically, if the mapping has only `__key` or `this` fields, and both the mapping and type define a schema (Portable class definition, Compact schema, Java class or Avro schema), the one defined by the type is used. This also means that, in such situations, as long as the type defines a schema, the mapping does not need to.
+- If type schema is not specified, it will be resolved from the parent structure, which may be a mapping or another type.
+- Type fields are optional and will be resolved from the schema if they are not specified. In that case, schema fields cannot be complex (`PORTABLE`/`COMPACT`/`RECORD`) for Portable/Compact/Avro formats. For Java format, complex fields are allowed and mapped to `OBJECT`. Type fields are resolved and materialized once the type is used in a `CREATE MAPPING` statement. The resolved fields can be checked by issuing a `GET_DDL` query after using the type in a mapping.
 
-- Support for three formats - `portable`, `compact` and `java` with varying level of sub-feature support
-- Instance and type-level cycles are only supported for Java types.
-- Type formats (referred to as type kind or kind hereafter) can not be mixed, both between mappings and types themselves.
-For example, you cannot use Java type in another Portable type or Portable mapping.
-- Limited support for instance and type-level cycles; only for Java types.
-Note that INSERT/UPDATE are disabled for mappings that use type hierarchies that contain cycles.
-- Limited support for `INSERT` and `UPDATE` queries for mappings using UDTs; if a `TYPE` hierarchy contains cycles,
-`INSERT` and `UPDATE` statements are disabled for the mapping.
-- Support for using UDT-based projections in both normal `SELECT` projection lists, `WHERE` filters and in `JOIN` conditions.
-
-== Enabling UDT Support
-You can enable UDTs by setting the `hazelcast.sql.experimental.custom.types.enabled` property to `true` in the member configuration.
+== Enabling Cycling Type Support
+You can enable cyclic types by setting the `hazelcast.sql.experimental.custom.cyclic.types.enabled` property to `true` in the member configuration.
 [tabs]
-==== 
-XML:: 
-+ 
--- 
+====
+XML::
++
 [source,xml]
 ----
 <hazelcast>
     <properties>
-        <property name="hazelcast.sql.experimental.custom.types.enabled">true</property>
+        <property name="hazelcast.sql.experimental.custom.cyclic.types.enabled">true</property>
     </properties>
 </hazelcast>
 ----
---
 
 YAML::
 +
@@ -46,7 +33,7 @@ YAML::
 ----
 hazelcast:
   properties:
-    hazelcast.sql.experimental.custom.types.enabled: true
+    hazelcast.sql.experimental.custom.cyclic.types.enabled: true
 ----
 
 Java::
@@ -54,7 +41,7 @@ Java::
 [source,java]
 ----
 final Config config = new Config();
-config.setProperty("hazelcast.sql.experimental.custom.types.enabled", "true");
+config.setProperty("hazelcast.sql.experimental.custom.cyclic.types.enabled", "true");
 ----
 ====
 
@@ -64,20 +51,26 @@ To create a new type, use the `CREATE TYPE` statement:
 
 [source,sql]
 ----
-CREATE [OR REPLACE] TYPE [IF NOT EXISTS] MyTypeName
-    [(colName colType, ...)]
-    OPTIONS (
-    'format'='{java|portable|compact}'
-    [, 'javaClass'='com.myPackage.MyJavaClass'] <1>
-    [, 'compactTypeName'='MyCompactRecordTypeName'] <2>
-    [, 'portableFactoryId'='123', 'portableClassId'='456', ['portableVersion'='789']] <3>
-    )
+CREATE [OR REPLACE] TYPE [IF NOT EXISTS] MyTypeName [(
+    colName colType,
+    ...
+)] [OPTIONS (
+    'javaClass'='com.mypackage.MyJavaClass' <1>
+  | 'compactTypeName'='MyCompactRecordName' <2>
+  | 'portableFactoryId'='123', 'portableClassId'='456', ['portableVersion'='789'] <3>
+  | 'avroSchema'='{"type":"record","name":"myType","fields":[{"name":"colName","type":"colType"},...]}' <4>
+)]
 ----
-<1> `java` format requires the `javaClass` option.
-<2> `portable` format requires `portableFactoryId` and `portableClassId` and optionally `portableVersion`.
-<3> `compact` format requires `compactTypeName` - this is not the name of the created type, but rather internal name of the Compact record type, used internally by the Compact Serialization format.
+<1> In `java` format, you can use `javaClass` option to override the type class with a subclass. Note that you must provide a concrete class for `INSERT` statement to work, which makes `javaClass` option mandatory when the original type class is abstract or interface.
+<2> In `portable` format, you can use `portableFactoryId`, `portableClassId` and `portableVersion` options to specify a portable ID for the type, which is only effective when the type is used for `__key` or `this` fields. `portableVersion` is optional, similarly to `CREATE MAPPING` statement. However, it is not possible to inherit some components from the parent, so portable ID components must be defined together. Incomplete definitions have the same effect as no definition.
+<3> In `compact` format, you can use `compactTypeName` option to specify the Compact record name. If unspecified, it is defaulted to `<TypeName>CompactType`.
+<4> In `avro` format, you can use `avroSchema` option to specify an inline Avro schema, which is only effective when the type is used for `__key` or `this` fields.
 
-NOTE: To reference another type, you must provide the column list. Otherwise, the column may be automatically resolved as an `OBJECT` type.
+NOTE: `EXTERNAL NAME` aliases are not supported for types, column names must have the same name as their corresponding Java class field or Portable/Compact/Avro schema field.
+
+NOTE: You can mix options that belong to different formats. When you create a mapping that references your UDT, the relevant options are used and the others are ignored, which makes it possible to use a UDT in multiple mappings having different serialization formats.
+
+NOTE: To reference another type, you must provide the column list. Otherwise, in Java format, the column will be resolved as `OBJECT` type; in Portable, Compact and Avro formats, an exception will be thrown once the type is used in a mapping.
 
 === Java Format notes
 For `java` format, if the column list is omitted, it will be automatically resolved from the corresponding
@@ -85,9 +78,7 @@ Java class. Note that the column list will only be extracted from the source cla
 if it has columns that are inherited from a superclass, these columns will not be resolved.
 
 === Portable Format notes
-When using the `portable` format, make sure that the `factoryId`, `classId`, and `version` tuples are unique within the cluster.
-Otherwise, deserialization issues can happen if the corresponding class IDs and factory IDs are registered in the client
-for serialization/deserialization.
+When using the `portable` format, make sure that (`factoryId`, `classId`, `version`) tuples are unique within the cluster. Otherwise, deserialization issues may happen.
 
 In addition, there is a rudimentary auto-resolution mechanism for column list. However, it is not recommended for use:
 it relies on the Portable class being already registered (through internal Portable means,
@@ -96,11 +87,7 @@ This mechanism is not reliable since this command will fail if the member that e
 the Portable class in question. Therefore, it is recommended to always specify the column list.
 
 === Support for Cycles
-Cycles between types are only supported for Java format however, the support is limited to querying only.
-If a Type hierarchy contains cycles, any mapping using any of these types (provided that type is not itself an acyclic branch)
-will have `INSERT` and `UPDATE` commands disabled.
-Additionally, support for cycles also means no validation for existence of custom types at the time of `CREATE TYPE` execution.
-**Type hierarchies are only verified for consistency upon actual use in `CREATE MAPPING`.**
+Cycles between types are only supported for Java format; however, the support is limited to querying only. **The existence of referenced types are only verified upon use in `CREATE MAPPING` statement.**
 
 === Replacing Types and Type Consistency
 Currently, there is a limitation on the replacement of existing types:
@@ -108,64 +95,7 @@ if the replaced type was already used in a mapping, you need to fully replace th
 to update its data type information using the `DROP MAPPING` and `CREATE MAPPING` statements.
 However, if the type hierarchy was not used in a mapping, any type in that hierarchy can be safely
 changed, and these changes will appear in the new mapping. This is because the links
-between types are symbolic (based on the name only), and they're only "materialized" once used in a mapping.
-
-=== CREATE TYPE Examples
-
-NOTE: `EXTERNAL NAME` aliases are not supported for types, column names have to have exact
-same name as their corresponding Java/Portable/Compact class fields.
-
-Java Type with auto-resolution for columns:
-
-[source,sql]
-----
-CREATE TYPE MyType OPTIONS (
-    'format'='java',
-    'javaClass'='com.example.MyJavaClass'
-)
-----
-
-Java type with explicit columns:
-
-[source,sql]
-----
-CREATE TYPE MyType (
-    id BIGINT,
-    name VARCHAR,
-    other MyOtherType
-) OPTIONS (
-    'format'='java',
-    'javaClass'='com.example.MyJavaClass'
-)
-----
-
-Portable Type:
-
-[source,sql]
-----
-CREATE TYPE MyPortableType (
-    id BIGINT,
-    name VARCHAR
-) OPTIONS (
-    'format'='java',
-    'portableFactoryId'='1',
-    'portableClassId'='1'
-    -- 'portableVersion'='0' - specified by default
-)
-----
-
-Compact Type:
-
-[source,sql]
-----
-CREATE TYPE MyCompactType (
-   id BIGINT,
-   name VARCHAR
-) OPTIONS (
-    'format'='java',
-    'compactTypeName'='MyCompactTypeInternalCompactNameExample',
-)
-----
+between types are symbolic (based on the name only), and they are only "materialized" once used in a mapping.
 
 ==== Creating Java Type Hierarchy with Cycles
 
@@ -200,7 +130,6 @@ CREATE TYPE AType (
     name VARCHAR,
     b BType
 ) OPTIONS (
-    'format'='java',
     'javaClass'='com.example.A'
 );
 
@@ -208,7 +137,6 @@ CREATE TYPE BType (
     name VARCHAR,
     c CType
 ) OPTIONS (
-    'format'='java',
     'javaClass'='com.example.B'
 );
 
@@ -216,22 +144,12 @@ CREATE TYPE CType (
     name VARCHAR,
     a AType
 ) OPTIONS (
-    'format'='java',
     'javaClass'='com.example.C'
 );
 ----
 
-== Creating Mappings with UDT Columns
-
-The syntax of the `CREATE MAPPING` statement is virtually unchanged, except now, UDT names can be used
-in the column type.
-
-NOTE: UDT columns must be explicitly declared as of UDT type in the column list, even if the underlying
-Java class of the column is registered as a backing Java class for an existing UDT.
-Otherwise, the column in question will be auto-resolved as `OBJECT`.
-
 === Java Class Hierarchy for Reference:
-The following classes will be used as a reference in the following sections to create types and mappings
+The following classes will be used as a reference in the following sections to create types and mappings.
 
 [source,java]
 ----
@@ -255,37 +173,32 @@ class Office implements Serializable {
 }
 ----
 
-=== Creating Types[[normalTypeDefinitions]]
+=== Creating Types [[normalTypeDefinitions]]
 
-NOTE: The `Type` suffix in the Type Names below is just for convenience. Types can have the same name
-as their Java/Portable/Compact class, and are otherwise not limited naming-wise. The only limitation is that the
-types must have distinct names within the set of names of all mappings and views as they
-all share the same name space.
+NOTE: Types can have a different name than their Java class or Portable/Compact/Avro schema. The only limitation is that the types must have distinct names within the set of names of all mappings and views as they all share the same namespace.
 
 [source,sql]
 ----
-CREATE TYPE OrganizationType (
+CREATE TYPE Organization (
     id BIGINT
     name VARCHAR,
-    office OfficeType
+    office Office
 ) OPTIONS (
     'format'='java',
     'javaClass'='com.example.Organization'
 );
 
-CREATE TYPE OfficeType (
+CREATE TYPE Office (
     id BIGINT
     name VARCHAR
 ) OPTIONS (
-    'format'='java',
     'javaClass'='com.example.Office'
 );
 ----
 
 === Creating Mappings
 
-NOTE: The `organization` column is explicitly specified as `OrganizationType`. Without this definition, it would be
-auto-resolved as generic `OBJECT`, and would not allow querying its sub-columns.
+NOTE: The `organization` column is explicitly specified as `Organization`. Without this definition, it would be auto-resolved as generic `OBJECT`, and would not allow querying its sub-columns.
 
 ==== Normal Type Hierarchy [[normalMappings]]
 
@@ -295,7 +208,7 @@ CREATE MAPPING users (
     __key BIGINT,
     id BIGINT,
     name VARCHAR,
-    organization OrganizationType
+    organization Organization
 ) TYPE IMap OPTIONS (
     'keyFormat'='bigint',
     'valueFormat'='java',
@@ -377,9 +290,8 @@ SELECT (organization).office FROM users
 NOTE: When selecting the entire object, the query will always try to return the underlying object verbatim.
 For Java Types, this means returning an underlying Java class instance, which can fail with a `ClassNotFoundException`
 if the class is not in the classpath of the client (or embedded server) JVM.
-A way to avoid this is to select field by field instead. Additionally, this issue is not relevant for Compact
-and Portable types as sub-objects in these mappings and types are of `GenericRecord` subclass;
-`PortableGenericRecord` and `CompactGenericRecord` are present in the base distribution of Hazelcast.
+A way to avoid this is to select field by field instead. Additionally, this issue is not relevant for Portable, Compact and Avro types as sub-objects in these mappings and types are of `GenericRecord` subclass;
+`PortableGenericRecord`, `CompactGenericRecord` and `org.apache.avro.generic.GenericRecord` are present in the base distribution of Hazelcast.
 
 Using projections:
 [source,sql]
@@ -546,14 +458,11 @@ Result:
 
 INSERT and UPDATE queries are supported in a limited way, specifically:
 
-- `INSERT` and `UPDATE` queries are only supported for non-cyclic type hierarchies. Presence of a cycle
-in a type hierarchy automatically disables the ability to run these queries against any MAPPING that uses UDTs
-from that type hierarchy. However, it's still possible to use an acyclic branch of a type hierarchy
-even if that branch is used in a cyclic type hierarchy.
+- `INSERT` and `UPDATE` queries are only supported for non-cyclic type hierarchies. Presence of a cycle in a type hierarchy automatically disables the ability to run these queries against any MAPPING that uses UDTs from that type hierarchy.
 - `INSERT` queries require specifying the full list of columns even if the column of a nested type needs to be set to `NULL`.
 - `UPDATE` queries only work on the root column and also require the full list of columns and sub-columns to work.
 Updating sub-columns is technically possible by specifying column projections in place of sub-columns that shouldn't be changed.
-- Both `UPDATE` and `INSERT` work through the usage of Row Value expression (which is similar to VALUES clause of INSERT).
+- Both `UPDATE` and `INSERT` work through the usage of Row Value expression (which is similar to `VALUES` clause of `INSERT`).
 
 === Examples[[upsertExamples]]
 
@@ -626,3 +535,6 @@ Updating nested UDT column with Query Parameter:
 ----
 hz.getSql().execute("UPDATE users SET organization = ((organization).id, (organization).name, ?)", office);
 ----
+
+== Rolling Upgrade Notes
+You must drop all user-defined types and mappings with UDTs before the rolling upgrade, and recreate them with the new semantics after upgrading.
--- a/docs/modules/sql/pages/user-defined-types.adoc
+++ b/docs/modules/sql/pages/user-defined-types.adoc
@@ -1,24 +1,128 @@
 = User-Defined Types in SQL
 
-User-Defined Types (also known as UDTs, nested types, or nested fields) allow you to create custom data types that can be referenced in the xref:sql:create-mapping.adoc[CREATE MAPPING statement].
-
-UDTs are useful for creating and accessing hierarchical data structures, which can even be cyclic. See <<queryingExamples,examples>> for more information.
+User-Defined Types (also known as UDTs, nested types, or nested fields) allow you to create custom data types that can be referenced in the xref:sql:create-mapping.adoc[CREATE MAPPING statement]. UDTs are also useful for creating and accessing hierarchical data structures.
 
 == Feature Overview
 
-- All serialization formats (namely `portable`, `compact`, `java` and `avro`) are supported.
+- `portable`, `compact`, `java` and `avro` formats are supported.
 - Cyclic definitions are only supported for Java-serialized structures. Portable and Compact formats do not support cyclic schemas, and Avro support is currently limited to acyclic schemas.
 +
 NOTE: If mappings use cyclic structures, the `INSERT` and `UPDATE` queries are disabled.
-- Type options override mapping options. Where both the mapping and type define a schema (Portable class definition, Compact schema, Java class or Avro schema) and the mapping contains only `__key` or `this` fields, the schema defined by the type is used. This means that the mapping does not need to define a schema if the type defines one.
-- If the type does not specify a schema, the parent structure is used. The parent structure could be a mapping or another type.
-- Type fields are optional and are resolved from the schema if not specified. Following a `CREATE MAPPING` statement, type fields are resolved and materialized. You can check the resolved fields using a `GET_DDL` query after the type has been used in a mapping. If type fields are not specified:
+- Type options override mapping options. Where both the mapping and type define a schema (Portable class definition, Compact schema, Java class or Avro schema) and the corresponding mapping field is `__key` or `this`, the schema defined by the type is used. This means that the mapping does not need to define a schema if the type defines one.
+- If the type does not specify a schema, it is resolved from the parent structure, which may be a mapping or another type.
+- Type fields are optional and if not specified, they are resolved from the class/schema when the type is used in a `CREATE MAPPING` statement. You can check the resolved fields using a `GET_DDL` query. The field resolution feature has the following limitations.
+ ** For Portable, Compact, or Avro formats, schema fields cannot be complex (`PORTABLE`, `COMPACT`, or `RECORD`). Otherwise, an exception will be thrown once the type is used in a mapping.
+ ** For Portable format, portable IDs (`portableFactoryId`, `portableClassId` and `portableVersion`) must be unique within the cluster. Otherwise, deserialization issues can occur.
+ ** For Portable format, the field resolution relies on the Portable class being already registered (by internal means, such as adding a configuration for Portable serialization) on the member executing the `CREATE MAPPING` command.
+ ** For Java format, complex class fields are allowed and mapped to `OBJECT`.
+ ** For Java format, only the fields declared by the type class are considered. The fields inherited from superclasses are ignored.
 
- ** For Portable, Compact, or Avro formats, schema fields cannot be complex (`PORTABLE`, `COMPACT`, or `RECORD`
- ). 
- ** For Java format, complex fields are allowed and map to `OBJECT`. 
+== Creating Types
 
-== Enabling Cycling Type Support
+To create a new type, use the `CREATE TYPE` statement:
+
+[source,sql]
+----
+CREATE [OR REPLACE] TYPE [IF NOT EXISTS] MyTypeName [(
+    colName colType,
+    ...
+)] [OPTIONS (
+    'javaClass'='com.mypackage.MyJavaClass' <1>
+  | 'compactTypeName'='MyCompactRecordName' <2>
+  | 'portableFactoryId'='123', 'portableClassId'='456', ['portableVersion'='789'] <3>
+  | 'avroSchema'='{"type":"record","name":"myType","fields":[{"name":"colName","type":"colType"},...]}' <4>
+)]
+----
+<1> In `java` format, you can use the `javaClass` option to override the type class with a subclass. This is necessary if the original type class is abstract or interface, and you want to use `INSERT` and `UPDATE` statements.
+<2> In `portable` format, you can use the `portableFactoryId`, `portableClassId` and `portableVersion` options to specify a portable ID for the type. This is only effective when the type is used for `__key` or `this` fields. `portableVersion` defaults to 0 if not specified. Portable ID components must be defined together; incomplete definitions have the same effect as no definition.
+<3> In `compact` format, you can use the `compactTypeName` option to specify the Compact record name. If unspecified, it defaults to `<TypeName>CompactType`.
+<4> In `avro` format, you can use the `avroSchema` option to specify an inline Avro schema. This is only effective when the type is used for `__key` or `this` fields.
+
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


rebekah-lawrence (changes requested):

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -255,37 +173,32 @@ class Office implements Serializable {
...
 ----
 
-=== Creating Types[[normalTypeDefinitions]]
+=== Creating Types [[normalTypeDefinitions]]
 
-NOTE: The `Type` suffix in the Type Names below is just for convenience. Types can have the same name
-as their Java/Portable/Compact class, and are otherwise not limited naming-wise. The only limitation is that the
-types must have distinct names within the set of names of all mappings and views as they
-all share the same name space.
+NOTE: Types can have a different name than their Java class or Portable/Compact/Avro schema. The only limitation is that the types must have distinct names within the set of names of all mappings and views as they all share the same namespace.
```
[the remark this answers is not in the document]
burakgok: I will accept this if you change
1. "names used in all mappings" to "names of all mappings", and
2. "a namespace" to "the same namespace".

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -64,30 +51,34 @@ To create a new type, use the `CREATE TYPE` statement:
...
 === Java Format notes
 For `java` format, if the column list is omitted, it will be automatically resolved from the corresponding
 Java class. Note that the column list will only be extracted from the source class itself;
 if it has columns that are inherited from a superclass, these columns will not be resolved.
 
 === Portable Format notes
-When using the `portable` format, make sure that the `factoryId`, `classId`, and `version` tuples are unique within the cluster.
-Otherwise, deserialization issues can happen if the corresponding class IDs and factory IDs are registered in the client
-for serialization/deserialization.
+When using the `portable` format, make sure that (`factoryId`, `classId`, `version`) tuples are unique within the cluster. Otherwise, deserialization issues may happen.
```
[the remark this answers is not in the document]
burakgok: You shouldn't change the tuple notation as it has a mathematical meaning, which is very hard to express using plain English: (`factoryId`, `classId`, `version`) tuples. "mut" should be "must".

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -64,30 +51,34 @@ To create a new type, use the `CREATE TYPE` statement:
...
+)]
 ----
-<1> `java` format requires the `javaClass` option.
-<2> `portable` format requires `portableFactoryId` and `portableClassId` and optionally `portableVersion`.
-<3> `compact` format requires `compactTypeName` - this is not the name of the created type, but rather internal name of the Compact record type, used internally by the Compact Serialization format.
+<1> In `java` format, you can use `javaClass` option to override the type class with a subclass. Note that you must provide a concrete class for `INSERT` statement to work, which makes `javaClass` option mandatory when the original type class is abstract or interface.
+<2> In `portable` format, you can use `portableFactoryId`, `portableClassId` and `portableVersion` options to specify a portable ID for the type, which is only effective when the type is used for `__key` or `this` fields. `portableVersion` is optional, similarly to `CREATE MAPPING` statement. However, it is not possible to inherit some components from the parent, so portable ID components must be defined together. Incomplete definitions have the same effect as no definition.
```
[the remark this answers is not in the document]
burakgok: "like the `CREATE MAPPING` statement" means that `CREATE MAPPING` statement is optional. It should be "like in" or "similarly to".

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -64,30 +51,34 @@ To create a new type, use the `CREATE TYPE` statement:
...
+    'javaClass'='com.mypackage.MyJavaClass' <1>
+  | 'compactTypeName'='MyCompactRecordName' <2>
+  | 'portableFactoryId'='123', 'portableClassId'='456', ['portableVersion'='789'] <3>
+  | 'avroSchema'='{"type":"record","name":"myType","fields":[{"name":"colName","type":"colType"},...]}' <4>
+)]
 ----
-<1> `java` format requires the `javaClass` option.
-<2> `portable` format requires `portableFactoryId` and `portableClassId` and optionally `portableVersion`.
-<3> `compact` format requires `compactTypeName` - this is not the name of the created type, but rather internal name of the Compact record type, used internally by the Compact Serialization format.
+<1> In `java` format, you can use `javaClass` option to override the type class with a subclass. Note that you must provide a concrete class for `INSERT` statement to work, which makes `javaClass` option mandatory when the original type class is abstract or interface.
```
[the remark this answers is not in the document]
burakgok: "You must provide a concrete class for `INSERT` statements" can be replaced with "You must provide a concrete class to be able to `INSERT`".

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,60 +1,47 @@
...
-- Limited support for `INSERT` and `UPDATE` queries for mappings using UDTs; if a `TYPE` hierarchy contains cycles,
-`INSERT` and `UPDATE` statements are disabled for the mapping.
-- Support for using UDT-based projections in both normal `SELECT` projection lists, `WHERE` filters and in `JOIN` conditions.
+- All serialization formats (namely `portable`, `compact`, `java` and `avro`) are supported.
+- Cyclic definitions are only supported for Java-serialized structures. (Portable and Compact formats do not support cyclic schemas at all. Avro support is currently limited to acyclic schemas.) Note that `INSERT`/`UPDATE` queries are disabled for mappings that use cyclic structures.
+- Type options _always_ override mapping options. Specifically, if the mapping has only `__key` or `this` fields, and both the mapping and type define a schema (Portable class definition, Compact schema, Java class or Avro schema), the one defined by the type is used. This also means that, in such situations, as long as the type defines a schema, the mapping does not need to.
+- If type schema is not specified, it will be resolved from the parent structure, which may be a mapping or another type.
```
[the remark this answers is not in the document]
burakgok: "the parent structure must provide the schema in a mapping or another type" should be "it will be resolved from the parent structure, which may be a mapping or another type". The parent structure **is** a mapping or type; it cannot specify schema in a mapping or type.

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,60 +1,47 @@
...
+- Type fields are optional and will be resolved from the schema if they are not specified. In that case, schema fields cannot be complex (`PORTABLE`/`COMPACT`/`RECORD`) for Portable/Compact/Avro formats. For Java format, complex fields are allowed and mapped to `OBJECT`. Type fields are resolved and materialized once the type is used in a `CREATE MAPPING` statement. The resolved fields can be checked by issuing a `GET_DDL` query after using the type in a mapping.
```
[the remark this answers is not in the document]
burakgok: 1. When using Portable, Compact, or Avro formats, schema fields **can** be complex. You deleted "In that case", which is why the meaning changed.
2. "If using" sounds strange to me, "when using" might be better.
3. "`OBJECT`s" should be "`OBJECT`".

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: So, this is a mathematical tuple, not a Python tuple data type? A mathematical tuple is a finite sequence of elements, I believe, and the Python data type stores an ordered and unchangeable collection of data.

Are you listing available tuples, or providing an example of a tuple?

Either way, I feel that this needs more explanation, to effectively explain the issue and solution.

And, yes, mut ought to be must ;-) sorry, didn't mean to suggest a new typo!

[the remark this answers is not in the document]
burakgok: 1. "`factoryId`, `classId`, and `version` tuples" means that `factoryId`, `classId`, and `version` are tuples themselves.
2. "(`factoryId`, `classId`, `version`) tuples" means that we are talking about 3-tuples, whose elements denote `factoryId`, `classId` and `version`.

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: Amended suggestion

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: amended

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: No, it cannot. "to be able to" is not a good phrase.

How about "You must provide a concrete class to use `INSERT` statements" ?

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: Why do we need the comparison? It is optional; what else does the user need to know? What information are you trying to communicate by the comparison?

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: But then the sentence has no agreement: use INSERT statements vs use an INSERT statement

statements is the correct term here.

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
burakgok: Saying `portableVersion` is optional might imply that it is only optional in the `CREATE TYPE` statement if the user hasn't read the `CREATE MAPPING` statement yet. I want them to know that `version` of a portable ID is always optional. However, you can drop that part if you think that it complicates the explanation unnecessarily.

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: Amended the comment so that it, I hope, meets both our concerns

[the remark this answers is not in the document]
burakgok: "(`factoryId`, `classId`, and `version`)" doesn't conform the tuple notation and "triple" is unnecessary. If there is a way to render LaTeX, you can convert the tuple to a mathematical expression. Otherwise, you have to mix math and plain English. I admit that it might be a bit confusing to the reader, but non-standard notation doesn't help.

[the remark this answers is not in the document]
burakgok: Alternatively, you can say "that portable IDs (`factoryId`, `classId`, `version`) are unique".

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: Dropped the comparison

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: **The INSERT statement** is a type of SQL query
**An INSERT statement** is an instance of the INSERT SQL query type
**INSERT statements** is a generalised way of referring to INSERT SQL queries

I may not be a programmer, but I have worked in software development for over 30 years.

I really appreciate the work you are putting into this, and the time you are taking to explain things (such as the tuples conversation). However, on this point, I remain unmovable.

How about: "You must provide a concrete class for each `INSERT` statement;", would that allow us to agree?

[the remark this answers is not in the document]
burakgok: This is not really about software development. In English, a noun can refer to a "concept" instead of a particular instance (or instances). In such a case, it becomes uncountable, and consequently singular. For example, we can say that "police closed the investigation" (without a/the/-s) if we are not interested in a particular police. You must have seen such usages before in the news(papers). "`INSERT` statement" is not wrong (even without "the"), and it looks more natural to me. If you disagree, your suggestion can stay as it is.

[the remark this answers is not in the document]
rebekah-lawrence: I disagree.

I understand, but, in this case `INSERT` = police, `INSERT` statement = the police. 

Police is a group term, or with your obvious mathematical bias, uncountable. Police is used when speaking of the force in general.
The police is specific, hence the definite article, used when we think the reader knows what we are referring to. 
The plural of police is still police.
You would never say a police; it would be a policeman, an officer, a policewoman.

This isn't a news(paper). It is technical communication. We need to be clear (hence why we don't use LaTex and mathematical equations, which "might be a bit confusing to the reader"). We need to make sure our information is clear and usable, with no misinterpretation, or cognitive dissonance.

You can check Microsoft's SQL docs to check the use if you like; for example: https://learn.microsoft.com/en-us/sql/t-sql/queries/top-transact-sql?view=sql-server-ver16

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,60 +1,53 @@
...
-User-Defined Types (also known as UDTs, nested types, or nested fields) is an experimental feature that allows you to create
- custom data types that can be referenced in the xref:sql:create-mapping.adoc[CREATE MAPPING statement].
+User-Defined Types (also known as UDTs, nested types, or nested fields) allow you to create custom data types that can be referenced in the xref:sql:create-mapping.adoc[CREATE MAPPING statement].
 
-UDTs are useful for creating and accessing hierarchical data structures, including simple cases of nested objects and
-more complex cases of fully/partially connected graphs of objects. See <<queryingExamples,examples>> for more information.
-
-WARNING: Support for UDTs is an experimental feature, and disabled by default.
-See <<enabling-udt-support, Enabling UDT support>> to learn how to enable it.
+UDTs are useful for creating and accessing hierarchical data structures, which can even be cyclic. See <<queryingExamples,examples>> for more information.
```
k-jamroz, line 5: I would not mention cyclic types in into. They only work for Java format and require enabling feature flag

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,60 +1,53 @@
...
 
 == Feature Overview
 
-Due to the experimental nature of UDTs, the feature set is limited to the following:
+- All serialization formats (namely `portable`, `compact`, `java` and `avro`) are supported.
```
k-jamroz, line 9: mapping format != serialization format. In particular Avro is not HZ serialization format, DS/IDS is not explicitly configured in SQL.

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -625,4 +547,7 @@ Updating nested UDT column with Query Parameter:
 [source,java]
 ----
 hz.getSql().execute("UPDATE users SET organization = ((organization).id, (organization).name, ?)", office);
-----
\ No newline at end of file
+----
+
+== Rolling Upgrade Notes
+You must drop all user-defined types and mappings with UDTs before the rolling upgrade, and recreate them with the new semantics after upgrading.
```
k-jamroz, line 553: add information about versions to which this restricion applies. This documentation will probably not be updated in 5.5 and later

docs/modules/sql/pages/user-defined-types.adoc
k-jamroz: BTW, we don't support JSON for UDTs, do we? (it is supported as MAPPING format)

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,60 +1,53 @@
...
-- Limited support for `INSERT` and `UPDATE` queries for mappings using UDTs; if a `TYPE` hierarchy contains cycles,
-`INSERT` and `UPDATE` statements are disabled for the mapping.
-- Support for using UDT-based projections in both normal `SELECT` projection lists, `WHERE` filters and in `JOIN` conditions.
+ ** For Portable, Compact, or Avro formats, schema fields cannot be complex (`PORTABLE`, `COMPACT`, or `RECORD`
+ ). 
+ ** For Java format, complex fields are allowed and map to `OBJECT`. 
 
-== Enabling UDT Support
-You can enable UDTs by setting the `hazelcast.sql.experimental.custom.types.enabled` property to `true` in the member configuration.
+== Enabling Cycling Type Support
```
k-jamroz, line 21: I would move this section toward end of the document

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -64,30 +57,39 @@ To create a new type, use the `CREATE TYPE` statement:
...
+<2> In `portable` format, you can use the `portableFactoryId`, `portableClassId` and `portableVersion` options to specify a portable ID for the type. This is only effective when the type is used for `__key` or `this` fields. `portableVersion` is optional. However, components cannot be inherited from the parent, so portable ID components must be defined together. Incomplete definitions have the same effect as no definition.
+<3> In `compact` format, you can use the `compactTypeName` option to specify the Compact record name. If unspecified, it defaults to `<TypeName>CompactType`.
+<4> In `avro` format, you can use the `avroSchema` option to specify an inline Avro schema. This is only effective when the type is used for `__key` or `this` fields.
+
+NOTE: `EXTERNAL NAME` aliases are not supported for types, column names must have the same name as their corresponding Java class field or Portable/Compact/Avro schema field.
```
k-jamroz, line 75:
```suggestion
```

note is duplicated

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -64,30 +57,39 @@ To create a new type, use the `CREATE TYPE` statement:
...
+[NOTE]
+====
+. `EXTERNAL_NAME` aliases  are not supported for types, column names must have the same name as their corresponding Java class field or Portable/Compact/Avro schema field.
+. You can mix options that belong to different formats. When you create a mapping that references your UDT, the relevant options are used and the others are ignored, which makes it possible to use a UDT in multiple mappings having different serialization formats.
+. To reference another type, you must provide the column list. Otherwise, in Java format, the column will be resolved as `OBJECT` type; in Portable, Compact and Avro formats, an exception will be thrown once the type is used in a mapping.
+====
+NOTE: You can mix options that belong to different formats. When you create a mapping that references your UDT, the relevant options are used and the others are ignored, which makes it possible to use a UDT in multiple mappings having different serialization formats.
```
k-jamroz, line 82: I think " it possible to use a UDT in multiple mappings having different serialization formats" is quite important feature (and significant improvement over previous approach) and should be mentioned in the overview. Do we have examples for that like read from kafka and write to IMap with the same UDT (OFC the syntax is not pretty as you have to recreate the object from fields)?

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,60 +1,53 @@
...
+- All serialization formats (namely `portable`, `compact`, `java` and `avro`) are supported.
+- Cyclic definitions are only supported for Java-serialized structures. Portable and Compact formats do not support cyclic schemas, and Avro support is currently limited to acyclic schemas.
++
+NOTE: If mappings use cyclic structures, the `INSERT` and `UPDATE` queries are disabled.
+- Type options override mapping options. Where both the mapping and type define a schema (Portable class definition, Compact schema, Java class or Avro schema) and the mapping contains only `__key` or `this` fields, the schema defined by the type is used. This means that the mapping does not need to define a schema if the type defines one.
+- If the type does not specify a schema, the parent structure is used. The parent structure could be a mapping or another type.
+- Type fields are optional and are resolved from the schema if not specified. Following a `CREATE MAPPING` statement, type fields are resolved and materialized. You can check the resolved fields using a `GET_DDL` query after the type has been used in a mapping. If type fields are not specified:
```
k-jamroz, line 15: AFAIK this "magic" materialization was a side effect and not something uses should rely on, as it is subject to change in the future

k-jamroz: and were not sure if it makes sense in the long term

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -64,30 +57,39 @@ To create a new type, use the `CREATE TYPE` statement:
...
+. You can mix options that belong to different formats. When you create a mapping that references your UDT, the relevant options are used and the others are ignored, which makes it possible to use a UDT in multiple mappings having different serialization formats.
+. To reference another type, you must provide the column list. Otherwise, in Java format, the column will be resolved as `OBJECT` type; in Portable, Compact and Avro formats, an exception will be thrown once the type is used in a mapping.
+====
+NOTE: You can mix options that belong to different formats. When you create a mapping that references your UDT, the relevant options are used and the others are ignored, which makes it possible to use a UDT in multiple mappings having different serialization formats.
 
-NOTE: To reference another type, you must provide the column list. Otherwise, the column may be automatically resolved as an `OBJECT` type.
+NOTE: To reference another type, you must provide the column list. Otherwise, in Java format, the column will be resolved as `OBJECT` type; in Portable, Compact and Avro formats, an exception will be thrown once the type is used in a mapping.
```
k-jamroz, line 84: we have more notes than normal text here: we have "notes" boxes and "java format notes" and "protable format notes" sections. IMO it would be simpler if we had normal section for each format

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -377,9 +303,8 @@ SELECT (organization).office FROM users
 NOTE: When selecting the entire object, the query will always try to return the underlying object verbatim.
 For Java Types, this means returning an underlying Java class instance, which can fail with a `ClassNotFoundException`
 if the class is not in the classpath of the client (or embedded server) JVM.
-A way to avoid this is to select field by field instead. Additionally, this issue is not relevant for Compact
-and Portable types as sub-objects in these mappings and types are of `GenericRecord` subclass;
-`PortableGenericRecord` and `CompactGenericRecord` are present in the base distribution of Hazelcast.
+To avoid this, you can select a field-by-field instead. Additionally, this issue does not apply to Portable, Compact, or Avro types as sub-objects in these mappings and types are `GenericRecord` subclasses;
```
k-jamroz, line 306: remeber that there are different clients that java. In particular will Avro be supported in them?

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -96,76 +98,17 @@ This mechanism is not reliable since this command will fail if the member that e
...
 
 === Replacing Types and Type Consistency
 Currently, there is a limitation on the replacement of existing types:
 if the replaced type was already used in a mapping, you need to fully replace that mapping
 to update its data type information using the `DROP MAPPING` and `CREATE MAPPING` statements.
 However, if the type hierarchy was not used in a mapping, any type in that hierarchy can be safely
 changed, and these changes will appear in the new mapping. This is because the links
-between types are symbolic (based on the name only), and they're only "materialized" once used in a mapping.
-
-=== CREATE TYPE Examples
```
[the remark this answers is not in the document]
burakgok: 099f756d

docs/modules/sql/pages/user-defined-types.adoc
burakgok: I removed it from notes. Now, it is only mentioned in "Replacing Types and Type Consistency". (099f756d)

docs/modules/sql/pages/user-defined-types.adoc
burakgok: It is now mentioned in "Support for Cycles". (099f756d)

docs/modules/sql/pages/user-defined-types.adoc
burakgok: 099f756d

docs/modules/sql/pages/user-defined-types.adoc
burakgok: I removed the "serialization" word. (099f756d) We don't support JSON for UDTs.

docs/modules/sql/pages/user-defined-types.adoc
burakgok: 099f756d

docs/modules/sql/pages/user-defined-types.adoc
burakgok: Since there is no `format` option for UDTs, users would expect them to be format-agnostic. That's why, I think, we don't need to highlight this fact.

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,161 @@
...
+- Type fields are optional and if not specified, they are resolved from the class/schema when the type is used in a `CREATE MAPPING` statement. You can check the resolved fields using a `GET_DDL` query. The field resolution feature has the following limitations.
+ ** For Portable, Compact, or Avro formats, schema fields cannot be complex (`PORTABLE`, `COMPACT`, or `RECORD`). Otherwise, an exception will be thrown once the type is used in a mapping.
+ ** For Portable format, portable IDs (`portableFactoryId`, `portableClassId` and `portableVersion`) must be unique within the cluster. Otherwise, deserialization issues can occur.
+ ** For Portable format, the field resolution relies on the Portable class being already registered (by internal means, such as adding a configuration for Portable serialization) on the member executing the `CREATE MAPPING` command.
+ ** For Java format, complex class fields are allowed and mapped to `OBJECT`.
+ ** For Java format, only the fields declared by the type class are considered. The fields inherited from superclasses are ignored.
```
k-jamroz, line 18: this is a bit surprising? were inherited fields ignored before?

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,161 @@
...
+) TYPE IMap OPTIONS (
+    'keyFormat'='bigint',
+    'valueFormat'='java',
+    'valueJavaClass'='com.example.User'
+);
 ----
 
-Compact Type:
+== Support for Cycles
+Cycles between types are only supported for Java format. However, this is limited to querying, not inserting or updating.
```
k-jamroz, line 121:
```suggestion
Cycles between types are only supported for Java format. However, the support is limited only to querying. Inserting or updating with cyclic types is currently not supported.
```

docs/modules/sql/pages/user-defined-types.adoc
k-jamroz: but we say in the beginning that "portable, compact, java and avro formats are supported" - by analogy with mappings it might be assumed that UDT has also single assigned format. IMO worth mentioning this explicitly.

docs/modules/sql/pages/user-defined-types.adoc
k-jamroz: I do not see the place you mentioned. However I can see:

> If the type does not specify a schema, it is resolved from the parent structure, which may be a mapping or another type.

(and next bullet) which makes this magic UDT field resolution a "normal" feature, but we had doubts if we want to keep it

docs/modules/sql/pages/user-defined-types.adoc
k-jamroz: ok, at least in the docs it was said that it is not supported

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,161 @@
...
-    <properties>
-        <property name="hazelcast.sql.experimental.custom.types.enabled">true</property>
-    </properties>
-</hazelcast>
-----
---
-
-YAML::
+- `portable`, `compact`, `java` and `avro` formats are supported.
+- Cyclic definitions are only supported for Java-serialized structures. Portable and Compact formats do not support cyclic schemas, and Avro support is currently limited to acyclic schemas.
```
k-jamroz, line 8: I don't think cyclic stuctures are a first thing that comes to mind when someone reads about UDTs. We can keep it in "Feature Overview" but move it to be last in the list.

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,161 @@
...
-
-Java::
-+
-[source,java]
-----
-final Config config = new Config();
-config.setProperty("hazelcast.sql.experimental.custom.types.enabled", "true");
-----
-====
+NOTE: If mappings use cyclic structures, the `INSERT` and `UPDATE` queries are disabled.
```
k-jamroz, line 10: this is repeated later in "Support for Cycles". IMO this is too much detail for "feature overview". In this section we can mention that we support cycles but there are important limitations described in detail later.

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,161 @@
...
+- Type options override mapping options. Where both the mapping and type define a schema (Portable class definition, Compact schema, Java class or Avro schema) and the corresponding mapping field is `__key` or `this`, the schema defined by the type is used. This means that the mapping does not need to define a schema if the type defines one.
```
k-jamroz, line 11: mapping -> MAPPING? we mean specifically SQL MAPPING not any mapping

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,161 @@
...
+    ...
+)] [OPTIONS (
+    'javaClass'='com.mypackage.MyJavaClass' <1>
+  | 'compactTypeName'='MyCompactRecordName' <2>
+  | 'portableFactoryId'='123', 'portableClassId'='456', ['portableVersion'='789'] <3>
+  | 'avroSchema'='{"type":"record","name":"myType","fields":[{"name":"colName","type":"colType"},...]}' <4>
+)]
+----
+<1> In `java` format, you can use the `javaClass` option to override the type class with a subclass. This is necessary if the original type class is abstract or interface, and you want to use `INSERT` and `UPDATE` statements.
+<2> In `portable` format, you can use the `portableFactoryId`, `portableClassId` and `portableVersion` options to specify a portable ID for the type. This is only effective when the type is used for `__key` or `this` fields. `portableVersion` defaults to 0 if not specified. Portable ID components must be defined together; incomplete definitions have the same effect as no definition.
```
k-jamroz, line 37:
```suggestion
<2> In `portable` format, you can use the `portableFactoryId`, `portableClassId` and `portableVersion` options to specify a portable ID for the type. This is only effective when the type is used for `__key` or `this` fields. `portableVersion` defaults to 0 if not specified. Required Portable ID components must be defined together; incomplete definitions are ignored.
```

docs/modules/sql/pages/user-defined-types.adoc
k-jamroz: this is documented explicitly only in note in "Creating Types"

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,161 @@
...
 
 === Replacing Types and Type Consistency
 Currently, there is a limitation on the replacement of existing types:
-if the replaced type was already used in a mapping, you need to fully replace that mapping
+if the replaced type was already used in a mapping, you need to recreate that mapping
 to update its data type information using the `DROP MAPPING` and `CREATE MAPPING` statements.
 However, if the type hierarchy was not used in a mapping, any type in that hierarchy can be safely
 changed, and these changes will appear in the new mapping. This is because the links
-between types are symbolic (based on the name only), and they're only "materialized" once used in a mapping.
+between types are symbolic (based on the name only), and they are materialized only when used in a mapping.
```
k-jamroz, line 53: and fields for fieldless UDT are captured and stored at this time

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,161 @@
...
+    'valueJavaClass'='com.example.User'
+);
 ----
 
-Compact Type:
+== Support for Cycles
+Cycles between types are only supported for Java format. However, this is limited to querying, not inserting or updating.
 
-[source,sql]
+NOTE: The existence of referenced types are only verified when the type is used in a `CREATE MAPPING` statement.
```
k-jamroz, line 123:
```suggestion
NOTE: It is possible to reference a not yet existing type (including the type being defined) in type definition which is necessary to create cyclic structure. The existence of referenced types is only verified when the type is used in a `CREATE MAPPING` statement.
```
Without the addition I did not understand how this note relates to cyclic types.

"including the type being defined" - is that true?

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,161 @@
...
+
+=== Enabling Cycling Type Support
+You can enable cyclic types by setting the `hazelcast.sql.experimental.custom.cyclic.types.enabled` property to `true` in the member configuration.
```
k-jamroz, line 126:
```suggestion
You can enable cyclic types by setting the `hazelcast.sql.experimental.custom.cyclic.types.enabled` property to `true` in the member configuration. It is disabled by default.
```

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -352,15 +252,13 @@ Querying is provided with the field access operator which has the following synt
 `mappingColumn` must be the top-level column inside a mapping that has a UDT as its type,
```
k-jamroz, line 252: I would add a note pointing out that parentheses are required. This is very surprising quirk of our syntax

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -591,11 +478,9 @@ Replacing nested column value:
 UPDATE users SET organization = ((organization).id, (organization).name, ((organization).office.id, 'new-office-name'))
 ----
 
-NOTE: Updating UDT-based columns requires providing a value for every column in the UDT and its child UDTs, however
-`null` can also be specified in place of nested UDT column to initialize it to `null`. Not providing full list of columns
-will cause a query validation error.
+NOTE: Updating UDT columns requires providing a value for every column in the UDT and its child UDTs unless it needs to be set to `null`. Not providing full list of columns will cause a query validation error.
 
-Inserting with Query Parameter (java only):
+Inserting with query parameter (Java only):
```
k-jamroz, line 483: is it really java only? the example maybe (uses `Serializable`), but this could work in principle for portable and compact from other languages as well.

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -620,9 +505,12 @@ Using `organization` from the example above.
...
+Updating nested UDT column with query parameter:
 
 [source,java]
 ----
 hz.getSql().execute("UPDATE users SET organization = ((organization).id, (organization).name, ?)", office);
-----
\ No newline at end of file
+----
+
+== Rolling Upgrade Notes
```
k-jamroz, line 515: this affects also normal upgrade with persistent SQL catalog. both are EE features - we should have "enterprise" marker here

docs/modules/sql/pages/user-defined-types.adoc
burakgok: Field materialization is mentioned in "Replacing Types and Type Consistency" as the following, and I think this is the right place for such an implementation detail.
> ... if the type hierarchy was not used in a mapping, any type in that hierarchy can be safely changed, and these changes will appear in the new mapping. **This is because the links between types are symbolic (based on the name only), and they are materialized only when used in a mapping.**

docs/modules/sql/pages/user-defined-types.adoc
burakgok: Are you suggesting capitalizing all "mapping" usages? The name of the page is "User-Defined Types in SQL" and I think it is not ambiguous what "mapping" refers to.

docs/modules/sql/pages/user-defined-types.adoc
burakgok: In Feature Overview, it says
> Type fields are optional and if not specified, they are resolved from the class/schema when the type is used in a `CREATE MAPPING` statement.

Thus, both the field resolution and materialization happens when the type is used in mapping. They are mentioned in different sections, which is, I think, not a shortcoming since the documentation is not pivoted around mapping/type lifecycle.

docs/modules/sql/pages/user-defined-types.adoc
burakgok: fef8c8c2

docs/modules/sql/pages/user-defined-types.adoc
burakgok: fef8c8c2

docs/modules/sql/pages/user-defined-types.adoc
burakgok: fef8c8c2

docs/modules/sql/pages/user-defined-types.adoc
burakgok: fef8c8c2

docs/modules/sql/pages/user-defined-types.adoc
burakgok: fef8c8c2

docs/modules/sql/pages/user-defined-types.adoc
burakgok: fef8c8c2

docs/modules/sql/pages/user-defined-types.adoc
burakgok: Yes, it is true. (fef8c8c2)

docs/modules/sql/pages/user-defined-types.adoc
burakgok: fef8c8c2

docs/modules/sql/pages/user-defined-types.adoc
burakgok: fef8c8c2

docs/modules/sql/pages/user-defined-types.adoc
burakgok: I removed the "Java only" notice and added a note for Avro support on non-Java clients. (fef8c8c2)

docs/modules/sql/pages/user-defined-types.adoc
burakgok: fef8c8c2

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,159 @@
...
+) TYPE IMap OPTIONS (
+    'keyFormat'='bigint',
+    'valueFormat'='java',
+    'valueJavaClass'='com.example.User'
+);
 ----
 
-Compact Type:
+== Support for Cycles
+When creating a UDT, the existence of referenced types are only verified when the type is used in a `CREATE MAPPING` statement. This makes it possible to create cyclic types.
```
k-jamroz, line 119:
```suggestion
When creating a UDT, the existence of referenced types is only verified when the type is used in a `CREATE MAPPING` statement. This makes it possible to create cyclic types.
```

docs/modules/sql/pages/user-defined-types.adoc
k-jamroz: this sentence is about UDT references ("links between types") not fields of fieldless UDT. It is not obvious that is it the same.

And this is not only implementation detail. It has visible impact on the user. We say that:

> Type fields are optional and if not specified, they are resolved from the class/schema when the type is used in a CREATE MAPPING statement.

but we do not say what happens if the underlying schema changes (eg. Java class is updated).
On it's own the above sentence could mean that fieldless UDT may get different set of fields for each context in which it used which is not the way we work.

docs/modules/sql/pages/user-defined-types.adoc
k-jamroz: we can keep this as is

docs/modules/sql/pages/user-defined-types.adoc
k-jamroz: let's move discussion to https://github.com/hazelcast/hz-docs/pull/969#discussion_r1466279286

docs/modules/sql/pages/user-defined-types.adoc
burakgok: 7bc63e25

k-jamroz (approved):

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -591,11 +481,9 @@ Replacing nested column value:
 UPDATE users SET organization = ((organization).id, (organization).name, ((organization).office.id, 'new-office-name'))
 ----
 
-NOTE: Updating UDT-based columns requires providing a value for every column in the UDT and its child UDTs, however
-`null` can also be specified in place of nested UDT column to initialize it to `null`. Not providing full list of columns
-will cause a query validation error.
+NOTE: Updating UDT columns requires providing a value for every column in the UDT and its child UDTs unless it needs to be set to `null`. Not providing full list of columns will cause a query validation error.
```
rebekah-lawrence, line 484:
```suggestion
NOTE: To update UDT columns, a value must be provided for every column in the UDT and its child UDTs unless it needs to be set to `null`. If a full list of columns is not provided, a query validation error occurs.
```

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -611,7 +499,9 @@ organization.office = office;
 hz.getSql().execute("INSERT INTO users VALUES (1, 'user1', ?)", organization);
 ----
 
-Updating with Query Parameter:
+NOTE: For Avro-serialized types, the query parameters need to be subclasses of `org.apache.avro.generic.GenericRecord`, whose (de)serialization is supported by Java clients only.
```
rebekah-lawrence, line 502:
```suggestion
NOTE: For Avro-serialized types, the query parameters must be subclasses of `org.apache.avro.generic.GenericRecord`, whose (de)serialization is supported by Java clients only.
```

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -620,9 +510,14 @@ Using `organization` from the example above.
...
 ----
 hz.getSql().execute("UPDATE users SET organization = ((organization).id, (organization).name, ?)", office);
-----
\ No newline at end of file
+----
+
+== Upgrade Notes
+[.enterprise]*Enterprise*
+
+When performing a (normal or rolling) upgrade from version 5.3 to 5.4, you must drop all user-defined types and mappings with UDTs before the upgrade, and recreate them with the new semantics after upgrading.
```
rebekah-lawrence, line 523:
```suggestion
When performing a normal or rolling upgrade from version 5.3 to 5.4, you must drop all user-defined types and mappings with UDTs before the upgrade, and recreate them with the new semantics after upgrading.
```

rebekah-lawrence (changes requested):
You've put a lot of work in here, Burak, thanks. 
Just a few more changes, and I think we'll be there.

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,160 @@
...
-
-Java::
-+
-[source,java]
-----
-final Config config = new Config();
-config.setProperty("hazelcast.sql.experimental.custom.types.enabled", "true");
-----
-====
+- UDTs can only be incorporated into mappings that has `portable`, `compact`, `java` or `avro` format although UDTs themselves are not tied to a specific format.
```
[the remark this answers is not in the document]
burakgok: 1. What is wrong with "incorporated into"?
2. "a ... format" should be changed to "the ... format".

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -1,173 +1,160 @@
...
+- Type options override mapping options. Where both the mapping and type define a schema (Portable class definition, Compact schema, Java class or Avro schema) and the corresponding mapping field is `__key` or `this`, the schema defined by the type is used. This means that the mapping does not need to define a schema if the type defines one.
+- If the type does not specify a schema, it is resolved from the parent structure, which may be a mapping or another type.
+- Type fields are optional and if not specified, they are resolved from the class/schema when the type is used in a `CREATE MAPPING` statement. You can check the resolved fields using a `GET_DDL` query. The field resolution feature has the following limitations.
```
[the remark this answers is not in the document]
burakgok: I can accept this, but in English, sentences cannot end with a semicolon. The following usage is acceptable, which forms a single sentence.
> Sentence begins:
    1. words,
    2. words, and
    3. words.

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -343,24 +242,22 @@ CREATE MAPPING tableC (
 
 == Querying Support
 
-Querying is provided with the field access operator which has the following syntax:
+Querying is provided with the field access operator which has the following syntax (parentheses are required):
```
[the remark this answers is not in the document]
burakgok: Parenthesis is singular.

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -374,24 +271,22 @@ Selecting whole sub-object:
...
-For Java Types, this means returning an underlying Java class instance, which can fail with a `ClassNotFoundException`
-if the class is not in the classpath of the client (or embedded server) JVM.
-A way to avoid this is to select field by field instead. Additionally, this issue is not relevant for Compact
-and Portable types as sub-objects in these mappings and types are of `GenericRecord` subclass;
-`PortableGenericRecord` and `CompactGenericRecord` are present in the base distribution of Hazelcast.
+[NOTE]
+====
+. When selecting the entire object, the query will always try to return the underlying object verbatim. For Java-serialized types, this means returning an underlying Java class instance, which can fail with a `ClassNotFoundException` if the class is not in the classpath of the client (or embedded server) JVM. To avoid this, you can select individual fields instead. This issue does not apply to Portable- or Compact-serialized types as sub-objects in these mappings and types are `GenericRecord` subclasses; `PortableGenericRecord` and `CompactGenericRecord` are present in the base distribution of Hazelcast.
```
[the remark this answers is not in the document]
burakgok: I disagree. My usage of commas is correct.

docs/modules/sql/pages/user-defined-types.adoc
burakgok: "A full list" sounds very strange.

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: 1. I didn't say into was wrong. For me, incorporate = add to a body, so the "to" is already present. If you feel strongly, leave it as into.
2.  As you give alternatives, "a" is more appropriate, but "the" is not going to be wrong. It was the lack of an article (definite or indefinite) that I was commenting on.

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: It isn't a semicolon, it is a colon. 

The colon is used because the sentence is a lead-in to the list.

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: Great, make it plural then (I would have used brackets myself, to keep the level of English lower, but I was trying to keep your wording).

Typos are for everyone, you know ;-)

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
rebekah-lawrence: Keep the comma after types, then. I altered it because the sentence becomes very choppy when you use the other comma that I added. 

That would be the comma before "which", which is correct. If you prefer not to add the comma, change "which" to "that"

I was trying to do a really light-touch review this time, as you obviously take offence at every suggestion. However, if I was to have been picky, I would have pointed out that this sentence contains 41 words and needs to be re-written to address this.

docs/modules/sql/pages/user-defined-types.adoc
rebekah-lawrence: Does a "partial list" sound strange?

Also, I didn't change that, that was from your original text

rebekah-lawrence: Oh, hang on, reading that again, it is the use of the indefinite article that you are questioning. 

Nah, you need it, whether it sounds strange to you or not.

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
burakgok: 90767538

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
burakgok: 90767538

docs/modules/sql/pages/user-defined-types.adoc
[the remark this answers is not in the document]
burakgok: 90767538

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -374,24 +271,22 @@ Selecting whole sub-object:
...
 ----
-Projections work as usual as field access expressions have virtually same semantics and possible usage contexts as normal
-column projections.
+Projections work as usual since field access expressions have virtually the same semantics and possible usage contexts as normal column projections.
 
-==== Cyclic Type Hierarchy Querying
+==== Querying Cyclic Types
 
-Following examples use <<cyclicTypeDefinitions, cyclic type definitions>> and following mapping:
+Following examples use <<a-type, AType>>.
```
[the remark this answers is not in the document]
burakgok: 90767538

docs/modules/sql/pages/user-defined-types.adoc
```diff
@@ -546,22 +439,19 @@ Result:
...
 - `UPDATE` queries only work on the root column and also require the full list of columns and sub-columns to work.
 Updating sub-columns is technically possible by specifying column projections in place of sub-columns that shouldn't be changed.
-- Both `UPDATE` and `INSERT` work through the usage of Row Value expression (which is similar to VALUES clause of INSERT).
+- Both `UPDATE` and `INSERT` queries use the Row Value expression, which is similar to the `VALUES` clause of an `INSERT` query.
 
-=== Examples[[upsertExamples]]
+=== Examples [[upsertExamples]]
 
-Following examples use <<normalTypeDefinitions, normal type definitions>> and <<normalMappings, normal mappings>>.
+Following examples use <<users-mapping, `users` mapping>>, and <<organization-office-types, `Organization` and `Office` types>>.
```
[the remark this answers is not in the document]
burakgok: 90767538

docs/modules/sql/pages/user-defined-types.adoc
burakgok: 90767538

docs/modules/sql/pages/user-defined-types.adoc
burakgok: 90767538

Serdaro (approved):
LGTM

rebekah-lawrence (approved):
Thanks, LGTM

PR #980, opened 2024-01-28
[DOCS-738] - Adding content for data member demotion.

Serdaro:


docs/modules/ROOT/pages/system-properties.adoc
```diff
@@ -691,6 +691,11 @@ for processing the operations sent from the connected Management Center instance
 |int
 |Management Center maximum visible slow operations count.
 
+|`hazelcast.member.demote.max.wait`
+| 600
+| int
+| Maximum wait duration in seconds during the demotion of a data member to a lite one.
```
rebekah-lawrence, line 697:
```suggestion
| Maximum wait duration in seconds during the demotion of a data member to a lite member.
```

rebekah-lawrence (approved):
Two small suggestions; once the ??question?? is addressed, it looks good to me.

docs/modules/maintain-cluster/pages/lite-members.adoc
```diff
@@ -64,7 +80,7 @@ config.setLiteMember(true);
 
 HazelcastInstance hazelcastInstance = Hazelcast.newHazelcastInstance(config);
 Cluster cluster = hazelcastInstance.getCluster();
-cluster.promoteLocalLiteMember();
+cluster.demoteLocalDataMember();
 ----
 
-NOTE: A data member cannot be downgraded to a lite member back.
\ No newline at end of file
+You can specify a timeout value for waiting during the data member demotion, using the `hazelcast.member.demote.max.wait` property. Default timeout value is 600 seconds. ??? Any recommendations on when to increase/decrease the default value ???
```
vbekiaris, line 86: This timeout should be large enough to allow for partition migrations to complete.
A data member demotion is similar to gracefully shutting down the member: any partitions it owns will be assigned to other members of the cluster and partition data will be moved off that member. For that reason, the default demotion timeout value is the same as the graceful shutdown timeout.

PR #992, opened 2024-02-04
[DOCS-739] - Adding deprecation notice for Transactions (5.4)

Serdaro:


docs/modules/clients/pages/java.adoc
```diff
@@ -393,7 +393,7 @@ See xref:transactions:providing-xa-transactions.adoc[Transactions] for more deta
 [CAUTION]
 .Deprecation Notice for Transactions
 ====
-Hazelcast will deprecate transactions in the upcoming release (5.4). An improved version of this feature is under consideration. If you are already using transactions, get in touch and share your use case. Your feedback will help us to develop a solution that meets your needs.
+Transactions have been deprecated, and will be removed as of Hazelcast version 7.0. An improved version of this feature is under consideration. If you are already using transactions, get in touch and share your use case. Your feedback will help us to develop a solution that meets your needs.
```
avtarraikmo, line 396: Do we refer to the product as ‘Hazelcast’ or ‘Hazelcast Platform’ in the docs?

I think we need to switch to the latter if not already done so. But we should get confirmation from Product Marketing.

Serdaro: as "Hazelcast" mostly. Will talk with the PM.

avtarraikmo: Let’s pick that up in a single PR where we do it across the docs.

avtarraikmo (approved):

PR #994, opened 2024-02-05
Drafting the deprecation notice for REST API.

Serdaro:
After the wording is finalized as an output of the review cycles, the final wording will be placed into the relevant parts of the whole Platform documentation.

### DIFF
--- a/docs/modules/clients/pages/rest.adoc
+++ b/docs/modules/clients/pages/rest.adoc
@@ -5,3 +5,9 @@ Hazelcast offers an HTTP REST client with which you can communicate with your
 cluster over HTTP using the REST approach.
 See the xref:maintain-cluster:rest-api.adoc[REST API section] for details
 on its usage.
+
+[CAUTION]
+.Deprecation Notice for the REST API
+====
+REST API is deprecated and will be removed as of Hazelcast version 6.0. An improved version of this feature is under development.
+====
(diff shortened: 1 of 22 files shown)


docs/modules/clients/pages/rest.adoc
```diff
@@ -5,3 +5,9 @@ Hazelcast offers an HTTP REST client with which you can communicate with your
 cluster over HTTP using the REST approach.
 See the xref:maintain-cluster:rest-api.adoc[REST API section] for details
 on its usage.
+
+[CAUTION]
+.Deprecation Notice for the REST API
+====
+REST API is deprecated and will be removed as of Hazelcast version 6.0. An improved version of this feature is under development.
```
rebekah-lawrence, line 12: Can we give an indication of when the improved version will be available? Just so people can plan ahead, and be assured that they won't be left without any REST API between v6 and the new version.

rebekah-lawrence (commented):
I'd also suggest creating the note as a partial, and including it where required, but that is just an opinion ;-)

docs/modules/clients/pages/rest.adoc
oliverhowell: As per https://hazelcast.atlassian.net/browse/DOCS-742 should be 7.0 for removal so suggesting:

```suggestion
The REST API has been deprecated and will be removed as of Hazelcast version 7.0. An improved version of this feature is under development.
```

PR #998, opened 2024-02-07
Upgrading GH actions for doc builds

Serdaro:
As per https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/#what-you-need-to-do


rebekah-lawrence (approved):
LGTM

PR #1002, opened 2024-02-08
DO NOT MERGE - TPC Draft

oliverhowell:
Closes DOCS-719

### DIFF
--- a/docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
+++ b/docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
@@ -41,7 +41,7 @@ XML::
 [source,xml]
 ----
 <tpc enabled="true">
-    <eventloop-count>12</eventloop-count>
+    <eventloop-count>12</eventloop-count> <1>
 </tpc>
 ----
 --
@@ -52,11 +52,11 @@ YAML::
 ----
  tpc:
     enabled: true
-    eventloop-count: 12
+    eventloop-count: 12 <1>
 ----
 ====
 
-NOTE: The eventloop-count will default to the number of available CPU cores.  
+<1> The eventloop-count will default to the number of available CPU cores.  
 
 === Enabling TPC on Clients
 
--- a/docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
+++ b/docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
@@ -1,0 +1,92 @@
+= Thread-Per-Core (TPC)
+:description: Find out more about Thread-Per-Core (TPC) and how to enable this feature on clients and cluster members.
+
+You can enable Thread-Per-Core (TPC) on clients and cluster members to improve system performance, especially on machines with either fewer or more CPU cores than the recommended hardware listed on the xref:ROOT:production-checklist.adoc[]. 
+
+[[tpc-what]]
+== What is Thread-Per-Core (TPC)?
+
+The threading model is responsible for networking/storage and compute, which means it has a big impact on system performance. With Staged Event-Driven Architecture (SEDA) you have dedicated threads that do one particular task (for example, reading from a socket or evaluating a map.get) and thread pools that manage these task groups.  
+
+This architecture can lead to throughput and latency issues on machines with limited cores (fewer than 8), as the frequent context switching and state sharing can add significant overhead. Conversely, on machines with a very high number of CPUs (more than 32), a bottleneck can be created by the limited number of I/O threads vs. operation threads. Typically, smaller servers are more readily available through virtualization and for cloud-specific workloads, which can limit system performance in a SEDA environment. 
+
+A Thread-Per-Core (TPC) design uses one thread for every CPU core and every thread does everything. With each TPC thread able to do networking/storage and compute, scaling is mostly a case of increasing the number of threads to spread the work. This means you can scale up and down easily based on the number of available cores, and make more efficient use of resources on machines with either a low or high number of cores. Every task is non-blocking and you can scale vertically much easier, which means high throughput, low latency and significantly improved system performance.  
+
+When TPC is enabled, regardless of the routing mode (smart or unisocket), the client will establish connections to all cluster members. If the server side has also enabled TCP, the client will open connections to all cores of all members and route partition-specific requests to the correct cores.  TCP-enabled servers will continue using the same ports for discovery, which means there's no difference in how the cluster member list is created and TPC-aware servers will be backwards compatible with clients not using TPC.
+
+[[tpc-config]]
+== Configuration Options
+
+TPC is disabled by default and must be enabled on each cluster member and client.  
+
+NOTE: For example configuration files (XML and YAML) with TPC enabled, see xref:configuration:configuring-declaratively.adoc[].  
+
+=== Enabling TPC on Members
+
+To enable TPC on a Hazelcast cluster member, you need to set the `tpc` system property to true.  
+
+You can do this by using the following command:
+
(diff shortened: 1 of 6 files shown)


docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
```diff
@@ -0,0 +1,92 @@
+= Thread-Per-Core (TPC)
+:description: Find out more about Thread-Per-Core (TPC) and how to enable this feature on clients and cluster members.
+
+You can enable Thread-Per-Core (TPC) on clients and cluster members to improve system performance, especially on machines with either fewer or more CPU cores than the recommended hardware listed on the xref:ROOT:production-checklist.adoc[]. 
+
+[[tpc-what]]
+== What is Thread-Per-Core (TPC)?
+
+The threading model is responsible for networking/storage and compute, which means it has a big impact on system performance. With Staged Event-Driven Architecture (SEDA) you have dedicated threads that do one particular task (for example, reading from a socket or evaluating a map.get) and thread pools that manage these task groups.
```
pveentjer, line 9: manage these threads.

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+You can enable Thread-Per-Core (TPC) on clients and cluster members to improve system performance, especially on machines with either fewer or more CPU cores than the recommended hardware listed on the xref:ROOT:production-checklist.adoc[]. 
+
+[[tpc-what]]
+== What is Thread-Per-Core (TPC)?
+
+The threading model is responsible for networking/storage and compute, which means it has a big impact on system performance. With Staged Event-Driven Architecture (SEDA) you have dedicated threads that do one particular task (for example, reading from a socket or evaluating a map.get) and thread pools that manage these task groups.  
+
+This architecture can lead to throughput and latency issues on machines with limited cores (fewer than 8), as the frequent context switching and state sharing can add significant overhead. Conversely, on machines with a very high number of CPUs (more than 32), a bottleneck can be created by the limited number of I/O threads vs. operation threads. Typically, smaller servers are more readily available through virtualization and for cloud-specific workloads, which can limit system performance in a SEDA environment.
```
pveentjer, line 11: SEDA can lead to ..

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+A Thread-Per-Core (TPC) design uses one thread for every CPU core and every thread does everything. With each TPC thread able to do networking/storage and compute, scaling is mostly a case of increasing the number of threads to spread the work. This means you can scale up and down easily based on the number of available cores, and make more efficient use of resources on machines with either a low or high number of cores. Every task is non-blocking and you can scale vertically much easier, which means high throughput, low latency and significantly improved system performance.  
+
+When TPC is enabled, regardless of the routing mode (smart or unisocket), the client will establish connections to all cluster members. If the server side has also enabled TCP, the client will open connections to all cores of all members and route partition-specific requests to the correct cores.  TCP-enabled servers will continue using the same ports for discovery, which means there's no difference in how the cluster member list is created and TPC-aware servers will be backwards compatible with clients not using TPC.
```
pveentjer, line 15: Only in case of a smart client and TPC enabled, a client will connect to the configured number of cores (which by default is 1).

(The problem with connecting to all cores is that it leads to smaller packets and hence on cloud environments you quickly run into the packets per second limit).

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+TPC is disabled by default and must be enabled on each cluster member and client.  
+
+NOTE: For example configuration files (XML and YAML) with TPC enabled, see xref:configuration:configuring-declaratively.adoc[].  
+
+=== Enabling TPC on Members
+
+To enable TPC on a Hazelcast cluster member, you need to set the `tpc` system property to true.  
+
+You can do this by using the following command:
```
pveentjer, line 28: command --> system property.

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
```diff
@@ -0,0 +1,92 @@
...
++
+[source,yaml]
+----
+ tpc:
+    enabled: true
+    eventloop-count: 12 <1>
+----
+====
+
+<1> The eventloop-count will default to the number of available CPU cores.
```
oliverhowell, line 59: @pveentjer how is this set or not? What do we need to tell users here?

pveentjer, line 59: TPC can only be used by EE and the max number of cores is determined by the license.

It the license allows for more eventloops than there are cores, then on every core there will be an eventloop. But it could be that the license limits the maximum number of eventloops used.

E.g. license says 10 eventloop and you have 16 cpu's, you only get 10 eventloops.

You are allowed to create less eventloops than there are cores. In some cases this can lead to a performance improvement due to better batching and less packet processing overhead.

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
```diff
@@ -0,0 +1,92 @@
...
+----
+====
+
+<1> The eventloop-count will default to the number of available CPU cores.  
+
+=== Enabling TPC on Clients
+
+To enable TPC on a Hazelcast client, you need to set the `tpc` system property to true.  
+
+You can do this by using the following command:
```
pveentjer, line 65: command -> System property

pveentjer: If you configure a client with TPC and the server doesn't support it, it is ignored.

If you configure a client without TPC and the server supports it, TPC is ignored.

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
pveentjer: Typically it is best to let the system decide. But keep in mind that the license can reduce the numberof allowed eventloops. SO it could be you have 64 cores and you configured 64 eventloops and your license allows for 16.. then 16 it is.

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
oliverhowell: Thanks for context - can you use GH suggest changes to suggest how above text should change or does below look reasonable?

> When TPC is enabled, smart clients will connect to the configured number of cores (default is 1) but unisocket clients will establish connections to all cluster members (which can reduce performance in cloud environments). If the server side has also enabled TPC, the client will open connections to all cores of all members and route partition-specific requests to the correct cores.  TPC-enabled servers will continue using the same ports for discovery, which means there's no difference in how the cluster member list is created and TPC-aware servers will be backwards compatible with clients not using TPC.

Where would users change `configured number of cores (default is 1)`?

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
oliverhowell: I was following this kind of example as convention - is this not correct?

> For example, the following command sets the cluster.name variable to test, using a system property.
> 
> `-Dcluster.name=test`

https://docs.hazelcast.com/hazelcast/5.3/configuration/configuring-declaratively#variables

PR #1006, opened 2024-02-12
Add default WAN republishing behaviour information

JamesHazelcast:
It turns out there is no documentation about the default state of WAN's republishing feature (that I could find). I've also updated the Javadocs in this PR: https://github.com/hazelcast/hazelcast-mono/pull/728


rebekah-lawrence (approved):
LGTM

Even if it does exist elsewhere (I couldn't find it either), this provides the information in-context.

Serdaro (approved):

PR #1009, opened 2024-02-14
Fix `reflections` terminology

JackPGreen:
Reflections (not reflections) is the [terminology used with Java](https://www.oracle.com/technical-resources/articles/java/javareflection.html).


rebekah-lawrence (approved):
LGTM

Serdaro:
The conflict is caused by a source file renaming performed during the day. 
@JackPGreen , if it'd be simpler, we can close this one and apply the fix.

PR #1014, opened 2024-02-15
Document that a non-PK column shouldn't be specified as the id-column [HZ-4211]

Edizc:
Added Microsoft SQL Server and Oracle to the supported databases and added a warning for declaring non-pk columns as the id-column.

### DIFF
--- a/docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
+++ b/docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
@@ -257,7 +257,7 @@ mapConfig.setMapStoreConfig(mapStoreConfig);
 ====
 
 |[[id-column]]`id-column`
-|Name of the column that contains the primary key. Do not specify a column without a primary key constraint as the id-column.
+|Name of the column that contains the primary key. Do not specify a column without a unique key constraint as the id-column.
 
 |id
 |
--- a/docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
+++ b/docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
@@ -257,7 +257,7 @@ mapConfig.setMapStoreConfig(mapStoreConfig);
 ====
 
 |[[id-column]]`id-column`
-|Name of the column that contains the primary key.
+|Name of the column that contains the primary key. A column without a primary key constraint should not be specified as the id-column.
 
 |id
 |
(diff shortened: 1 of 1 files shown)


docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
```diff
@@ -257,7 +257,7 @@ mapConfig.setMapStoreConfig(mapStoreConfig);
 ====
 
 |[[id-column]]`id-column`
-|Name of the column that contains the primary key.
+|Name of the column that contains the primary key. A column without a primary key constraint should not be specified as the id-column.
```
rebekah-lawrence, line 260:
```suggestion
|Name of the column that contains the primary key. Do not specify a column without a primary key constraint as the id-column.
```

rebekah-lawrence: Can we make the statement stronger, as suggested above, or is it only a recommendation (as implied by "should")?

rebekah-lawrence (commented):
A suggestion with a related question

docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
Edizc: Yes it would be better like this.

rebekah-lawrence (approved):
LGTM

AleksPeychev (approved):

docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
```diff
@@ -257,7 +257,7 @@ mapConfig.setMapStoreConfig(mapStoreConfig);
 ====
 
 |[[id-column]]`id-column`
-|Name of the column that contains the primary key.
+|Name of the column that contains the primary key. Do not specify a column without a primary key constraint as the id-column.
```
frant-hartm, line 260: @Edizc does it really need to be primary key constraint? Wouldn't limiting to any unique key be enough?

frant-hartm (changes requested):

docs/modules/mapstore/pages/configuring-a-generic-mapstore.adoc
AleksPeychev: @frant-hartm you are right. I have tested it and limiting to any unique key is enough.

AleksPeychev:
Fixes/Documents https://github.com/hazelcast/hazelcast/issues/26223

Serdaro (approved):

PR #1018, opened 2024-02-15
Docs 769: Add remaining Portable deprecations

oliverhowell:


rebekah-lawrence (approved):
LGTM

At least your updates; didn't check the duplicates, although they do seem to already be merged as you noted. I don't see why they are showing a difference in that case - just because they aren't present on this branch, I guess.....

PR #1021, opened 2024-02-19
Write fields in alphabetical order when using GenericRecord and compact serialization for the same object [API-2215]

ihsandemir:
Added the warning about the write order of the fields for compact serialization when the genericrecord is also used for the same object.

Also, added the missing deprecation notice to the `Creating New Portable Objects` section.

fixes https://github.com/hazelcast/hazelcast-mono/issues/714

### DIFF
--- a/docs/modules/clusters/pages/accessing-domain-objects.adoc
+++ b/docs/modules/clusters/pages/accessing-domain-objects.adoc
@@ -41,6 +41,8 @@ With the introduction of `GenericRecord`, User Code Deployment should be used on
 
 == Creating New Portable Objects
 
+CAUTION: Portable Serialization has been deprecated. We recommend you use Compact Serialization as Portable Serialization will be removed as of version 7.0.
+
 To create a `GenericRecord` object in portable format, use the `GenericRecordBuilder.portable()` method:
 
 [source,java]
@@ -73,6 +75,8 @@ GenericRecord namedRecord = GenericRecordBuilder.compact("employee")
 
 Note that there is no need to create a class definition, or a schema in this case. A schema will be created
 from the fields of the builder automatically.
+
+WARNING: If you want to use the same object using the `GenericRecord` and the compact serializer, then you should write your object fields in the `CompactSerializer.write` method in the alphabetical order. Otherwise, the fields will not be matched correctly.
 
 == Adding and Changing Values in Domain Objects
 
--- a/docs/modules/clusters/pages/accessing-domain-objects.adoc
+++ b/docs/modules/clusters/pages/accessing-domain-objects.adoc
@@ -80,7 +80,7 @@ GenericRecord namedRecord = GenericRecordBuilder.compact("employee")
 Note that there is no need to create a class definition, or a schema in this case. A schema will be created
 from the fields of the builder automatically.
 
-WARNING: If you want to use the same object using the `GenericRecord` and the compact serializer, then you should write your object fields in the `CompactSerializer.write` method in the alphabetical order. Otherwise, the fields will not be matched correctly.
+WARNING: To use the same object when using `GenericRecord` and the compact serializer, write your object fields in alphabetic order in the `CompactSerializer.write` method. If this is not done, the fields are not correctly matched.
 
 == Adding and Changing Values in Domain Objects
 
--- a/docs/modules/clusters/pages/accessing-domain-objects.adoc
+++ b/docs/modules/clusters/pages/accessing-domain-objects.adoc
@@ -80,7 +80,7 @@ GenericRecord namedRecord = GenericRecordBuilder.compact("employee")
 Note that there is no need to create a class definition, or a schema in this case. A schema will be created
 from the fields of the builder automatically.
 
-WARNING: To use the same object when using `GenericRecord` and the compact serializer, write your object fields in alphabetic order in the `CompactSerializer.write` method. If this is not done, the fields are not correctly matched.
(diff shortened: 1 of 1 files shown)


docs/modules/clusters/pages/accessing-domain-objects.adoc
```diff
@@ -41,6 +41,8 @@ With the introduction of `GenericRecord`, User Code Deployment should be used on
 
 == Creating New Portable Objects
 
+CAUTION: Portable Serialization has been deprecated. We recommend you use Compact Serialization as Portable Serialization will be removed as of version 7.0.
```
rebekah-lawrence, line 44:
```suggestion
CAUTION: 
.Deprecation Notice for Portable Serialization
====
Portable Serialization has been deprecated. We recommend you use Compact Serialization as Portable Serialization will be removed as of version 7.0.
====
```

rebekah-lawrence: The amended format is the one used in other instances of this caution

docs/modules/clusters/pages/accessing-domain-objects.adoc
```diff
@@ -74,6 +76,8 @@ GenericRecord namedRecord = GenericRecordBuilder.compact("employee")
 Note that there is no need to create a class definition, or a schema in this case. A schema will be created
 from the fields of the builder automatically.
 
+WARNING: If you want to use the same object using the `GenericRecord` and the compact serializer, then you should write your object fields in the `CompactSerializer.write` method in the alphabetical order. Otherwise, the fields will not be matched correctly.
```
rebekah-lawrence, line 79:
```suggestion
WARNING: To use the same object when using `GenericRecord` and the compact serializer, write your object fields in alphabetic order in the `CompactSerializer.write` method. If this is not done, the fields are not correctly matched.
```

rebekah-lawrence (changes requested):

rebekah-lawrence (approved):
LGTM

Serdaro (approved):
LGTM

docs/modules/clusters/pages/accessing-domain-objects.adoc
```diff
@@ -74,6 +80,8 @@ GenericRecord namedRecord = GenericRecordBuilder.compact("employee")
 Note that there is no need to create a class definition, or a schema in this case. A schema will be created
 from the fields of the builder automatically.
 
+WARNING: To use the same object when using `GenericRecord` and the compact serializer, write your object fields in alphabetic order in the `CompactSerializer.write` method. If this is not done, the fields are not correctly matched.
```
srknzl, line 83: I think more detailed description will be better. The users might think that, my serializer works like that so I can ignore this warning. 

It's the same object always. The binary changes. The change of binary is especially important when compact object is used as a key (or any partition related Data). The Data mismatch only occurs when there are multiple variable size fields and the order is different between the two serializers. To avoid that, we advice them to write their explicit serializer's variable size fields in alphabetical order.

rebekah-lawrence:
```suggestion
WARNING: If you use a compact object as a key, or any partition-related data, data mismatches can occur when the binary changes. Such data mismatches occur only when there are multiple variable size fields, and different ordering in the serializers. To avoid mismatches in your data, write the explicit variable size fields for your serializers in alphabetic order.
```

rebekah-lawrence: Just a suggestion to try and capture @srknzl 's comment

srknzl: Couple change suggestions. Feel free to improve it further

WARNING: If you use a compact object as a key, or any partition-related data; and you use Compact Generic Records and Compact explicit serializers together, data mismatches can occur. Such data mismatches occur only when there are multiple variable size fields, and different ordering in the serializers. Compact Generic Record serializer built-in to Hazelcast writes variable size fields in the alphabetical order. To avoid mismatches in your data, write the explicit variable size fields for your serializers in alphabetic order.

rebekah-lawrence: I have edited the original suggestion above - @srknzl is this better?

ihsandemir: Detailed explanation is fine but isn't it a little too much detail which may confuse the users? I understand the fact that we want to identify that there are use cases where you can use genericrecord and the compact without alphabetical order and it works but do we really want to promote this? I wanted to keep it simple and limited. If you are using GenericRecord please use the fields as ordered when you write your compact serializer. what is your take on this @rebekah-lawrence ? Simplicity vs. detailed?

rebekah-lawrence: Simple is always best, as long as all the information that the user requires can be provided in a simple manner!

Suggestion for a simpler version provided below, I'll leave it to you to discuss and decide which is the better fit.

rebekah-lawrence:
```suggestion
WARNING: Data mismatches can occur if you use both generic and specific Compact serializers. For example, the Compact Generic Record serializer, which is included with Hazelcast, writes variable size fields in alphabetic order based on the field name. To avoid potential mismatches in your data, write explicit variable size fields for Compact serializers in alphabetic order.
```

srknzl: It should be both simple and detailed. That's why we provide the situation where it can occur. If that's not the case for them, they wont read the rest.

srknzl: I don't like advicing them to do something without them knowing why.

srknzl: I think the latest one is pretty good, ty @rebekah-lawrence

srknzl (approved):

PR #1022, opened 2024-02-19
Add CPMap to Spring Resources [HZ-4196]

gbarnett-hz:
Add `CPMap` to the set of configurables in the Spring configuration section with an example.


rebekah-lawrence (approved):
LGTM

PR #1026, opened 2024-02-23
DOCS-719: TPC updates

oliverhowell:
Closes DOCS-719 - draft PR for now while I work through comments on initial PR (might be cleaner to just take the page and nav changes into a clean commit off main as the Beta changes are being pulled in...)

### DIFF
--- a/docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
+++ b/docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
@@ -41,7 +41,7 @@ XML::
 [source,xml]
 ----
 <tpc enabled="true">
-    <eventloop-count>12</eventloop-count>
+    <eventloop-count>12</eventloop-count> <1>
 </tpc>
 ----
 --
@@ -52,11 +52,11 @@ YAML::
 ----
  tpc:
     enabled: true
-    eventloop-count: 12
+    eventloop-count: 12 <1>
 ----
 ====
 
-NOTE: The eventloop-count will default to the number of available CPU cores.  
+<1> The eventloop-count will default to the number of available CPU cores.  
 
 === Enabling TPC on Clients
 
--- a/docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
+++ b/docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
@@ -1,0 +1,92 @@
+= Thread-Per-Core (TPC)
+:description: Find out more about Thread-Per-Core (TPC) and how to enable this feature on clients and cluster members.
+
+You can enable Thread-Per-Core (TPC) on clients and cluster members to improve system performance, especially on machines with either fewer or more CPU cores than the recommended hardware listed on the xref:ROOT:production-checklist.adoc[]. 
+
+[[tpc-what]]
+== What is Thread-Per-Core (TPC)?
+
+The threading model is responsible for networking/storage and compute, which means it has a big impact on system performance. With Staged Event-Driven Architecture (SEDA) you have dedicated threads that do one particular task (for example, reading from a socket or evaluating a map.get) and thread pools that manage these task groups.  
+
+This architecture can lead to throughput and latency issues on machines with limited cores (fewer than 8), as the frequent context switching and state sharing can add significant overhead. Conversely, on machines with a very high number of CPUs (more than 32), a bottleneck can be created by the limited number of I/O threads vs. operation threads. Typically, smaller servers are more readily available through virtualization and for cloud-specific workloads, which can limit system performance in a SEDA environment. 
+
+A Thread-Per-Core (TPC) design uses one thread for every CPU core and every thread does everything. With each TPC thread able to do networking/storage and compute, scaling is mostly a case of increasing the number of threads to spread the work. This means you can scale up and down easily based on the number of available cores, and make more efficient use of resources on machines with either a low or high number of cores. Every task is non-blocking and you can scale vertically much easier, which means high throughput, low latency and significantly improved system performance.  
+
+When TPC is enabled, regardless of the routing mode (smart or unisocket), the client will establish connections to all cluster members. If the server side has also enabled TCP, the client will open connections to all cores of all members and route partition-specific requests to the correct cores.  TCP-enabled servers will continue using the same ports for discovery, which means there's no difference in how the cluster member list is created and TPC-aware servers will be backwards compatible with clients not using TPC.
+
+[[tpc-config]]
+== Configuration Options
+
+TPC is disabled by default and must be enabled on each cluster member and client.  
+
+NOTE: For example configuration files (XML and YAML) with TPC enabled, see xref:configuration:configuring-declaratively.adoc[].  
+
+=== Enabling TPC on Members
+
+To enable TPC on a Hazelcast cluster member, you need to set the `tpc` system property to true.  
+
+You can do this by using the following command:
+
+```
+-Dhazelcast.internal.tpc.enabled=true
+```
+
+Or by setting these variables within your configuration file:
+
+[tabs] 
+==== 
+XML:: 
++ 
+-- 
+[source,xml]
+----
+<tpc enabled="true">
+    <eventloop-count>12</eventloop-count>
+</tpc>
+----
+--
+
+YAML::
++
+[source,yaml]
+----
+ tpc:
+    enabled: true
+    eventloop-count: 12
+----
+====
+
+NOTE: The eventloop-count will default to the number of available CPU cores.  
+
+=== Enabling TPC on Clients
+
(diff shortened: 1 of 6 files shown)


docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
```diff
@@ -0,0 +1,95 @@
...
+A Thread-Per-Core (TPC) design uses one thread for every CPU core and every thread does everything. With each TPC thread able to do networking/storage and compute, scaling is mostly a case of increasing the number of threads to spread the work. This means you can scale up and down easily based on the number of available cores, and make more efficient use of resources on machines with either a low or high number of cores. Every task is non-blocking and you can scale vertically much easier, which means high throughput, low latency and significantly improved system performance.  
+
+When TPC is enabled, smart clients will connect to the configured number of cores (default is 1) but unisocket clients will establish connections to all cluster members (which can reduce performance in cloud environments). If the server side has also enabled TPC, the client will open connections to all cores of all members and route partition-specific requests to the correct cores.  TPC-enabled servers will continue using the same ports for discovery, which means there's no difference in how the cluster member list is created and TPC-aware servers will be backwards compatible with clients not using TPC.
```
pveentjer, line 16: Uniisocket will completely ignore TPC. Unisocket client will connect to random member in cluster.

pveentjer, line 16: "If the server side has also enabled TPC, the client will open connections to all cores of all members and route partition-specific requests to the correct cores."

If TPC is enabled on both client and server, by default the client will connect directly to 1 of the tpc threads instead of using the legacy network threads.

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
```diff
@@ -0,0 +1,95 @@
...
+
+[[tpc-config]]
+== Configuration Options
+
+TPC is disabled by default and must be enabled on each cluster member and client.  
+
+NOTE: If you configure a client with TPC and the server doesn't support it, the client is ignored. If you configure a client without TPC and the server supports it, TPC is ignored (server and client will still communicate).
```
pveentjer, line 23: 'the client is ignored' --> The client will not make use of TPC.

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
oliverhowell: So you _must_ use smart clients with TPC?

oliverhowell: > configured number of cores (default is 1)

Where/how is this configured? Is there a client TPC config too? https://github.com/hazelcast/hazelcast/blob/v5.3.6/hazelcast/src/main/resources/hazelcast-client-full-example.xml only has enabled or not. We have eventloop at the server which defaults to 12 but this seems to relate to something else...

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
oliverhowell:
```suggestion
NOTE: If you configure a client with TPC and the server doesn't support it, or if you configure a client without TPC and the server supports it, TPC will be ignored in both cases (server and client will still communicate).
```

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
oliverhowell:
```suggestion
When TPC is enabled, smart clients will connect to the configured number of cores (default is 1). If TPC is enabled on both client and server, the client will connect directly to one of the TPC threads instead of using the legacy network threads. TPC-enabled servers will continue using the same ports for discovery, which means there's no difference in how the cluster member list is created and TPC-aware servers will be backwards compatible with clients not using TPC.

NOTE: Unisocket clients will ignore TPC and connect to a random member in the cluster. You must use smart clients with TPC.
```

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
```diff
@@ -0,0 +1,95 @@
+= Thread-Per-Core (TPC)
+:description: Find out more about Thread-Per-Core (TPC) and how to enable this feature on clients and cluster members.
+:page-enterprise: true
+
+You can enable Thread-Per-Core (TPC) on clients and cluster members to improve system performance, especially on machines with either fewer or more CPU cores than the recommended hardware listed on the xref:ROOT:production-checklist.adoc[].
```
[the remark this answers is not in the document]
oliverhowell:
```suggestion
Enable Thread-Per-Core (TPC) on both clients and cluster members to improve system performance, especially on machines with either fewer or more CPU cores than the recommended hardware listed on the xref:ROOT:production-checklist.adoc[]. 
```

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
pveentjer: In the CLientTpcConfig there is a connection count:

```
public final class ClientTpcConfig {

    private boolean enabled;

    private int connectionCount;```

So if the server has configured e.g. 12 (that should not be a default), the client can pick between -1 and 12.

```

    /**
     * Sets the number of connections to TPC ports offered by a Hazelcast member.
     *
     * <ol>
     *     <li>If set to a negative value, an IllegalArgumentException will be
     *     thrown.</li>
     *     <li>If set to 0, the client will connect to every TPC port.
     *     </li>
     *     <li>If set to the same number as returned by the server, the client
     *     will connect to every TPC port.
     *     </li>
     *     <li>If set to a number larger than 0 and smaller than the number of
     *     returned TPC ports, the client will randomize the list of ports and
     *     make the configured number of connections.
     *     </li>
     *     <li>If set to a number larger than the number of TPC ports, the client
     *     will connect to each tpc port (has same effect as configuring 0)
     *     </li>
     * </ol>
     * Increasing the number of connections leads to more packets with a smaller
     * payload and this can lead to a performance penalty. Also in cloud environments
     * e.g. AWS there can be a packets per second limit (pps) and it pretty easy
     * to run into this limit if an equal number of connections is created as
     * TPC ports on the server and a lot of small interactions are done e.g. a
     * map.get with small payloads.
     *
     * @param connectionCount throws IllegalArgumentException when connectionCount
     *                        is negative.
     * @since 5.4
     */
    public ClientTpcConfig setConnectionCount(int connectionCount) {
        this.connectionCount = checkNotNegative(connectionCount, "connectionCount");
        return this;
    }```

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
pveentjer: Yes. If either side doesn't have TPC configured, then the client will not use TPC.

Internal:

If TPC on the server is enabled, the server sends back a list of TPC ports the client can pick from.

If TPC on the server is disabled, this list is empty and the client has nothing to pick from and hence will only use the classic networking.

docs/modules/cluster-performance/pages/thread-per-core-tpc.adoc
pveentjer: Correct. TPC is only supported for smart clients (so the non unisocket ones)

oliverhowell: Should we link to the API docs for this then? i.e. https://docs.hazelcast.org/docs/latest/javadoc/com/hazelcast/config/tpc/package-summary.html

I'm not sure if you want to expand the current paragraph we're reviewing or not. Based on what I understand from above I would either remove the (default is 1) comment or expand it and link to the API docs. WDYT?

pveentjer: I would provide an explanation of this property and you can use the javadoc as a source of inspiration.

oliverhowell:
```suggestion
When TPC is enabled, smart clients will connect to the number of cores specified in the client config. If TPC is enabled on both client and server, clients connect directly to one of the TPC threads instead of using the legacy network threads. TPC-enabled servers continue to use the same ports for discovery, which means there's no difference in how the cluster member list is created and TPC-aware servers are backward compatible with clients that don't use TPC.

NOTE: Unisocket clients will ignore TPC and connect to a random member in the cluster. You must use smart clients with TPC. For further information on the `ClientTpcConfig` class used to specify the number of cores to use for connections, refer to https://docs.hazelcast.org/docs/{full-version}/javadoc/com/hazelcast/client/config/ClientTpcConfig.html[Class ClientTpcConfig] in the Java API documentation.
```

oliverhowell:
Superseded by https://github.com/hazelcast/hz-docs/pull/1035 so closing without merging...

PR #1045, opened 2024-03-06
Correct UCN Namespace definition key

JamesHazelcast:
The `user-code-namespace` key was changed to `namespace` before release as the `user-code-` prefix is redundant when part of the parent `user-code-namespaces` property anyway.


rebekah-lawrence (approved):
Thanks James! LGTM

PR #1072, opened 2024-04-08
Externalise OS/EE Hazelcast edition names

JackPGreen:
The names of the editions of Hazelcast should not be hardcoded, to allow adjusting the edition names to suit business requirements, localization etc.

I've updated references to `Open Source` and `Enterprise` to a key to an externalised value, where sensible. The actual names remain unchanged, as updating these terms is [larger than just this repo](https://hazelcast.slack.com/archives/C035HQET5/p1712585262831789?thread_ts=1712583856.778069&cid=C035HQET5).

I've **not** done this for:
- Code samples
- URLs
- Examples of Hazelcast outputs
- Hazelcast configuration properties
- Migration guides from older versions (e.g. IMDG, Jet) where the names can be hardcoded as they are correct _for that version_
- The `ENTERPRISE` suffix on EE-only pages (this is coming from CSS somewhere)
   - image

[Slack discussion](https://hazelcast.slack.com/archives/C035HQET5/p1712583856778069).


rebekah-lawrence (commented):
The impact of discrepancies in code (examples/properties and such like) ad the docs is an interesting thought - we'll need to come up with a plan for that (a) code is also updated (b) the docs have to explain each time that where the code says x, the docs say y (c) a better approach that didn't come immediately to mind.

There's also the question of whether we want the variable to be "Hazelcast x" or just "x", and how we use "Platform" - does that mean both Community & Enterprise (and is it Edition or edition?); not to mention how we use Hazelcast on its own, which seems to be a different way of saying Platform.....  Although this isn't really achieved in the web content either, from https://hazelcast.com/community-edition-projects/: For mission critical applications, we strongly recommend **Hazelcast Enterprise edition**. 
Have questions about changes to the **Community Edition** in **Hazelcast Platform** 5.4 and 5.5? 

However, these existing inconsistencies can be dealt with using this approach once we know how we want to use each term. And it'll all improve the docs. 

Nice work - thanks.

JackPGreen:
> There's also the question of whether we want the variable to be "Hazelcast x" or just "x"

There’s an awkward overlap between:
- Hazelcast, the software company
- Hazelcast, the data platform

the terms are used interchangeably in the documentation, but should be split (even if they refer to the same display value) to support changing names (eg IMDG) or white labelling by a different vendor (eg Acme Premium).

JackPGreen:
@Serdaro @rebekah-lawrence just wondered if you'd had a chance to have a look at this?

rebekah-lawrence:
Looks fine to me, in the implementation and coverage of hz-docs.

However, I think @Serdaro needs to review/approve this one; the mere adding of the variables here, but not in the hazelcast-docs antora playbooks could have a negative effect on the build, I'm not sure, but I suspect there has to be some tie-in work there.....

Serdaro:
Will check before the week ends.

Serdaro:
This test PR looks good, thanks a lot @JackPGreen . Once the naming is decided, I can update the "Enterprise" labels accordingly (it is a JS). Also, if we put the attributes in the "/docs/antora.yml" file here, we won't need to deal with the hazelcast-docs repo playbook files.

JackPGreen:
> ### **For testing only** I've renamed the products to `Free` and `Premium` to allow it to be tested via the [netlify preview](https://deploy-preview-1072--hardcore-allen-f5257d.netlify.app/).
> This will be reverted before merge.

> This test PR looks good, thanks a lot @JackPGreen

> Looks fine to me, in the implementation and coverage of hz-docs.

Ok, as your happy, I've updated the product names to be correct to be suitable for merging.

132ef50dc722612c2e4637832700be00fd32215e

 > Once the naming is decided, I can update the "Enterprise" labels accordingly (it is a JS).

I don't _think_ they need to change (unless they can be tied to these properties also?).
We aren't actually renaming Enterprise so it's fine.

> Also, if we put the attributes in the "/docs/antora.yml" file here, we won't need to deal with the hazelcast-docs repo playbook files.

Not sure I understand, sorry?

Serdaro:
> > Also, if we put the attributes in the "/docs/antora.yml" file here, we won't need to deal with the hazelcast-docs repo playbook files.
> 
> Not sure I understand, sorry?
I'll take care of this one, if/when needed. It is an Antora (our site generator) configuration thing.

Serdaro (approved):

rebekah-lawrence (approved):
LGTM

PR #1082, opened 2024-04-16
Remove `$` from Docker getting started IP address

JackPGreen:
This is the only reference to `$<host_ip>`, everywhere else uses `<host_ip>`. The `$` shouldn't be included in the command, otherwise it'll break.


rebekah-lawrence (approved):
LGTM

PR #1099, opened 2024-04-30
Fix XML/YAML for UCN

Serdaro:
As per https://hazelcast.slack.com/archives/C01JU7ZJYGP/p1714463053524479


olukas (approved):
Thanks @Serdaro !

PR #1103, opened 2024-05-01
Documentation for LTS Upgradeability

rebekah-lawrence:
Initial draft, based on PRD and TDD, and the resulting Documentation Plan


docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -107,6 +134,12 @@ include::clients:partial$rest-deprecation.adoc[]
 
 The cluster is now operating on the new codebase version: Both the codebase version and the cluster version have been upgraded.
 
+=== Restart Members on Previous Version
+
+If you experience any issues during the rolling upgrade, you can recover to the previous version by restarting the members.
```
srknzl, line 139: we are testing on between 5.2-5.4 source version to 5.5 target version upgrades. So should we add a note about this is supported only for LTS upgrades? In the future (post 5.5) though, it'll be supported for one-minor upgrades as well as I know.

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -220,3 +253,14 @@ binary and algorithmic compatibility and some other steps.
 
 It is worth mentioning that a business app upgrade is orthogonal to a
 rolling member upgrade. A rolling business app upgrade may be done without upgrading the members.
+
+**Are features marked as Beta in my current version upgraded?**
```
srknzl, line 257: well, in theory they are upgraded. The important part about them is that features marked as Beta in a minor version are only supported in a cluster of the same minor version. So during upgrades, errors can happen. 

For example, if you upgrade compact from 5.2 -> 5.3 there are binary format changes, but only in some field types: Date Time etc. If the user does not use those fields, there'll be no fails. Otherwise, they'll fail. Also all the features that depend on SQL can fail in the same way.

but yes, after these fail, any new usages of the SQL will have the latest features available in the target release (GA version)

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -220,3 +253,14 @@ binary and algorithmic compatibility and some other steps.
...
+
+Features marked as Beta are not upgraded. This is due to changes that can occur between Beta and general availability releases. If a Beta feature has been updated or made generally available in the target release, that version of the feature is included in the upgraded feature set.
+
+**Can I upgrade directly to the latest release?**
+
+You can upgrade directly to the latest release in the following circumstances:
+
+* Your current version is no more than one minor version below the current release
+* The latest release is an xref:migrate:lts.adoc[LTS release] and your current version is among those supported
```
srknzl, line 266: yes, and fyi the current supported version will be only the previous LTS and the previous minor in the future. For now it's 5.2-5.4 for 5.5.

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,121 @@
+= Long-term Support Releases
```
srknzl, line 1: not strong on this but we can do this
```suggestion
= Long-term Support (LTS) Releases
```

for this to appear in navigation

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+:description: Hazelcast simplifies your upgrade experience with the introduction of long-term support (LTS) releases. You can upgrade directly from a supported previous release to the LTS release or directly between LTS releases using a rolling upgrade.
+:page-enterprise: true
+
+{description}
+
+LTS releases have been introduced with 5.5. You can upgrade directly from 5.2 or later to 5.5 without the need to upgrade to each release between your current version and the LTS release.
+
+NOTE: Any features that are marked as Beta in your current version, or any intermediate version, are not upgraded. This is due to changes that can occur between the Beta and general availability releases. If a Beta feature has been updated or made generally available in the LTS releae, that feature is included in the LTS release. For further information on rolling upgrade compatibilities, see xref:maintain-cluster:rolling-upgrades.adoc#hazelcast-members-compatibility-guarantees[Compatibility Guarantees].
```
srknzl, line 10: note: if you upgrade the beta question part in the rolling-upgrades page, you should also change here.

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,121 @@
...
+|Focus on stability
+|Focus on innovation
+
+|Consolidates the functionality introduced in all STS releases since the last LTS release. Can include improvements, such as increased capacity limits
+|Provides the latest features and bug fixes
+
+|Released every two years
+|Released at least twice each year
+
+|Upgrade directly from the previous LTS, or from the last available STS release
```
srknzl, line 27:
```suggestion
|Upgrade directly from the previous LTS, or from the last available STS before the LTS release
```

If we don't add this, is this ambiguous? for example, last STS is 5.7 and last LTS is 5.5, then is "upgrade" (well actually downgrade :D) from 5.7 -> 5.5 possible? Probably, I am overthinking it.

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -48,8 +60,23 @@ As a result, you can use rolling upgrades to upgrade clusters to the following m
 +
 This also includes the ability of rolling upgrade from Hazelcast IMDG 4.2 to Platform 5.0, since IMDG is part of the Platform and 5.0 is considered to be the next minor version after IMDG 4.2.
 
+- Releases from 5.2 to 5.4, directly to the 5.5 LTS release. For further information on additional compatibilities for direct-to-LTS rolling upgrades, see xref:migrate:lts.adoc[].
```
AleksPeychev, line 63: Upon first reading it was not clear to me that 5.3 is also supported. My suggestion is : Releases 5.2 , 5.3 and 5.4 can be upgraded directly to the 5.5 LTS release.

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
AleksPeychev: I am a little confused with `current version` and `current release`. Is it going to be better if `source version` and `target version` are used?

AleksPeychev: Suggesting using `source` and `target` wherever possible. e.g. Your source version is no more than one minor version below the current release (target)

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -5,6 +5,18 @@
...
 {description}
 
+== Best Practices
+
+Before starting a rolling upgrade, consider the following best practises:
+
+* Ensure that the <<hazelcast-members-compatibility-guarantees,Compatibility Guarantees>> cover your requirements
+* If the minimum version of JDK has changed between releases, <<updated-jdk,upgrade the JDK version>> before starting a rolling upgrade
+* If you used Hazelcast Operator to deploy your cluster, refer to the link:https://docs.hazelcast.com/operator/latest/scaling-upgrading#upgrading[Operator documentation, window=_blank]
+* If using Hazelcast Management Center to trigger your rolling upgrade, upgrade Management Center before upgrading your members. For information on upgrading Management Center, refer to the link:https://docs.hazelcast.com/management-center/5.4/getting-started/install[Management Center documentation, window=_blank]
```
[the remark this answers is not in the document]
rebekah-lawrence: This is the Installing and Starting page, under the Getting Started section; the Upgrading page only lists the changelog, so as far as I can see, this is the only relevant page....

I've also added a link to the Operator docs for K8s users

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -5,6 +5,18 @@
...
+* If using the `hz-cluster-admin` script to update your cluster, enable the `CLUSTER_WRITE` endpoints in the REST API. For further information on the REST API, see xref:maintain-cluster:rest-api.adoc#using-the-rest-endpoint-groups[REST API]
+* Plan sufficient time for the migration and upgrade to complete, and remember that members cannot join or leave during the upgrade
```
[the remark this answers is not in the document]
rebekah-lawrence: reworded to "the members cannot change during the upgrade"

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
rebekah-lawrence: I've reworded, let me know if it still needs work!

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
rebekah-lawrence: Amended

docs/modules/migrate/pages/lts.adoc
rebekah-lawrence: It is best to avoid acronyms in headings where possible - if you can provide a reason to include it here, I won't argue; however, I would prefer not to, the acronym is given in the description and I believe most people would be aware of it

docs/modules/migrate/pages/lts.adoc
rebekah-lawrence: amended

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,121 @@
...
+|5.2
+|5.3 member joins 5.2 cluster
+
+|Further member tries to join cluster
+|LTS
+|5.2
+|LTS member identifies that two different minor versions are already in the cluster and rejects joining the cluster
+|===
+
+The following scenario is not supported:
```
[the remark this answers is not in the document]
rebekah-lawrence: It comes from the TDD: https://hazelcast.atlassian.net/wiki/spaces/PM/pages/4903993364/TDD+LTS+Upgradability#Mixed-version-clusters

It is a difficult once to cover in a positive way, I have amended, but any suggestions are more than welcome!

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -48,8 +60,23 @@ As a result, you can use rolling upgrades to upgrade clusters to the following m
...
+With the introduction of Hazelcast's first LTS release, you can upgrade directly from 5.2, or later, to 5.5. This means that, if using a supported earlier version, you do not need to upgrade to each version between your current version and the LTS release.
+
+For more information on LTS releases, and any specific considerations when upgrading, see xref:migrate:lts.adoc[].
+
+[[updated-jdk]]
+== JDK Version Upgrade
+
+As newer JDK releases contain improvements, we recommend that you upgrade to a newer JDK version where xref:deploy:versioning-compatibility.adoc#supported-java-virtual-machines[supported].
+
+Upgrade the JDK version before upgrading your members if support for the JDK version has changed between your current and planned cluster versions. For example, in 5.2 JDK 8 was supported, but 5.5 supports only JDK 17 or later.
```
frant-hartm, line 78:
```suggestion
Upgrade the JDK version before upgrading your members if support for the JDK version has changed between your current and planned cluster versions. For example, in 5.2 JDK 8 or later was supported, but 5.5 supports only JDK 17 or later. 
```
To make it cleare that there is an overlap of supported JDK versions.

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,128 @@
+= Long-term Support Releases
+
+:description: Hazelcast simplifies your upgrade experience with the introduction of long-term support (LTS) releases. You can upgrade directly from a supported previous release to the LTS release or directly between LTS releases using a rolling upgrade.
+:page-enterprise: true
+
+{description}
+
+LTS releases have been introduced with 5.5. You can upgrade directly from 5.2 or later to 5.5 without the need to upgrade to each release between your current version and the LTS release.
```
frant-hartm, line 8: Should we specify that they should upgrade from the latest patch of each release line? We only test latest patch (at the time of the release).

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,128 @@
...
+.. Upgrade the member's codebase
+.. Start the member and wait for it to join the cluster
+
+. Trigger the rolling upgrade on the cluster
+
+For further information on performing a rolling upgrade, including the steps to take, compatibility guarantees and any constraints, see xref:maintain-cluster:rolling-upgrades.adoc[]. 
+
+=== Restart Members on Previous Version
+
+If you experience any issues during the rolling upgrade, you can recover to the previous version by restarting the members.
```
frant-hartm, line 126: This can be done only if the version hasn't been upgraded (either automaticaly or via API/script).

frant-hartm (changes requested):

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,128 @@
...
+|5.4
+|All features
+
+|5.2
+
+5.3
+
+a|* IMap data structure. This includes the API, configuration options, and persistence
+* CP subsystem. This inclues the API and peristence
+* WAN replication. Ongoing replication is supported during the upgrade. This includes both source and target; however, the source and the target cannot be upgraded simultaneously
```
frant-hartm, line 56: We tested replication to clusters 5.2+ during the RU, not to earlier clusters. So we should specify we support only that.

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -5,6 +5,18 @@
 
 {description}
 
+== Best Practices
+
+Before starting a rolling upgrade, consider the following best practises:
```
oliverhowell, line 10:
```suggestion
Before starting a rolling upgrade, consider the following best practices:
```

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -5,6 +5,18 @@
...
+
+* Ensure that the <<hazelcast-members-compatibility-guarantees,Compatibility Guarantees>> cover your requirements
+* If the minimum version of JDK has changed between releases, <<updated-jdk,upgrade the JDK version>> before starting a rolling upgrade
+* If you used Hazelcast Operator to deploy your cluster, refer to the link:https://docs.hazelcast.com/operator/latest/scaling-upgrading#upgrading[Operator documentation, window=_blank]
```
oliverhowell, line 14: Trying to avoid a big naming thread but should 1st ref be Hazelcast Platform Operator? I might be out of date on this...

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -5,6 +5,18 @@
...
+* If using Hazelcast Management Center to trigger your rolling upgrade, upgrade Management Center before upgrading your members. For information on upgrading Management Center, refer to the link:https://docs.hazelcast.com/management-center/5.4/getting-started/install[Management Center documentation, window=_blank], or if you deployed Management Center using the Hazelcast Operator, refer to the link:https://docs.hazelcast.com/operator/latest/deploy-management-center.adoc[Operator documentation, window=_blank]
+* If using the `hz-cluster-admin` script to update your cluster, enable the `CLUSTER_WRITE` endpoints in the REST API. For further information on the REST API, see xref:maintain-cluster:rest-api.adoc#using-the-rest-endpoint-groups[REST API]
```
oliverhowell, line 16: Which REST API page should this be? The new enterprise-rest-api or the old legacy one (you've linked here)

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
```diff
@@ -48,8 +60,23 @@ As a result, you can use rolling upgrades to upgrade clusters to the following m
 +
 This also includes the ability of rolling upgrade from Hazelcast IMDG 4.2 to Platform 5.0, since IMDG is part of the Platform and 5.0 is considered to be the next minor version after IMDG 4.2.
 
+- Releases from 5.2 to 5.4, directly to the 5.5 Long-term Support (LTS) release. For further information on additional compatibilities for direct-to-LTS rolling upgrades, see xref:migrate:lts.adoc[].
```
oliverhowell, line 63:
```suggestion
- Directly to the 5.5 Long-term Support (LTS) release from releases 5.2 through 5.4. For further information on additional compatibilities for direct-to-LTS rolling upgrades, see xref:migrate:lts.adoc[].
```
Reads a little awkwardly with the source releases at front, just a suggestion...

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,130 @@
+= Long-term Support Releases
+
+:description: Hazelcast simplifies your upgrade experience with the introduction of long-term support (LTS) releases. You can upgrade directly from a supported previous release to the LTS release or directly between LTS releases using a rolling upgrade.
+:page-enterprise: true
+
+{description}
+
+LTS releases have been introduced with 5.5. You can upgrade directly from 5.2.x or later to 5.5 without the need to upgrade to each release between your current version and the LTS release. We recommend that you upgrade to 5.5 from the latest available patch for your current release.
```
oliverhowell, line 8: I know the context is already Platform but with the various product/version combinations in this section, is it worth being explicit with?
 
```suggestion
LTS releases have been introduced with Hazelcast Platform 5.5. You can upgrade directly from 5.2.x or later to 5.5 without the need to upgrade to each release between your current version and the LTS release. We recommend that you upgrade to 5.5 from the latest available patch for your current release.
```

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,130 @@
...
+|Focus on innovation
+
+|Consolidates the functionality introduced in all STS releases since the last LTS release. Can include improvements, such as increased capacity limits
+|Provides the latest features and bug fixes
+
+|Released every two years
+|Released at least twice each year
+
+|Upgrade directly from the previous LTS, or from the STS release that precedes the latest LTS release
+|Upgrade from last available STS release only
```
oliverhowell, line 28: Minor style query but period after 1st para?

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,130 @@
...
+For example, if there are three STS releases between LTS releases, you cannot upgrade to the latest LTS unless you have upgraded to all three STS releases 
+
+|Full support for three years with the option of extended support for a further year
+|Maintenance support until the next STS or LTS release only
+
+|Choose if you value stability over new feature adoption, you want to upgrade less frequently, and you want full or extended support
+|Choose if you value new features over stability, you are willing to upgrade several times a year, and you are happy with maintenance support for the life of the STS
+|===
+
+== Hazelcast 5.5 LTS
```
oliverhowell, line 39: Is it just Hazelcast 5.5 in this context?

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,130 @@
...
+[cols="1,1"]
+|===
+|Upgrade From|LTS Release Support
+
+|5.2
+
+5.3
+
+a|* IMap data structure. This includes the API, configuration options, and persistence
+* CP subsystem. This inclues the API and peristence
```
oliverhowell, line 52:
```suggestion
* CP subsystem. This includes the API and peristence
```

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,130 @@
...
+* WAN replication. Ongoing replication is supported during the upgrade. This includes both source and target; however, the source and the target cannot be upgraded simultaneously
+
+NOTE: Replication has been tested for 5.2 or later clusters only.
```
oliverhowell, line 55: But context for this line is only versions 5.2 and 5.3 anyway - does this add anything?

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,130 @@
...
+|5.3 member joins 5.2 cluster
+
+|Further member tries to join cluster
+|LTS
+|5.2
+|LTS member identifies that two different minor versions are already in the cluster and rejects joining the cluster
+|===
+
+The scenario outlined below can occur because the logic for a mixed-version cluster is introduced in the first LTS release and cannot be enforced by earlier releases.
+However, the scenario is not supported and results from such a cluster would not be reliable.
```
oliverhowell, line 88:
```suggestion
However, this scenario is not supported and results from such a cluster would not be reliable. 
```

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,130 @@
...
+.. Upgrade the member's codebase
+.. Start the member and wait for it to join the cluster
+
+. Trigger the rolling upgrade on the cluster
+
+For further information on performing a rolling upgrade, including the steps to take, compatibility guarantees and any constraints, see xref:maintain-cluster:rolling-upgrades.adoc[]. 
+
+=== Restart Members on Previous Version
+
+If an automatic upgrade, an upgrade using the API, or using a script has not successfully upgraded the version, you can recover to the previous version by restarting the members.
```
oliverhowell, line 128:
```suggestion
If an automatic upgrade, or an upgrade using the API or script, has not successfully upgraded the version, you can recover to the previous version by restarting the members.
```

oliverhowell (approved):
Coupla minor typos and a few minor suggestions and comments but pre-approving

docs/modules/migrate/pages/lts.adoc
```diff
@@ -0,0 +1,130 @@
...
+|Focus on stability
+|Focus on innovation
+
+|Consolidates the functionality introduced in all STS releases since the last LTS release. Can include improvements, such as increased capacity limits
+|Provides the latest features and bug fixes
+
+|Released every two years
+|Released at least twice each year
+
+|Upgrade directly from the previous LTS, or from the STS release that precedes the latest LTS release
```
srknzl, line 27: > Upgrade directly from the previous LTS, or from the STS release that precedes the latest LTS release

@frant-hartm should this be "Upgrade directly from the previous LTS" only? I think for 5.5 we are making an exception and make it work for "Upgrade directly from the previous LTS, or from the STS release that precedes the latest LTS release"

srknzl (approved):
lgtm from my perspective

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
rebekah-lawrence: Dagnamit! When will those US English practitioners learn how to use words properly ;-)

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
rebekah-lawrence: The (last) rename that I saw dropped Platform to make it just Hazelcast Operator - and this also matches the [current] naming in the side nav of the docs.

Actually, I think the last one I saw was Hazelcast **Enterprise** Operator, but Platform was definitely not part of the name.

I suggest leaving as is - if it needs to be changed, we can do that anytime.

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
rebekah-lawrence: @frant-hartm @srknzl can you guide me on this one? Ta!

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
rebekah-lawrence: Changed to: The 5.5 Long-term Support (LTS) release direct from 5.2.x to 5.4.x.

Which is actually better for list parallelism as well as readability! Good catch.

docs/modules/migrate/pages/lts.adoc
rebekah-lawrence: Seems this is an area where Google and I part ways - added periods to all table items.

Can't insist on a style guide if I don't use it myself - even if it disagrees with me;-)

docs/modules/migrate/pages/lts.adoc
rebekah-lawrence: Thank you!

docs/modules/migrate/pages/lts.adoc
rebekah-lawrence: Added from a comment from @frant-hartm - but a fair point.

What do you think, @frant-hartm, or did I miss the point of your original comment?

docs/modules/migrate/pages/lts.adoc
rebekah-lawrence: But surely there is no previous STS or LTS for 5.5 - and we have covered this in the remaining text? 

This table covers the logic/behaviour for LTS/STS releases as they will be - it was difficult to convey the idea without getting bogged down in the initial implementation where there are no other LTS or STS releases avaialbe.

We could be specific for this release and say something like:

> For the initial LTS release with 5.5, you can upgrade directly from 5.2.x or later. 
> 
> In future LTS releases, you can <existing wording>

Of course, this then raises the question of whether we need to add something like (when available) or (available from release ????) in the STS header. 

Then this additional wording can be stripped as we progress through the availability of STS releases and further LTS releases.

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
frant-hartm: The new REST API supports changing the version, so it should be the new one. However, I don't see the change version endpoint in the docs yet.

docs/modules/migrate/pages/lts.adoc
frant-hartm: When there is WAN replication you have 2 clusters and 3 different versions.

*from* version of cluster A (any of 5.2, 5.3, 5.4), 
*to* version of cluster A 

version of cluster B - this could be any Hazelcast version 4.0-5.4. 

We tested only 5.2, 5.3 and 5.4 for cluster B.

rebekah-lawrence: Expanded to "NOTE: Hazelcast has tested WAN Replication to validate the upgrade solution. The test covered two clusters of different versions; for scenarios where WAN Replication is used from a cluster with a version or 5.2, 5.3, or 5.4 to another cluster with a version between 4.0 and 5.4, only the replication to clusters of version 5.2 or later have been tested."

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
rebekah-lawrence: I've added a link to the Enterprise REST API topic, and added Community to the existing link; if the new REST API adds a better place to link to, we can adjust it later.

frant-hartm (approved):

docs/modules/maintain-cluster/pages/rolling-upgrades.adoc
srknzl: Hey @rebekah-lawrence  I'm informed that old REST api is deprecated in the upcoming 5.5 release, so we should start referring to the new documentaiton instead. 

There's a swagger API documentation page with all endpoints. The new REST API's reference manual may be a bit missing right now @frant-hartm

rebekah-lawrence: @srknzl yes, I added a link to the new REST API as well as the existing in the final updates. The swagger docs are going to be embedded in the docs, I believe. If the link needs to be updated before 5.5 goes live I will take care of it. 

Thanks for  mentioning it, though, just to keep it fresh in my mind - I've added a note to the docs Jira to double-check the link destinations once the REST API doc updates are merged.

PR #1111, opened 2024-05-14
Update outdated infor about client versions

frant-hartm:
The correct state is
- 3.6+ clients can connect to 3.6+ members, but not to 4.x or 5.x
- 4.x and 5.x clients can connect to 4.x and 5.x members

I am not sure we want to talk about 3.x at all here so simple update should be fine.


rebekah-lawrence (approved):
LGTM

PR #1117, opened 2024-05-17
Update capacity-planning.adoc

naadhira:
Added a few words to clarify the point that this calculation is only for the HybridLog requirement, and that the startup sanity check also includes the amount allocated at the data structure configuration level. It also clarifies the point that memory capacity planning includes hybridlog and the data itself - which should be obvious but isn't explicitly stated.


rebekah-lawrence (approved):
LGTM, thanks Hendy!

PR #1119, opened 2024-05-22
DOCS-822: minor text tweaks and backport

oliverhowell:


docs/modules/ROOT/pages/index.adoc
```diff
@@ -2,11 +2,9 @@
 :!page-pagination:
 :description: Hazelcast Platform uniquely combines a distributed compute engine and a fast data store in one runtime. It offers unmatched performance, resilience and scale for real-time and AI-driven applications. 
 
-{description} 
-
 Hazelcast is a distributed computation and storage platform for consistently
 low-latency querying, aggregation and stateful computation against event
-streams and traditional data sources. It allows you to quickly build
+streams and traditional data sources. {description} It allows you to quickly build
```
rebekah-lawrence, line 7: interesting, I've never seen {description} used twice on a page; does this have any implications?

docs/modules/ROOT/pages/index.adoc
```diff
@@ -131,6 +127,8 @@ among the available cluster members. They can contain hundreds or thousands of d
 depending on the memory capacity of your system. Hazelcast also automatically creates backups of these partitions
 which are also distributed in the cluster. This makes Hazelcast resilient to data loss.
 
+NOTE: “Node” and “Member” are interchangeable, and both mean a Java Virtual Machine (“JVM”) on which one or more instances of Hazelcast software are in operation.
```
rebekah-lawrence, line 130:
```suggestion
NOTE: _Node_ and _Member_ are interchangeable, and both mean a Java Virtual Machine (JVM) on which one or more instances of Hazelcast are in operation.
```

rebekah-lawrence (approved):
Couple of things for consideration - one question and some removal of double quotes ;-)

I also removed _software_ - not sure it adds anything, or that I've seen it anywhere else; by _Hazelcast_ do we mean _Hazelcast Platform_, or _Hazelcast clusters/cluster members_ or something else here? I'd suggest it could possibly be recast to _one or more Hazelcast clusters_, but that would need to be checked by ??? for technical correctness in this context.....

Approved to keep it moving and because I am being picky about existing content!

docs/modules/ROOT/pages/index.adoc
oliverhowell: I took the first instance out and just plopped it in the middle of the existing description. Made more sense rather than looking like two descriptions but will get revised no doubt later in this cycle when we have a landing page template.

PR #1143, opened 2024-06-20
Fixing variable name error

rebekah-lawrence:
Spotted that I had used the wrong variable name - so fixing quickly.


rebekah-lawrence:
As @oliverhowell  is tied up at the off-site, I am breaking process and merging without review - just fixes for already-broken variables visible on live site.

PR #1145, opened 2024-06-24
OS to EE Rolling Restarts Doc Updates

rebekah-lawrence:
New information and cross-references added to document the OS to EE rolling restart mission-driven cycle.

NOTE: Updates will be required for the Release notes and What's New pages, as with other features; these files will be updated separately when all information is populated.

NOTE: The update made to the data-migration-tool.adoc file has also been made locally to the updated version in https://github.com/hazelcast/hz-docs/pull/936. I will ensure that any further changes are rolled into that file too so it does not affect things when that one is merged.

### DIFF
--- a/docs/modules/migrate/pages/community-to-enterprise.adoc
+++ b/docs/modules/migrate/pages/community-to-enterprise.adoc
@@ -1,0 +1,52 @@
+= Move to {enterprise-product-name} from {open-source-product-name}
+:description: With Enterprise Edition, you can benefit from additional and extended features, patch releases, Hazelcast Support, and more connectors. If you are using the Open Source Edition and want to upgrade to Enterprise, you can do this with clusters of the same version from 5.5.
+
+{description}
+
+The {open-source-product-name} Edition does not include the following, which are be included with an Enterprise license:
+
+* Long term support (LTS) releases and short term support (STS) releases. For more information on LTS and STS releases, see xref:migrate:lts.adoc[]
+
+* Data Migration Tool. For more information on the Data Migration Tool, see xref:migrate:data-migration-tool.adoc[]
+* Patch releases, including those for security vulnerabilities from March 2024
+* Hazelcast Support with defined SLAs. For more information on the benefits of Hazelcast Support, see xref:getting-started:support.adoc#customer-support[]
+
+NOTE: To use the following {enterprise-product-name} features they might need to be enabled on your {enterprise-product-name} license, and must be configured after all members have been moved to {enterprise-product-name}. 
+For further information on the license requirements and configuration, see the linked documentation for the specific feature.
+
+* Security persistence. For more information on storing sensitive data on disk, see xref:secure-cluster:hardening-recommendations.adoc[]
+* CP Subsystem. For more information on building a strongly consistent layer for a set of distributed data structures that withstands server and client failures, see xref:cp-subsystem:cp-subsystem.adoc[]
+* High-Densisty Memory Store (HD Memory). For more information on Hazelcast's in-memory storage solution, which supports predictable application scaling and improved performance while minimizing pauses caused by garbage collection, see xref:storage:high-density-memory.adoc[]
+* WAN Replication. For more information on synchronizing the state of multiple Hazelcast clusters across a WAN, see xref:wan:wan.adoc[]
+* Rolling upgrades. For more information on upgrading the members of a cluster without downtime, see xref:maintain-cluster:rolling-upgrades.adoc[]
+* Thread-Per-Core (TPC). For more information on improving the system performance by using one thread for networking, storage, and compute on each CPU, see xref:cluster-performance:thread-per-core-tpc.adoc[]
+* User Code Namespaces. For more information on using a container for Java classpath resources, such as user code and accompanying artifacts, see xref:clusters:user-code-namespaces.adoc[]
+* Helm charts, with the exception of link:https://github.com/hazelcast/charts/tree/master/stable/hazelcast[Commmunity-driven Open Source Helm Charts, window=_blank]. For more information on using the Helm package manager for Kubernetes with Hazelcast, see xref:kubernetes:deploying-in-kubernetes.adoc[]
+* Updates to connectors available before version 5.4
+* Any connectors added after version 5.4 
++
+For more information on Connectors, see xref:integrate:connectors.adoc[].
+
+In addition, if you are using our tools, {open-source-product-name} does not include the following:
+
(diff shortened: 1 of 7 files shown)


docs/modules/migrate/pages/rolling-restart.adoc
```diff
@@ -0,0 +1,45 @@
+= Upgrade Cluster to {enterprise-product-name}
+:description: You can use an in-place rolling restart to move each member on your Open Source cluster to an Enterprise cluster.
```
frant-hartm, line 2: This wording is a bit weird. 

Maybe something like
```suggestion
:description: You can use an in-place rolling restart to change each member to upgrade your cluster from Open Source to Enterprise cluster.
```

docs/modules/migrate/pages/rolling-restart.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+{description}
+
+This approach ensures that you can move your cluster without interruption to the service. 
+Similar to the rolling upgrade process, a rolling restart allows you to move one member at a time.
+
+The codebase of your {open-source-product-name} cluster must match the target {enterprise-product-name} release. 
+Currently, rolling restart can only be used to update your 5.5 {open-source-product-name} cluster members.
```
frant-hartm, line 10: You can replace 5.5 with current version.

Going forwards, it will be possible to update each version from OS to EE.

docs/modules/migrate/pages/rolling-restart.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+== Best Practice
+
+Before starting a rolling restart, consider the following best practices:
+
+* Backup your IMap and ReplicatedMap data structures. This is not essential, but it is recommended practice
```
frant-hartm, line 16: How should they do the backup? AFAIK we don't provide any tooling for that. (DMT can transfer data between clusters, but not simply export...)

docs/modules/migrate/pages/rolling-restart.adoc
```diff
@@ -0,0 +1,45 @@
...
+* If the {open-source-product-name} member is a different version to the target {enterprise-product-name} member, update the codebase to match the target {enterprise-product-name} release
```
frant-hartm, line 17: Wouldn't this defeat the whole purpose of the OS to EE restart?

Say (in the future) user is running 5.5.x OS cluster, and they want to upgrade to 5.6.x EE.

This says to update OS cluster to 5.5 first to 5.6, and then do OS to EE restart.
But they can't update the cluster without losing the data.

What they need to do is OS to EE restart first on 5.5 and the do RU from 5.5 to 5.6.

docs/modules/migrate/pages/rolling-restart.adoc
```diff
@@ -0,0 +1,45 @@
...
++
+**[RJL: Do we need to include any additional information to cover moving from - for example - 5.3 with CP Subsystem to Enterprise 5.5?]**
```
frant-hartm, line 19: This was removed from the scope, currently there is no good solution for that.

docs/modules/migrate/pages/rolling-restart.adoc
```diff
@@ -0,0 +1,45 @@
...
++
+During migrations, members cannot join or leave the cluster.
+
+. Upgrade the member's codebase to match the {enterprise-product-name} version.
+
+. Start the member and wait until it joins the {enterprise-product-name} cluster.
+
+. Repeat this process for each member in the {open-source-product-name} cluster.
+
+. When all members on the {open-source-product-name} cluster have been moved to the {enterprise-product-name} cluster, enable and configure the
```
frant-hartm, line 42: I would link dynamic configuration here. It's the only well supported way to enable EE features on a running cluster.

frant-hartm (changes requested):

docs/modules/migrate/pages/community-to-enterprise.adoc
```diff
@@ -0,0 +1,52 @@
+= Move to {enterprise-product-name} from {open-source-product-name}
+:description: With Enterprise Edition, you can benefit from additional and extended features, patch releases, Hazelcast Support, and more connectors. If you are using the Open Source Edition and want to upgrade to Enterprise, you can do this with clusters of the same version from 5.5.
```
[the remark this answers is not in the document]
rebekah-lawrence: Yup, unfortunately, the variable does not seem to resolve when used in the :description: - even when using the pass:q[<description>] description.

docs/modules/migrate/pages/rolling-restart.adoc
rebekah-lawrence: changed to:

You can upgrade your Open Source cluster to an Enterprise cluster using an in-place rolling restart to change each member in turn.

docs/modules/migrate/pages/rolling-restart.adoc
rebekah-lawrence: I have removed the line, as I think it is covered by the previous statement on matching codebases.

The issue is - if we change to current version - this could be interpreted as their OS current version, so any version, which would be incorrect.

Feel free to disagree ;-)

docs/modules/migrate/pages/rolling-restart.adoc
rebekah-lawrence: I have added "using your usual method", but if this is too much of a fudge (or not something that is generally/easily done) I'd suggest removing this bullet from the list.

docs/modules/migrate/pages/rolling-restart.adoc
rebekah-lawrence: Possibly my misunderstanding (and related to the comment above over 5.5 v current version, I understood that the OS members need to be 5.5 before they can be moved.

If I was incorrect, this bullet can be removed. Otherwise, we need to clarify somehow that the OS must be on 5.5; and that statement would need to be removed for the next release.

docs/modules/migrate/pages/rolling-restart.adoc
rebekah-lawrence: Not ideal. And I'm not sure we can manage that expectation through the docs without it - removed question; but I'd note there are likely to be support calls around this if we cannot provide the solution 😬

docs/modules/migrate/pages/rolling-restart.adoc
frant-hartm: Good.

docs/modules/migrate/pages/rolling-restart.adoc
frant-hartm: I suppose that's good.

frant-hartm (approved):

PR #1152, opened 2024-07-01
DOC-65: replicate new nav for 5.4

oliverhowell:


rebekah-lawrence (approved):
LGTM

I don't see anything obvious missing from the nav, and random selections from nav and x-refs seem to be correct. Build errors are only those expected.

PR #1153, opened 2024-07-01
Update protocol documentation

srknzl:
See  [here](https://github.com/hazelcast/hazelcast-mono/commit/77b1e0eec4691a2e452eb0da5e38d8d60025baf7#diff-37d9504ce60ecdf27711f36de75d440107d6ef841648903f4baf2df8344bac3aL56) where default TLSv1.2 is changed.

And the current master has TLS [here](https://github.com/hazelcast/hazelcast-mono/blob/9102331bb4025eb793f405aeaff05ae96c15edfd/hazelcast-enterprise/src/main/java/com/hazelcast/internal/nio/ssl/SSLEngineFactorySupport.java#L76)

Also SSL and TLSv1 and TLSv1.1 is old according to @kwart so I removed them

### DIFF
--- a/docs/modules/security/pages/integrating-openssl.adoc
+++ b/docs/modules/security/pages/integrating-openssl.adoc
@@ -198,7 +198,7 @@ Available values are:
 require it
 * `ciphersuites`: Comma-separated list of cipher suite names allowed to be used.
 * `protocol`: Name of the algorithm which is used in your TLS/SSL. Its default
-value is `TLSv1.2`. Available values are:
+value is `TLS`. Available values are:
 ** `TLS`
 ** `TLSv1`
 ** `TLSv1.1`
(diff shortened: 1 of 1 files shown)


rebekah-lawrence (approved):
LGTM

Please do not merge at the moment, however, there are some issues on the doc site which are being investigated. I will confirm when resolved and merging can be started again.

kwart (changes requested):
Serkan, thanks for identifying and fixing the issue. Could you extend the PR and rephrase the whole `protocol` bullet? The supported values depend on the underlying engine, and the current versions allow neither `SSL*` nor `TLSv1` / `TLSv1.1`. 

I mean something like:
```
* `protocol`: Name of the TLS algorithm the engine uses. The default
value is empty, meaning we don't specify the exact protocol version. Available values depend on the engine implementation but usually support:
** `TLS` (equivalent to the empty value)
** `TLSv1.2`
** `TLSv1.3`
For the
`protocol` property, we recommend you provide TLS with its version
information, e.g., `TLSv1.3`. It ensures the engine won't allow a fallback to an old, insecure version.
```

kwart (approved):

PR #1162, opened 2024-07-04
New UCN YAML format updates

JamesHazelcast:
Following changes in https://github.com/hazelcast/hazelcast-mono/pull/2446 these config examples need to be updated.

I believe this will need to be backported to `5.4` docs as well, but @Fly-Style can confirm.

### DIFF
--- a/docs/modules/clusters/pages/ucn-member-side.adoc
+++ b/docs/modules/clusters/pages/ucn-member-side.adoc
@@ -60,7 +60,8 @@ user-code-namespaces:
       blacklist:
         class:
           - com.acme.app.BeanComparator
-   myNameSpace: []
+   name-spaces:
+      myNameSpace: []
 ----
 ====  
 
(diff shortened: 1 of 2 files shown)


Fly-Style (approved):
This is a new and a better way to deal with UCN in YAML.
Re: 5.4 version -- we need to explicitly say that new format is applicable only starting from 5.4.1.

rebekah-lawrence (approved):
I've added the backport to 5.4 label - once merged, I'll do a quick update to the 5.4 branch to add a note saying: You can only define multiple namespaces in the YAML configuration file from 5.4.1.

Alternatively, we could try a condition in this file that only displays in the 5.4 version - I've never tried this, so can't be sure it will work - but the following _should_ work:

ifdef::5.4[]
NOTE: You can only define multiple namespaces in the YAML configuration file from 5.4.1.
endif::[]

LGTM

JamesHazelcast:
> Alternatively, we could try a condition in this file that only displays in the 5.4 version - I've never tried this, so can't be sure it will work - but the following _should_ work:
> 
> ifdef::5.4[] NOTE: You can only define multiple namespaces in the YAML configuration file from 5.4.1. endif::[]

This sounds fancy, but not a risk I want to take if it's never been tried before 😄 I'm afraid I'll leave that with you!

PR #1179, opened 2024-07-15
Update java.adoc around client routing

jgardiner68:
Updated the headings to include "Client" as a prefix and "Routing" so that customers can read the announcement and find it in the documentation

### DIFF
--- a/docs/modules/clients/pages/java.adoc
+++ b/docs/modules/clients/pages/java.adoc
@@ -87,7 +87,7 @@ The client is updated whenever the cluster version, partition groups, or CP grou
 
 From {enterprise-product-name} 5.5, you can use any of the following cluster routing modes:
 
-* **ALL_MEMBERS** 
+* **Client ALL_MEMBERS Routing** 
 +
 The default mode, and is the equivalent of the legacy **Smart** client operation mode. 
 +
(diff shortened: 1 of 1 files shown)


rebekah-lawrence:
Have to say I am not totally on board with this for the following reasons:

- These are not headings; this is the list of the options; the options are ALL_MEMBERS, SINGLE_MEMBER, and MULTI_MEMBER
- The heading is **Java Client Cluster Routing Modes** to try and bridge the gap (in the best way James Holgate and I could come up with, but still a terrible noun beast kind of way) - but the code is `cluster-routing`, so it is difficult to avoid using cluster routing when documenting it
- There is explanatory text, and a tie in with the previous client operation mode settings 

Questions:
a) Why does the announcement use different naming from the docs (which were written to match the extensive renaming exercise)?
b) Can't the announcement just link directly to https://docs.hazelcast.com/hazelcast/latest/clients/java#java-client-operation-modes so customers don't have to "find" the information?
c) Did you expect me to suddenly become easier to deal with in my last week ;-)

jgardiner68:
Two reasons to do so:
1) Feedback via PSB members is some of them thought this feature relates cluster member routing.
2) improves search

Point (c) -- yes always surprising :-)

rebekah-lawrence:
The potential for this confusion was discussed during the renaming rounds. Disappointing it still remains. This is really not a docs issue (whistles innocently)

I winced and used "Java client cluster routing mode" after the discussions in a desperate and unsatisfying attempt to cover this.

Tell you what I'll do, I'll skulk round the corner, and @oliverhowell can approve while I'm not looking - don't say I'm not good to you - but use this wisely; I still have more days to go and you won't get another pass 🤓

oliverhowell:
I've updated some of the references to include the suggested text so we have both sides covered: https://github.com/hazelcast/hz-docs/pull/1289

Closing this PR...

PR #1188, opened 2024-07-19
Add Note for REST dependency conflicts [DEX-210]

sumnerib:


oliverhowell (approved):
(Normally we'd replace 'utilize' with 'use' but in the spirit of getting this merged quickly...)

PR #1189, opened 2024-07-22
Fix jetty plugin documentation link

srknzl:
The old link redirects to https://jetty.org/docs/

Feel free to change the new link, I'm not sure if we should give a link to 12. Using latest in the link does not work

Should be backported and forwardported to all versions imo.

### DIFF
--- a/docs/modules/plugins/pages/web-session-replication.adoc
+++ b/docs/modules/plugins/pages/web-session-replication.adoc
@@ -31,5 +31,4 @@ Jetty based web session replication is offered through Hazelcastty Session Manag
 It is a container specific module that enables session replication for
 JEE Web Applications without requiring changes to the application.
 
-For further information, refer to https://jetty.org/docs/jetty/12/operations-guide/session/index.html#hazelcast[Persistent Sessions with Hazelcast^] in the Jetty documentation.
-for more details.
+For further information, refer to the HTTP Session Management topic in the relevant version of the jetty documentation. The related documentation for Jetty 12 can be found https://jetty.org/docs/jetty/12/operations-guide/session/index.html#hazelcast[here].
--- a/docs/modules/plugins/pages/web-session-replication.adoc
+++ b/docs/modules/plugins/pages/web-session-replication.adoc
@@ -31,5 +31,5 @@ Jetty based web session replication is offered through Hazelcastty Session Manag
 It is a container specific module that enables session replication for
 JEE Web Applications without requiring changes to the application.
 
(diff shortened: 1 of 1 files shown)


rebekah-lawrence:
Hi @srknzl I've added the back/forward port labels so the update is included in all versions.

On the link - is 12 the version Hz recommends/uses, or can any version be used? If the former, leave the direct link to 12; otherwise, just link to https://jetty.org/docs/ and let the user select the version they need.

srknzl:
> Hi @srknzl I've added the back/forward port labels so the update is included in all versions.
> 
> On the link - is 12 the version Hz recommends/uses, or can any version be used? If the former, leave the direct link to 12; otherwise, just link to https://jetty.org/docs/ and let the user select the version they need.

I don't think we have a version we recommend. However, the downside of providing link to https://jetty.org/docs/ is that the link is so general and not related to Hazelcast. We should additional instructions for user such as: "for documentation, search for Hazelcast in jetty docs" etc

wdyt?

rebekah-lawrence:
Looking at the jetty docs, Hazelcast is only specifically mentioned in 12 - so, it is probably best to keep the link direct to the pertinent information. 

Or, we fork their documentation and configure a latest path... 😇 🤣

docs/modules/plugins/pages/web-session-replication.adoc
```diff
@@ -31,5 +31,5 @@ Jetty based web session replication is offered through Hazelcastty Session Manag
 It is a container specific module that enables session replication for
 JEE Web Applications without requiring changes to the application.
 
-See the https://www.eclipse.org/jetty/documentation/current/configuring-sessions-hazelcast.html[Jetty: Persistent Sessions with Hazelcast^]
+See the https://jetty.org/docs/jetty/12/operations-guide/session/index.html#hazelcast[Jetty: Persistent Sessions with Hazelcast^]
```
rebekah-lawrence, line 34:
```suggestion
For further information, refer to https://jetty.org/docs/jetty/12/operations-guide/session/index.html#hazelcast[Persistent Sessions with Hazelcast^] in the Jetty documentation.
```

rebekah-lawrence (changes requested):

srknzl:
@rebekah-lawrence I see https://jetty.org/docs/jetty/11/operations-guide/session/index.html#hazelcast as well. I think it's not only jetty 12 Hazelcast mentioned in

oliverhowell (approved):

PR #1194, opened 2024-07-24
REL-232: clean WN & RN changes into 5.5

oliverhowell:
- [x] 1st draft of What's New
- [x] Quick pull of RN issues
- [ ] Review and refine WN
- [ ] Add OS contributors to RN
- [ ] Do another scan of PRs that are missing label or have label added
- [ ] Review and refine RN
- [ ] Coordinate script and build variables


docs/modules/ROOT/pages/whats-new.adoc
```diff
@@ -1,63 +1,79 @@
...
+Before 5.5, Hazelcast clients could only have one of two types of connection to a cluster: single member (also known as unisocket) or all member. With a unisocket connection the client only connected to a single cluster member, which meant all communication was relayed via the connected cluster member and could therefore suffer from performance issues. With an all member connection the client connects to every member in the cluster and is partition-aware, which means that communications are sent directly to the cluster member holding the required data. If you're using partition groups to isolate data to specific cluster members, there was previously no way for the client to have visibility into this configuration and therefore was not connecting as efficiently as possible, or adapting to failovers. For example, CP Subsystem uses a group leader - clients would not be aware of this and there would an extra internal relay stage rather than sending directly to the group leader alone.
```
rebekah-lawrence, line 33:
```suggestion
Before 5.5, Hazelcast clients could only have one of two types of connection to a cluster: single member (also known as unisocket) or all member (also known as smart). With a unisocket connection the client only connected to a single cluster member, which meant all communication was relayed via the connected cluster member and could therefore suffer from performance issues. With an all member connection the client connects to every member in the cluster and is partition-aware, which means that communications are sent directly to the cluster member holding the required data. If you're using partition groups to isolate data to specific cluster members, there was previously no way for the client to have visibility into this configuration and therefore was not connecting as efficiently as possible, or adapting to failovers. For example, CP Subsystem uses a group leader - clients would not be aware of this and there would an extra internal relay stage rather than sending directly to the group leader alone.
```

rebekah-lawrence: Just to tie the two previous config options; not vital!

docs/modules/ROOT/pages/whats-new.adoc
```diff
@@ -1,63 +1,79 @@
...
+For more info, see xref:clients:java.adoc#java-client-cluster-routing-modes[].
 
-For more information on how to enable and configure TPC, see xref:cluster-performance:thread-per-core-tpc.adoc[].
+== Introduction of Long-term Support (LTS) releases
 
-== Tiered Storage now generally available (GA)
-[.enterprise]*Enterprise* 
+Hazelcast Platform 5.5 introduces Long-term Support (LTS) releases to simplify your upgrade experience and give you greater control over when and how you upgrade your software. You can perform upgrades with more precision and consistency, and with reduced effort and risk. You can upgrade directly from a supported previous release to the LTS release or directly between LTS releases using a rolling upgrade.
+
+You can upgrade from Hazelcast {open-source-product-name} to Hazelcast {enterprise-product-name} with zero downtime, or from Hazelcast 5.2.x or later to 5.5 without the need to upgrade to each release between your current version and the LTS release. We recommend that you upgrade to 5.5 from the latest available patch for your current release.
```
rebekah-lawrence, line 43: might be worth stipulating OS 5.5 to EE 5.5

rebekah-lawrence (commented):
just a couple of thoughts; overall looking good 👍

PR #1204, opened 2024-07-29
Add notice about memory leak in vector collections in beta version [AI-133]

k-jamroz:
Info and workaround for memory leaks described in https://github.com/hazelcast/hazelcast-mono/pull/2767

### DIFF
--- a/docs/modules/data-structures/pages/vector-collections.adoc
+++ b/docs/modules/data-structures/pages/vector-collections.adoc
@@ -695,3 +695,10 @@ As this is a beta version, Vector Collection has some limitations; the most sign
 3. The lack of fault tolerance, as backups cannot yet be configured. However, data in collections is migrated to other cluster members on graceful shutdown and a new member joining the cluster, which means that normal cluster maintenance (such as a rolling restart) is possible without data loss.
 4. Only on-heap storage of vector collections is available
 
+There is a known memory leak in Vector collection in some scenarios:
+
+1. Using `destroy` or `clear` on a non-empty Vector Collection
+2. Making Vector collection partition empty by removing the last entry using `deleteAsync` or `removeAsync` or overwriting it using one of the `put`/`set` methods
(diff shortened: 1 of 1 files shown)


k-jamroz:
@vbekiaris @oliverhowell I wondered if this should be mentioned more prominently, e.g. as warning or in release notes.

docs/modules/data-structures/pages/vector-collections.adoc
```diff
@@ -695,3 +695,11 @@ As this is a beta version, Vector Collection has some limitations; the most sign
...
 4. Only on-heap storage of vector collections is available
 
+There is a known memory leak in Vector collection in some scenarios:
+
+1. Using `destroy` or `clear` on a non-empty Vector Collection
+2. Making Vector collection partition empty by removing the last entry using `deleteAsync` or `removeAsync`
+3. Updating the only entry in Vector collection partition using one of the `put`/`set` methods
+4. Repeated migrations back and forth when a member is not restarted. This should not significantly affect rolling restart, provided sufficient heap margin. The leak may manifest itself when a subset of members is not restarted and the rest of them is repeatedly shutdown or restarted gracefully.
+
+The workaround in the first scenarios is to avoid those situations or restart affected cluster. For the last scenario workaround is to restart affected member or cluster. The restart can be graceful in which case no data should be lost.
```
oliverhowell, line 705:
```suggestion

== Known issue

There is currently a known issue that has potential for causing a memory leak in Vector collections in some scenarios:

1. Using `destroy` or `clear` on a non-empty vector collection.
2. Making a vector collection partition empty by removing the last entry using `deleteAsync` or `removeAsync`.
3. Updating the only entry in a vector collection partition using one of the `put`/`set` methods.
4. Repeating migrations back and forth when a member is not restarted. This should not significantly affect rolling restart, provided sufficient heap margin. The leak may manifest itself when a subset of members is not restarted and the rest of them are repeatedly shut down or restarted gracefully.

The workaround for scenarios 1 - 3 is to avoid those situations or restart the affected cluster. For scenario 4 the workaround is to restart the affected member or cluster. The restart can be graceful, which should not cause loss of data. 
```

oliverhowell (changes requested):
Thanks for raising this and sorry for delay - see suggested change and related PR for release notes

PR #1206, opened 2024-07-29
Use latest `checkout` action

JackPGreen:
Rather than hardcoding a specific (now outdated) version of `checkout`, we can specify the latest `4.x` as per everywhere else.


JackPGreen:
@oliverhowell bump :)

JackPGreen:
@oliverhowell bump :)

oliverhowell (approved):
Op, sorry for the delay and thanks for the bump

PR #1221, opened 2024-07-31
Update to connectors & data structures overview

avtarraikmo:


oliverhowell (approved):
Would suggest adding headings for each table (will populate right-hand TOC and help chunk/link the content) but approving in advance for this one.

e.g.
== AP Data Structures

PR #1231, opened 2024-08-01
Add more vector search tuning tips [AI-134]

k-jamroz:


docs/modules/data-structures/pages/vector-search-overview.adoc
```diff
@@ -235,4 +236,45 @@ To decrease pressure on heap memory, you can decrease the number of parallel mig
...
 2. Vector deduplication does not incur significant overhead for uploads (usually less than 1%) and searches. You may consider disabling it to get slightly better performance and smaller memory usage if your dataset does not contain duplicated vectors. However, be aware that in the presence of many duplicated vectors with deduplication disabled, a  similarity search may return poor quality results.
-
+3. For given query each vector index partition is searched by 1 thread. Number of concurrent partition searches is configured by specifying pool size for `hz:query` executor, which by default has 16 threads per member. In configurations with many cores it can be increased to fully utilize all available cores as follows:
```
oliverhowell, line 239:
```suggestion
3. For a given query, each vector index partition is searched by 1 thread. The number of concurrent partition searches is configured by specifying a pool size for `hz:query` executor, which by default has 16 threads per member. In configurations with many cores, it can be increased to fully utilize all available cores as follows:
```

docs/modules/data-structures/pages/vector-search-overview.adoc
```diff
@@ -235,4 +236,45 @@ To decrease pressure on heap memory, you can decrease the number of parallel mig
...
+----
+hazelcast:
+  ...
+  executor-service:
+    "hz:query":
+      pool-size: 100
+----
+====
++
+4. If there are fewer partitions than available cores, not all cores will be used for single search execution. This can be fine if you are focused on throughput as in general less partitions need less resources (mainly CPU). However, if you want to achieve best latency for single client, it is better to distribute the search to as many cores as possible which requires having at least as many partitions as cores in the cluster.
```
oliverhowell, line 280:
```suggestion
4. If there are fewer partitions than available cores, not all cores will be used for single search execution. This is ok if you are focused on throughput, as in general fewer partitions means you need less resources (mainly CPU). However, if you want to achieve the best latency for a single client, it is better to distribute the search to as many cores as possible, which requires having at least as many partitions as cores in the cluster.
```
How does the "mainly CPU" relate? You need less CPU resources or you need less resources because it's mainly CPU that is used? It's a bit ambiguous and could be clearer...

docs/modules/data-structures/pages/vector-search-overview.adoc
oliverhowell: Does TPC affect or change this guidance at all?

docs/modules/data-structures/pages/vector-search-overview.adoc
```diff
@@ -210,7 +210,8 @@ digraph singleStageSearch {
 The number of partitions has a big impact on the performance of the vector collection. The conflicting factors that can impact the selection of an optimal partition count are as follows:
 
 - *data ingestion*: a greater number of partitions results in improved parallelism, up to around the total number of partition threads in the cluster.
-  After this point, more partitions will not significantly improve ingestion speed.
+  After this point, more partitions will not significantly improve ingestion speed. 
+  If there are fewer partitions than number of cores, not all available resources will be utilized during ingestion because updates on given partition are executed by single thread.
```
oliverhowell, line 214:
```suggestion
  If there are fewer partitions than number of cores, not all available resources will be utilized during ingestion because updates on a given partition are executed by single thread. 
```

oliverhowell (approved):
Some minor suggestions/comments but approving in advance

docs/modules/data-structures/pages/vector-search-overview.adoc
k-jamroz: no, vector collection search will work in the same way with TPC. Search is not executed on partition thread. OTOH gets and puts should be able to benefit from direct routing to correct partition thread.

docs/modules/data-structures/pages/vector-search-overview.adoc
k-jamroz: search is CPU-bound. But I think this addition may be confusing and does not add much - removed

PR #1275, opened 2024-09-04
Update UCN notices to mention incompatibility with Jet

JamesHazelcast:
Removes the upgrade tip from the CDC Join tutorial (which uses Jet), and updates the migration tip to mention that Jet should continue to use UCD (for now).

This PR will be replaced in the future once HZG-88 is completed.


JamesHazelcast:
Hopefully a quick review when you have time @oliverhowell 😄

oliverhowell (approved):
One small suggestion but approving in advance

JamesHazelcast:
Thanks for the review @oliverhowell, glad to see the style guide being brought into force 👍

PR #1313, opened 2024-10-03
Clarify (lack of) backup redistribution on data structure level [HZG-156]

burakgok:
The info-box is intended to clarify the following sentence.
> The distribution happens on the partition level; the data and its backups are stored in the memory partitions.

If you don't like its position, you can move it at the end of the page.

### DIFF
--- a/docs/modules/fault-tolerance/pages/backups.adoc
+++ b/docs/modules/fault-tolerance/pages/backups.adoc
@@ -9,9 +9,9 @@ for more information about the partitioning.
 
 [IMPORTANT]
 ====
-Hazelcast does not support backup redistribution on a data structure level. Thus, changing the backup count of a nonempty data structure is **not** supported. For example, for a map with a nonzero backup count:
+Hazelcast does not support backup redistribution on a data structure level. Thus, changing the backup count of a nonempty data structure is **not** supported. For example, for a map:
 
-1. increasing the backup countfootnote:change-backup-count[The backup count of a map can only be changed by modifying the xref:configuration:understanding-configuration.adoc#static-configuration[static configuration] and restarting the cluster.] does **not** create additional copies of existing map entries,
+1. increasing the backup countfootnote:change-backup-count[The backup count of an empty map can only be changed by modifying the xref:configuration:understanding-configuration.adoc#static-configuration[static configuration] and restarting the cluster.] does **not** create additional copies of existing map entries,
 2. decreasing the backup countfootnote:change-backup-count[] does **not** remove any backups of existing map entries, so some backups will have stale data if the map entries are overwritten, and
 3. if some backups are lost due to node failure, additional copies are **not** created on other nodes to meet the backup count.
 ====
--- a/docs/modules/fault-tolerance/pages/backups.adoc
+++ b/docs/modules/fault-tolerance/pages/backups.adoc
@@ -7,6 +7,17 @@ The distribution happens on the partition level; the data and its backups are st
 memory partitions. See the xref:overview:data-partitioning.adoc[Data Partitioning] and xref:clusters:partition-group-configuration.adoc[Partition Grouping]
 for more information about the partitioning.
 
+[IMPORTANT]
+====
+Hazelcast does not support backup redistribution on data structure level. For example, for a map with a nonzero backup count,
+
+1. increasing the backup count does **not** create additional copies of existing map entries,
+2. decreasing the backup count does **not** remove any backups of existing map entries, so some backups will have stale data if the map entries are overwritten, and
+3. if some backups are lost due to node failure, additional copies are **not** created on other nodes to meet the backup count.
+
+Thus, it is unsupported to change the backup count of a data structure once it was populated with data.
+====
+
 When a member in your cluster is lost, Hazelcast redistributes the backups
 on the remaining members so that every partition has a backup.
 The number of backups is configurable.
--- a/docs/modules/fault-tolerance/pages/backups.adoc
+++ b/docs/modules/fault-tolerance/pages/backups.adoc
@@ -9,13 +9,11 @@ for more information about the partitioning.
 
 [IMPORTANT]
 ====
-Hazelcast does not support backup redistribution on data structure level. For example, for a map with a nonzero backup count,
+Hazelcast does not support backup redistribution on a data structure level. Thus, changing the backup count of a nonempty data structure is **not** supported. For example, for a map with a nonzero backup count:
 
-1. increasing the backup count does **not** create additional copies of existing map entries,
-2. decreasing the backup count does **not** remove any backups of existing map entries, so some backups will have stale data if the map entries are overwritten, and
+1. increasing the backup countfootnote:change-backup-count[The backup count of a map can only be changed by modifying the xref:configuration:understanding-configuration.adoc#static-configuration[static configuration] and restarting the cluster.] does **not** create additional copies of existing map entries,
+2. decreasing the backup countfootnote:change-backup-count[] does **not** remove any backups of existing map entries, so some backups will have stale data if the map entries are overwritten, and
 3. if some backups are lost due to node failure, additional copies are **not** created on other nodes to meet the backup count.
-
-Thus, it is unsupported to change the backup count of a data structure once it was populated with data.
 ====
 
 When a member in your cluster is lost, Hazelcast redistributes the backups


docs/modules/fault-tolerance/pages/backups.adoc
```diff
@@ -7,6 +7,17 @@ The distribution happens on the partition level; the data and its backups are st
...
 
+[IMPORTANT]
+====
+Hazelcast does not support backup redistribution on data structure level. For example, for a map with a nonzero backup count,
+
+1. increasing the backup count does **not** create additional copies of existing map entries,
+2. decreasing the backup count does **not** remove any backups of existing map entries, so some backups will have stale data if the map entries are overwritten, and
+3. if some backups are lost due to node failure, additional copies are **not** created on other nodes to meet the backup count.
+
+Thus, it is unsupported to change the backup count of a data structure once it was populated with data.
```
oliverhowell, line 18:
```suggestion
Hazelcast does not support backup redistribution on a data structure level. For example, for a map with a nonzero backup count:

1. increasing the backup count does **not** create additional copies of existing map entries,
2. decreasing the backup count does **not** remove any backups of existing map entries, so some backups will have stale data if the map entries are overwritten, and
3. if some backups are lost due to node failure, additional copies are **not** created on other nodes to meet the backup count.

Thus, changing the backup count of a data structure once it is populated with data is **not** supported.
```

oliverhowell (changes requested):
Thanks Burak, some minor suggestions

docs/modules/fault-tolerance/pages/backups.adoc
k-jamroz, line 18: >it is unsupported to change the backup count of a data structure once it was populated with data

this should be the first sentence of the note. It does not have to be a mathematical proof and this is the most important info for the user, the rest is just a more detailed explanation.

docs/modules/fault-tolerance/pages/backups.adoc
```diff
@@ -7,6 +7,17 @@ The distribution happens on the partition level; the data and its backups are st
 memory partitions. See the xref:overview:data-partitioning.adoc[Data Partitioning] and xref:clusters:partition-group-configuration.adoc[Partition Grouping]
 for more information about the partitioning.
 
+[IMPORTANT]
+====
+Hazelcast does not support backup redistribution on data structure level. For example, for a map with a nonzero backup count,
+
+1. increasing the backup count does **not** create additional copies of existing map entries,
```
k-jamroz, line 14: you do not distinguish static and dynamic config. Dynamic config cannot be changed at all

docs/modules/fault-tolerance/pages/backups.adoc
k-jamroz: also, does it make a difference if the structure is empty (eg. after `clear`), destroyed or never contained any data?

burakgok: ca51ef8c

burakgok: The limitation doesn't apply to nonempty data structures that once contained data. (ca51ef8c)

docs/modules/fault-tolerance/pages/backups.adoc
burakgok: I added a footnote to explain how the backup count can be changed. (ca51ef8c)

docs/modules/fault-tolerance/pages/backups.adoc
```diff
@@ -7,6 +7,15 @@ The distribution happens on the partition level; the data and its backups are st
 memory partitions. See the xref:overview:data-partitioning.adoc[Data Partitioning] and xref:clusters:partition-group-configuration.adoc[Partition Grouping]
 for more information about the partitioning.
 
+[IMPORTANT]
+====
+Hazelcast does not support backup redistribution on a data structure level. Thus, changing the backup count of a nonempty data structure is **not** supported. For example, for a map with a nonzero backup count:
+
+1. increasing the backup countfootnote:change-backup-count[The backup count of a map can only be changed by modifying the xref:configuration:understanding-configuration.adoc#static-configuration[static configuration] and restarting the cluster.] does **not** create additional copies of existing map entries,
+2. decreasing the backup countfootnote:change-backup-count[] does **not** remove any backups of existing map entries, so some backups will have stale data if the map entries are overwritten, and
```
k-jamroz, line 15: `countfootnote:change-backup-count[]`

the adoc syntax here is crazy, but surprisingly it works. Will it work also with extra space before `footnote:`?

docs/modules/fault-tolerance/pages/backups.adoc
```diff
@@ -7,6 +7,15 @@ The distribution happens on the partition level; the data and its backups are st
 memory partitions. See the xref:overview:data-partitioning.adoc[Data Partitioning] and xref:clusters:partition-group-configuration.adoc[Partition Grouping]
 for more information about the partitioning.
 
+[IMPORTANT]
+====
+Hazelcast does not support backup redistribution on a data structure level. Thus, changing the backup count of a nonempty data structure is **not** supported. For example, for a map with a nonzero backup count:
+
+1. increasing the backup countfootnote:change-backup-count[The backup count of a map can only be changed by modifying the xref:configuration:understanding-configuration.adoc#static-configuration[static configuration] and restarting the cluster.] does **not** create additional copies of existing map entries,
```
k-jamroz, line 14: the footnote says that backup count can only be changed. While I guess what is supposed to mean, it looks contradictory and puzzling.

```suggestion
1. increasing the backup countfootnote:change-backup-count[The backup count of an empty map can only be changed by modifying the xref:configuration:understanding-configuration.adoc#static-configuration[static configuration] and restarting the cluster.] does **not** create additional copies of existing map entries,
```

docs/modules/fault-tolerance/pages/backups.adoc
```diff
@@ -7,6 +7,15 @@ The distribution happens on the partition level; the data and its backups are st
...
+2. decreasing the backup countfootnote:change-backup-count[] does **not** remove any backups of existing map entries, so some backups will have stale data if the map entries are overwritten, and
+3. if some backups are lost due to node failure, additional copies are **not** created on other nodes to meet the backup count.
```
k-jamroz, line 16: This is more complicated.

Eg. if you have 3 member cluster and want to increase number of backups from 1 to 2, you might just do a rolling restart with new config (disclaimer: I did not test it). The trick is that actual number of replicas (owner+backups) is <= cluster size. So if you have too few members, the cluster does not try create replicas which do not fit. But when a new member joins, replicas with higher indexes can be crated.

Also, if you were able to force at some point in all partitions to be owned by members with updated config, then subsequent migrations should use new backup count. One way of doing that is as above but shutting down some members. Another way might be by demoting and promoting some members, but you would have to make sure that each partition is "touched".

OFC, none of these scenarios is a supported way of doing that.

k-jamroz: when I was thinking more of it, it should actually be possible to change number of backups in static config by doing 2 rolling restarts: first for updating the config in members, second for (indirectly) triggering migrations. This is wasteful, because you need to send all data back and forth many times and the restart should be done ACTIVE cluster state, but should work.

this might not work with hotrestart if the static config is saved also in hot restart data (I did not check), and will be not possible with dynamic config (with or without hotrestart) because dynamic config is immutable

burakgok: Partition rebalancing might correct the number of backups as reported [here][1]. However, I am not sure whether we should accept this as a workaround and promote it in our documentation.

[1]: https://hazelcast.atlassian.net/browse/SUP-416?focusedCommentId=102524

docs/modules/fault-tolerance/pages/backups.adoc
burakgok: Unfortunately, the extra space becomes part of the document, and [the workaround][1] suggested in the AsciiDoc documentation is ugly.

[1]: https://docs.asciidoctor.org/asciidoc/latest/macros/footnote#:~:text=If%20you%20find,%29.

docs/modules/fault-tolerance/pages/backups.adoc
k-jamroz: I am not sure if this procedure works by design or by accident. If we make sure that it really works, it might be provided by support if needed (and if is still works). But I would not add it to the docs as it does not seem to be tested automatically and could change in the future.

docs/modules/fault-tolerance/pages/backups.adoc
k-jamroz: meh

docs/modules/fault-tolerance/pages/backups.adoc
burakgok: f1b97d8d

k-jamroz (approved):

oliverhowell (approved):
Thanks @burakgok adding backport label and approving

PR #1315, opened 2024-10-07
Correct PassThroughMergePolicy documentation [DOC-174]

gareth-johnston:
Per the code and javadoc this should read ".. unless the incoming entry is null." 

Javadoc :

`Merges data structure entries from source to destination directly unless the merging entry is {@code null}`


JamesHazelcast (approved):
Thanks for cleaning this up @gareth-johnston!

oliverhowell (approved):

PR #1316, opened 2024-10-07
Use latest `setup-node` in GitHub Actions

JackPGreen:
Rather than specifying a specific `setup-node` version, and having to [manually increment it](https://github.com/hazelcast/hz-docs/pulls?q=actions%2Fsetup-node), instead we can specify the latest `v4` as we do with other GitHub Actions (e.g. `checkout`).


oliverhowell (approved):
Thanks @JackPGreen (will close the dependabot PR)

PR #1321, opened 2024-10-14
Update the javadoc link for the new client

ihsandemir:
Fıxed the javadoc link for the new client.

### DIFF
--- a/docs/modules/clients/pages/java.adoc
+++ b/docs/modules/clients/pages/java.adoc
@@ -19,7 +19,7 @@ NOTE: Where there are specific differences between {java-client} and {java-clien
 Both platform ({java-client}) and standalone ({java-client-new}) clients enable you to use the Hazelcast API, with this page explaining any differences or technical details that affect usage. This page should be read alongside the respective Javadoc-generated API documentation available from within your IDE and the following links:
 
 * https://docs.hazelcast.org/docs/{page-latest-supported-java-client}/javadoc[Hazelcast {java-client} API documentation]
-* https://docs.hazelcast.org/hazelcast-java-client-docs/{page-latest-supported-java-client-new}/apidocs[Hazelcast {java-client-new} API documentation]
+* https://docs.hazelcast.org/hazelcast-java-client/{page-latest-supported-java-client-new}/javadoc[Hazelcast {java-client-new} API documentation]
 
 == Get started
 
--- a/docs/modules/clients/pages/java.adoc
+++ b/docs/modules/clients/pages/java.adoc
@@ -16,7 +16,7 @@ NOTE: Where there are specific differences between {java-client} and {java-clien
 
 // check production recommendation
 
-Both platform ({java-client}) and standalone ({java-client-new}) clients enable you to use the Hazelcast API, with this page explaining any differences or technical details that affect usage. This page should be read alongside the respective Javadoc-generated API documentation available from within your IDE and the following links:
+Both clients enable you to use the Hazelcast API, with this page explaining any differences or technical details that affect usage. This page should be read alongside the respective Javadoc-generated API documentation available from within your IDE and the following links:
 
 * https://docs.hazelcast.org/docs/{page-latest-supported-java-client}/javadoc[Hazelcast {java-client} API documentation]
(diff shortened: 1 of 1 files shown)


docs/modules/clients/pages/java.adoc
```diff
@@ -1,5 +1,6 @@
 = Java Client
 :page-api-reference: https://docs.hazelcast.org/docs/{page-latest-supported-java-client}/javadoc
+:page-api-reference: https://docs.hazelcast.org/hazelcast-java-client-docs/{page-latest-supported-java-client-new}/apidocs
```
oliverhowell, line 3:
```suggestion

```
I don't think we want to add this. The page-api-reference is only available on 1-1 basis with page so this overwrites the API docs link at top of page from embedded to standalone. We can't have two links here so the choice is none (we take this variable out and rely on hardcoded lnks) or leave it as embedded version for now (until standalone out of beta and/or other page restructuring). WDYT?

docs/modules/clients/pages/java.adoc
```diff
@@ -16,10 +17,10 @@ NOTE: Where there are specific differences between {java-client} and {java-clien
 
 // check production recommendation
 
-Both native ({java-client}) and standalone ({java-client-new}) clients enable you to use the Hazelcast API, with this page explaining any differences or technical details that affect usage. This page should be read alongside the respective Javadoc-generated API documentation available from within your IDE and the following links:
+Both platform ({java-client}) and standalone ({java-client-new}) clients enable you to use the Hazelcast API, with this page explaining any differences or technical details that affect usage. This page should be read alongside the respective Javadoc-generated API documentation available from within your IDE and the following links:
```
oliverhowell, line 20: Is this change because we don't want to say 'native'?

docs/modules/clients/pages/java.adoc
```diff
@@ -16,10 +17,10 @@ NOTE: Where there are specific differences between {java-client} and {java-clien
...
 
 * https://docs.hazelcast.org/docs/{page-latest-supported-java-client}/javadoc[Hazelcast {java-client} API documentation]
-* https://docs.hazelcast.org/hazelcast-java-client/{page-latest-supported-java-client-new}/javadoc[Hazelcast {java-client-new} API documentation]
+* https://docs.hazelcast.org/hazelcast-java-client-docs/{page-latest-supported-java-client-new}/apidocs[Hazelcast {java-client-new} API documentation]
```
oliverhowell, line 23: Link works ok but not sure if we still have control over the generated url as I think the '-docs' is redundant and we're using /javadoc in 1st case and /apidocs in second. Ideally we should be consistent but no major objection if this is what we are sticking with.

oliverhowell (changes requested):

docs/modules/clients/pages/java.adoc
ihsandemir: OK, changed as suggested the folder names at the docs server.

srknzl (approved):

docs/modules/clients/pages/java.adoc
oliverhowell: Thanks, I think that's better

docs/modules/clients/pages/java.adoc
```diff
@@ -16,7 +16,7 @@ NOTE: Where there are specific differences between {java-client} and {java-clien
 
 // check production recommendation
 
-Both native ({java-client}) and standalone ({java-client-new}) clients enable you to use the Hazelcast API, with this page explaining any differences or technical details that affect usage. This page should be read alongside the respective Javadoc-generated API documentation available from within your IDE and the following links:
+Both platform ({java-client}) and standalone ({java-client-new}) clients enable you to use the Hazelcast API, with this page explaining any differences or technical details that affect usage. This page should be read alongside the respective Javadoc-generated API documentation available from within your IDE and the following links:
```
oliverhowell, line 19:
```suggestion
Both clients enable you to use the Hazelcast API, with this page explaining any differences or technical details that affect usage. This page should be read alongside the respective Javadoc-generated API documentation available from within your IDE and the following links:
```

I would just simplify to above ^

oliverhowell (approved):
Thanks @ihsandemir

PR #1339, opened 2024-10-21
Small CDC connector fixes

TomaszGaweda:
- use hazelcast-enterprise instead of hazelcast oss
- java 17 used


frant-hartm (changes requested):
The artifact `hazelcast-enterprise-cdc-debezium` is in our private repository. It should be mentioned and it should be present in the pom/gradle script. Otherwise the build fails.

TomaszGaweda:
@frant-hartm sorry I've forgot that. I've added to all existing pom.xml / build.gradle + added in the overall setup guide.

TomaszGaweda:
This should be resolved @oliverhowell 
image

oliverhowell:
Thanks @TomaszGaweda I can see the build errors are gone from log too

docs/modules/integrate/pages/cdc-connectors.adoc
```diff
@@ -58,6 +58,47 @@ PostgreSQL-specific connector:
 </dependency>
 ----
 
+Since all those artifacts are in Hazelcast Private Repository, you also have to add:
```
oliverhowell, line 61: 'Since' can confuse some non-native speakers so suggesting:
```suggestion
Because all these artifacts are in Hazelcast's private repository, you also have to add:
```

oliverhowell (approved):

PR #1343, opened 2024-10-22
DOC-191: Add Ask AI page

oliverhowell:
Adds Ask AI page to nav so it features prominently and also provides some detail and tips.

Backport strategy TBD but we can replicate this page in each of the other products too (initially the same but can always diverge in future and e.g. make tips more MC appropriate).

- [ ] Decide what versions we backport to


amandalindsay (changes requested):
tiny tweak, otherwise looks good!

oliverhowell:
Backporting: let's see if the backport to 5.5 works ok and then add a separate release notes entry to 5.5 branch

amandalindsay (approved):

oliverhowell:
I replaced the gif with a couple of images - on reflection I think it's not worth the extra file size in this context.

PR #1364, opened 2024-11-05
Add AccessControlException deprecation note

JamesHazelcast:
This exception will need to be replaced in the future, but a replacement is not yet planned, so we should update the documentation to acknowledge its deprecation at least.

Fixes https://hazelcast.atlassian.net/browse/HZG-245


amandalindsay (approved):

JamesHazelcast:
@amandalindsay could we add this to list of PRs to backport for 5.5 please?

PR #1366, opened 2024-11-07
Add vector collection backup documentation, update after upgrade to JVector 3 [AI-155] [AI-230]

k-jamroz:
In 6.0:

1. fault tolerance is supported
2. memory leak is no longer present
3. new default `max-degree`

Also fixed which config settings are optional.


docs/modules/data-structures/pages/vector-collections.adoc
```diff
@@ -75,19 +86,19 @@ For further information on distance metrics, see the <<available-metrics, Availa
 |`N/A`
 
 |max-degree
-|Used to calculate the maximum number of neighbors per node. The calculation used is max-degree * 2
-|Required
-|`16`
+|Maximum number of neighbors per node. Note that meaning of this parameter is different from 5.5 version.
```
amandalindsay, line 89:
```suggestion
|Maximum number of neighbors per node. Note that the meaning of this parameter differs from version 5.5.
```

amandalindsay (changes requested):
Looks good, suggest minor tweak only

docs/modules/data-structures/pages/vector-collections.adoc
```diff
@@ -75,19 +86,19 @@ For further information on distance metrics, see the <<available-metrics, Availa
 |`N/A`
 
 |max-degree
-|Used to calculate the maximum number of neighbors per node. The calculation used is max-degree * 2
-|Required
-|`16`
+|Maximum number of neighbors per node. Note that the meaning of this parameter differs from version 5.5.
+|Optional
```
shultseva, line 90: Is it accurate to write that this parameter is optional, given that it cannot be set to null?

k-jamroz: it has a default value, so it does not have to be configured explicitly. It cannot be set explicitly to null in Java, it should be similar in XML and YAML but I did not test what would happen.

k-jamroz: I define "optional" as "nullable or has reasonable default" (similar to https://docs.hazelcast.com/hazelcast/5.5/migrate/dmt-command-reference). If we want to inform precisely about nullability, it should be separate column. But IMO for the users the most import information is if they have to pass it or not.

shultseva (approved):

docs/modules/data-structures/pages/vector-search-overview.adoc
```diff
@@ -43,17 +43,26 @@ The index is based on the link:https://github.com/jbellis/jvector[JVector] libra
 
 Each collection is partitioned and replicated based on the system's general partitioning rules. Data partitioning is carried out using the collection key.
 
-For further information on Hazelcast partitioning, see xref:architecture:data-partitioning.adoc[Data Partitioning and Replication].
+Vector collection is AP data structure and implements standard xref:data-structures:backing-up-maps.adoc#in-memory-backup-types[in-memory backup types].
```
amandalindsay, line 46: Will users know what AP means or should we spell it out?

docs/modules/data-structures/pages/vector-search-overview.adoc
```diff
@@ -43,17 +43,26 @@ The index is based on the link:https://github.com/jbellis/jvector[JVector] libra
...
 Hazelcast distributes storage data across all cluster members.
 In the event of a graceful shutdown, the data is migrated to remaining active members.
-In version 5.5, there is no automatic data restoration in the event of an unexpected member loss.
+In the event of member crash, if backups were configured, they are used to restore the data.
+If no backups were configured, data can be lost.
+
+NOTE: You can disable backups to speed up data ingestion into Vector Collection and reduce memory usage
+in development and test environments which do not require fault tolerance.
+You do not have to do it in case of single-member cluster (for example embedded dev cluster) because in such case
+backups are not used.
```
amandalindsay, line 65:
```suggestion
With a single-member cluster (for example, an embedded dev cluster), you do not need to disable backups because these are not used in this case.
```

docs/modules/data-structures/pages/vector-collections.adoc
```diff
@@ -75,19 +86,19 @@ For further information on distance metrics, see the <<available-metrics, Availa
 |`N/A`
 
 |max-degree
-|Used to calculate the maximum number of neighbors per node. The calculation used is max-degree * 2
-|Required
-|`16`
+|Maximum number of neighbors per node. Note that the meaning of this parameter differs from version 5.5.
```
amandalindsay, line 89:
```suggestion
|Maximum number of neighbors per node. Note that the meaning of this parameter differs from that used in version 5.5.
```

amandalindsay (changes requested):
Looks good, just made some minor tweaks

docs/modules/data-structures/pages/vector-search-overview.adoc
k-jamroz: added link: https://github.com/hazelcast/hz-docs/pull/1366/commits/18692302ac5c19523788d15679a6fbdf3ae05bbb

PR #1371, opened 2024-11-11
Update backing-up-maps.adoc

amandalindsay:
removed "or to persistent memory" (and the 'either') as per the docs ticket


Rob-Hazelcast (approved):
I've not look at the ticket, but the docs change is fine

PR #1355, opened 2024-10-31
No snapshot releases for OSS

TomaszGaweda:
- Introduced property to set last OSS version (it can be different than EE version)
- removed all snapshot version-related parts from OSS install guide

PR #1377, opened 2024-11-18
Update map-config.adoc

amandalindsay:
Changed broken link to point to master to avoid a 404

### DIFF
--- a/docs/modules/data-structures/pages/map-config.adoc
+++ b/docs/modules/data-structures/pages/map-config.adoc
@@ -23,8 +23,8 @@ ifdef::snapshot[]
 endif::[]
 
 ifndef::snapshot[]
-* link:https://github.com/hazelcast/hazelcast/blob/v{full-version}/hazelcast/src/main/resources/hazelcast-full-example.xml[hazelcast-full-example.xml]
-* link:https://github.com/hazelcast/hazelcast/blob/v{full-version}/hazelcast/src/main/resources/hazelcast-full-example.yaml[hazelcast-full-example.yaml]
+* link:https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/resources/hazelcast-full-example.xml[hazelcast-full-example.xml]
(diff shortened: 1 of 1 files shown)


JackPGreen:
[Slack discussion](https://hazelcast.slack.com/archives/C035HQET5/p1731921884313829).

docs/modules/data-structures/pages/map-config.adoc
```diff
@@ -23,8 +23,8 @@ ifdef::snapshot[]
 endif::[]
 
 ifndef::snapshot[]
-* link:https://github.com/hazelcast/hazelcast/blob/v{full-version}/hazelcast/src/main/resources/hazelcast-full-example.xml[hazelcast-full-example.xml]
-* link:https://github.com/hazelcast/hazelcast/blob/v{full-version}/hazelcast/src/main/resources/hazelcast-full-example.yaml[hazelcast-full-example.yaml]
+* link:https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/resources/hazelcast-full-example.xml[hazelcast-full-example.xml]
```
fantkolja, line 26: Sorry for interfering, but maybe a better approach would be to introduce a new variable e.g. `hazelcast-tag`, and use it here?
It seems that we have here a wrong usage of `full-version` variable. That variable means the version of the documentation but it was used here ad the version of the hazelcast (set in GH tag)

image

JackPGreen: > Sorry for interfering, but maybe a better approach would be to introduce a new variable e.g. `hazelcast-tag`, and use it here? It seems that we have here a wrong usage of `full-version` variable. That variable means the version of the documentation but it was used here ad the version of the hazelcast (set in GH tag)

@fantkolja already under discussion [here](https://github.com/hazelcast/hz-docs/pull/1355).

oliverhowell: And related PR: https://github.com/hazelcast/hz-docs/pull/1355

PR #1382, opened 2024-11-18
Introduce GitHub Action to test resolvability of external links [DOC-253]

JackPGreen:
This PR adds an action that will, when a PR is opened, check the resolvability of any external links.

External links are defined as those using the `http` scheme - internal links _between_ documentation is already covered by our existing dead links check.

Action is deliberately modular to be be easily applied to other docs repos.

This runs whenever a PR is raised, which is sub-optimal because:
- links can break at any time as they reference external content
- there's no guarantees the breakage is anything to do with the new PR

But the other alternative - a scheduled check - would have difficulty in knowing _who_ to target a failure communication at to resolve the issues (other projects notify Slack channels, with varying degrees of responses).

An example of the output produced by this action can be found [here](https://github.com/hazelcast/hz-docs/actions/runs/11901935786).

Note that the external link check *fails* because of dead links, and won't pass until the following are addressed:
- https://github.com/hazelcast/hz-docs/pull/1355
- https://github.com/hazelcast/hz-docs/pull/1389
- `URL 'https://raw.github.com/olivernn/lunr.js/master/lunr.min.js' had status 404 (found in node_modules/lunr/index.html)`

We should consider whether adding a check that we know will fail is a good idea - comparing the annoyance of a failing (non-blocking) test against the coverage that at least we can ensure things don't get _worse_.

Fixes: [DOC-253](https://hazelcast.atlassian.net/browse/DOC-253)

[DOC-253]: https://hazelcast.atlassian.net/browse/DOC-253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fantkolja (commented):
Very nice!
The only thing - probably some URLs aren't properly parsed, so there are a lot of `bad/illegal format or missing URL` logs

image

JackPGreen:
> Very nice! The only thing - probably some URLs aren't properly parsed, so there are a lot of `bad/illegal format or missing URL` logs

Fixed with workaround - https://github.com/hazelcast/hz-docs/pull/1382/commits/7a781087b3e8f92b7bb55734c9b0bfded2b14fe7

Updated output - https://github.com/hazelcast/hz-docs/actions/runs/11907767488

fantkolja (approved):

JackPGreen:
Suggestion was to notify put on a cron schedule and have it notify a private `docs` slack channel.

oliverhowell (approved):

PR #1398, opened 2024-11-25
Reusable action for docs deployment

fantkolja:


.github/actions/deploy/action.yml
JackPGreen, line 1: I think this should be a re-usable action (`workflow_call`) as it's not a step you'd want to use in conjunction with anything else. Plus that would mean you could centralise the `runs-on` definition.

I also don't _think_ you can use relative paths with composite actions without a `checkout` first. But I could be wrong.

.github/workflows/deploy-docs-workflow.yml
```diff
@@ -0,0 +1,12 @@
...
+
+on:
+  workflow_call:
+
+jobs:
+  deploy:
+    runs-on: ubicloud-standard-2
+    steps:
+      - name: Dispatch docs deployment
+        run: curl -X POST -d {} https://api.netlify.com/build_hooks/60e71936b15bb431b42b698a
```
JackPGreen, line 12: Nit - a link to the [netlify documentation that specifies this command](https://docs.netlify.com/configure-builds/build-hooks/) might help to avoid someone wondering what it does.

JackPGreen (approved):

PR #1401, opened 2024-11-25
DOC-279: add blog link and description to page

oliverhowell:


Rob-Hazelcast (approved):
Text looks good.

The padding around the first image isn't ideal, but I see you've got another PR for that.

I think the call-out box is appropriate, it's just a little unfortunate that it's labelled "Tip" under the heading "Tips". Might be nice to change that heading. Not a big deal though.

oliverhowell:
> I think the call-out box is appropriate, it's just a little unfortunate that it's labelled "Tip" under the heading "Tips". Might be nice to change that heading. Not a big deal though.

Yeah, maybe just a note would be slightly better? I didn't want to introduce new nomenclature but agree Tip not ideal...

PR #1403, opened 2024-11-26
Optimize backport.yml

zhmaeff:
Use ENV var instead of empty job with output.


JackPGreen (approved):
This is nicer, thanks!

Could you do the same in [`hazelcast-platform-operator` as well, please](https://github.com/hazelcast/hazelcast-platform-operator-docs/blob/main/.github/workflows/backport.yml)?

fantkolja (approved):

PR #1417, opened 2024-12-05
Data connection

oliverhowell:
Quick revision to add to nav and correct some titles...

Not sure we have Data Connections in the right place - WDYT @TomaszGaweda ?


Rob-Hazelcast (approved):

docs/modules/data-connections/partials/nav.adoc
```diff
@@ -1,2 +1 @@
-* Data Connections
-** xref:data-connections:data-connections-configuration.adoc[]
+
```
TomaszGaweda, line 1: can we completely delete it now?

docs/modules/ROOT/nav.adoc
```diff
@@ -157,7 +157,11 @@ include::cp-subsystem:partial$nav.adoc[]
 * xref:storage:high-density-memory.adoc[]
 include::tiered-storage:partial$nav.adoc[]
 * xref:cluster-performance:thread-per-core-tpc.adoc[]
-include::data-connections:partial$nav.adoc[]
+* Data Connections
```
TomaszGaweda, line 160: As discussed on Slack, would be best to put in `Develop and Build`

TomaszGaweda (changes requested):

docs/modules/data-connections/partials/nav.adoc
oliverhowell: Yeah, we might as well

docs/modules/ROOT/nav.adoc
oliverhowell: What do you think of it under the pipelines section as per the latest commit?

TomaszGaweda: Perfect 👌

TomaszGaweda (approved):

oliverhowell:
Should we backport this to 5.5 and other versions? @TomaszGaweda

TomaszGaweda:
Definitely for 5.5 as it's our LTS version. 5.3 and 5.4 is optional IMHO

PR #1459, opened 2025-01-07
Add note about classpath for custom EPs in pipelines [HZG-287]

JamesHazelcast:
With the recently added support for UCN in Jet (available in next minor release), it is useful to mention this in documentation, particularly for the commonly used `EntryProcessor`. Even without UCN, the custom classes need to be on the classpath, so it's good to be explicit about this.

Only relevant for next minor release (5.6.0?)

Fixes https://hazelcast.atlassian.net/browse/HZG-287

### DIFF
--- a/docs/modules/integrate/pages/map-connector.adoc
+++ b/docs/modules/integrate/pages/map-connector.adoc
@@ -191,7 +191,7 @@ from the map.
 NOTE: Custom classes, such as an `EntryProcessor`, must be available to all members
 of the cluster. This can be done by adding the classes to each member's classpath
 directly, or by using {ucn}. If using {ucn}, ensure that the Job and underlying IMap
-are both associated with the correct namespace.
+are both associated with a namespace that contains the same `EntryProcessor` implementation.
 
 == Predicates and Projections
 


oliverhowell (approved):

docs/modules/integrate/pages/map-connector.adoc
```diff
@@ -188,6 +188,11 @@ setting their values to `null`. To put it another way, if these map sink
 variants set the entry’s value to null, the entry will be removed
 from the map.
 
+NOTE: Custom classes, such as an `EntryProcessor`, must be available to all members
+of the cluster. This can be done by adding the classes to each member's classpath
+directly, or by using {ucn}. If using {ucn}, ensure that the Job and underlying IMap
+are both associated with the correct namespace.
```
k-jamroz, line 194: should we clarify what the correct namespace is? it does not have to be the same namespace, but both namespaces (job and IMap) have to contain the entry processor, ideally with the same serialized format

JamesHazelcast: It feels obvious to me, but I know it's not always obvious to our users 😅 I think to avoid potential issues, we should state "are both associated with the same namespace" - you're right that they don't need to be the same, but if they deployed different versions across different namespaces, it could cause issues... so keep it safe for the laymen we say use the same namespace, and experts will know the difference themselves (and so could use different namespaces if they want)? wdyt?

JamesHazelcast: Sounds good, thanks @k-jamroz 👍

k-jamroz (approved):

JamesHazelcast:
Another one for the backlog (sorry for the spam) @amandalindsay!

PR #1461, opened 2025-01-07
Clarify MapLoader interaction behaviour [HZG-289]

JamesHazelcast:
Adds additional clarity about `#get` and `#getAll` behaviour on `IMap` when a `MapLoader` is provided. Since the `MapStore` is an extension of the `MapLoader`, and some behaviour is already documented in the `mapstore-triggers` page, I've added it to the existing table.

I've also added a line to the bottom of the page directing users to the javadocs for more detailed information about interactions as this is well documented in javadocs.

Fixes https://hazelcast.atlassian.net/browse/HZG-289

### DIFF
--- a/docs/modules/mapstore/pages/mapstore-triggers.adoc
+++ b/docs/modules/mapstore/pages/mapstore-triggers.adoc
@@ -14,6 +14,9 @@ is called, it triggers the `MapLoader.loadAllKeys()` method.
 |Flush all the local dirty entries by calling the `MapStore.storeAll(Map)` or `MapStore.deleteAll(Collection)` methods with the elements marked as dirty.
 
 |get()
+
+getAsync()
+
 |Retrieve an entry from the map. If the entry does not exist in memory the `MapLoader.load(Object)` method is called to retrieve
 the entry from the external system if it exists.
 
--- a/docs/modules/mapstore/pages/mapstore-triggers.adoc
+++ b/docs/modules/mapstore/pages/mapstore-triggers.adoc
@@ -13,6 +13,13 @@ is called, it triggers the `MapLoader.loadAllKeys()` method.
 |flush()
 |Flush all the local dirty entries by calling the `MapStore.storeAll(Map)` or `MapStore.deleteAll(Collection)` methods with the elements marked as dirty.
 
+|get()
+|Retrieve an entry from the map. If the entry does not exist in memory the `MapLoader.load(Object)` method is called to retrieve
+the entry from the external system if it exists.
+
+|getAll()
+|Fetches all entries from the map that match the provided key set. If any keys are not found in memory then this method calls the `MapLoader.loadAll(Collection)` method with the missing keys.
+
 |put()
 
 putAll()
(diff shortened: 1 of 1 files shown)


docs/modules/mapstore/pages/mapstore-triggers.adoc
```diff
@@ -66,3 +73,6 @@ executeOnAllEntries()
 
 |Apply the user-defined entry processors to one or more entries. These methods call the `MapLoader.load(Object)` method if the value with key is not found in the memory, to load the value from the external system. If the entry processor updates the entry, these methods call the `MapStore.store(Object, Object)` method to write the value into the external system. If the entry processor updates the entry's value to null value, they call the `MapStore.delete(Object)` method to delete the value from the external system.
 |===
+
+More information about the behaviour of IMap method calls and their relationship to `MapStore` methods can be found in the
```
amandalindsay, line 77:
```suggestion
More information about the behavior of IMap method calls and their relationship to `MapStore` methods can be found in the
```

amandalindsay: Annoyingly, we use US spelling :-)

JamesHazelcast: Good catch, thanks!

wyohm:
Thanks @JamesHazelcast 
> Since the MapLoader is an extension of the MapStore
[MapStore](https://docs.hazelcast.org/docs/5.5.2/javadoc/com/hazelcast/map/MapStore.html) extends [MapLoader](https://docs.hazelcast.org/docs/5.5.2/javadoc/com/hazelcast/map/MapLoader.html). Does that change your answer any?

Also, methods like [getAsync()](https://docs.hazelcast.org/docs/5.5.2/javadoc/com/hazelcast/map/IMap.html#getAsync(K)) should probably be added. And others like `values()` and `keySet()` would benefit from either being included or explicitly stating that they do not trigger anything. Would it make sense to have a table of all IMap methods? (possibly with a checkbox column for "triggers MapLoader", "triggers MapStore", and description)

JamesHazelcast:
> [MapStore](https://docs.hazelcast.org/docs/5.5.2/javadoc/com/hazelcast/map/MapStore.html) extends [MapLoader](https://docs.hazelcast.org/docs/5.5.2/javadoc/com/hazelcast/map/MapLoader.html). Does that change your answer any?

This was a typo, thanks for catching - it was meant to be the reverse, but does not change the content of the PR thankfully.

> Also, methods like [getAsync()](https://docs.hazelcast.org/docs/5.5.2/javadoc/com/hazelcast/map/IMap.html#getAsync(K)) should probably be added. And others like values() and keySet() would benefit from either being included or explicitly stating that they do not trigger anything. Would it make sense to have a table of all IMap methods? (possibly with a checkbox column for "triggers MapLoader", "triggers MapStore", and description)

I will add `getAsync` as an alias for `get` in this table for clarity. I don't think we should include methods that do not trigger the `MapStore` or `MapLoader` as the page title is "Map Operations that Trigger Methods on a MapStore" and it would likely become information overload to list all the IMap methods (most of which won't have value here).

I did add a note to the bottom of the page which directs people to the javadocs - they contain detailed information about the interaction of each method with `MapLoader` (there's a specific subsection for each method). I don't think there's much value in copying the same information into this page as to me it serves as a quick reference (and adding tonnes of information makes it less navigable).

@wyohm

wyohm:
@JamesHazelcast I certainly agree with not duplicating information.

Is a lack of a "Interactions with the map store" section in the IMap method docs a guarantee that there is no interaction? (as is the case for `values()`, `keySet()`)

JamesHazelcast:
> Is a lack of a "Interactions with the map store" section in the IMap method docs a guarantee that there is no interaction? (as is the case for `values()`, `keySet()`)

Looking at code paths from `values()` and `keySet()`, they're transformed into predicates and I don't see any reference to the `MapLoader` in the predicate chain, so it seems that is the case, yes @wyohm

amandalindsay (approved):

wyohm (approved):

JamesHazelcast:
Another backport for the list please @amandalindsay

PR #1485, opened 2025-01-17
Update versioning for Windows support, and clean up table structures.

avtarraikmo:


docs/modules/deploy/pages/versioning-compatibility.adoc
```diff
@@ -105,13 +110,13 @@ Hazelcast Platform has been tested against the following operating systems.
 |===
 |Environment | Platform {full-version}
 
-|AWS
+|Amazon AWS
```
JackPGreen, line 113:
```suggestion
|Amazon Web Services (AWS)
```

Nit (it's the "PIN Number" argument)

docs/modules/deploy/pages/versioning-compatibility.adoc
```diff
@@ -105,13 +110,13 @@ Hazelcast Platform has been tested against the following operating systems.
...
 |✓
 
-|GCP
+|Google GCP
```
JackPGreen, line 116:
```suggestion
|Google Cloud Platform (GCP)
```

Nit

docs/modules/deploy/pages/versioning-compatibility.adoc
```diff
@@ -78,11 +72,22 @@ Hazelcast Platform has been tested against the following operating systems.
 |Ubuntu
 |✓
 
+|Windows Server 2019
+|✓
+
+|Windows 10, 11footnote:dev[]
```
oliverhowell, line 78: If you take line 63 out the footnotes don't resolve (see the preview) so I think putting this back in should fix:

```suggestion
|Windows 10, 11footnote:dev[Suitable for the development of applications (or Hazelcast itself), _but not_ for production. Some features are not available or are not supported.]
```

oliverhowell (changes requested):
Thanks @avtarraikmo Suggested change on the footnotes - also a question around Docker versions as we're currently listing 18 and 19 - should we update this too?

avtarraikmo:
> Thanks @avtarraikmo Suggested change on the footnotes - also a question around Docker versions as we're currently listing 18 and 19 - should we update this too?

@fawazghali / @oliverhowell - can you check which Docker version we are validating against? If so - we can update that too.

oliverhowell (approved):
Approving in case you want to merge while we wait on Docker confirmation

PR #1498, opened 2025-01-23
Document expensive IMap reporting threshold & property

JamesHazelcast:
Adds documentation about the expensive IMap client invocations logging, including the property used to control this threshold and guidance on which methods utilize this reporting.

Related PR: https://github.com/hazelcast/hazelcast-mono/pull/3856

### DIFF
--- a/docs/modules/cluster-performance/pages/imap-bulk-read-operations.adoc
+++ b/docs/modules/cluster-performance/pages/imap-bulk-read-operations.adoc
@@ -19,6 +19,13 @@ Methods like `IMap#entrySet()` and `IMap#values()` can trigger an OOME, dependin
 on the size of your map and the available memory on each member.
 To mitigate this risk, you should follow these best practices.
 
+NOTE: Client invocations of certain `IMap` methods such as those described above
+are logged on the member side when their results meet or exceed a threshold defined
+by the Hazelcast property `hazelcast.expensive.imap.invocation.reporting.threshold`,
+which has a default value of `100` results.
+See the https://docs.hazelcast.org/docs/{full-version}/javadoc/com/hazelcast/spi/properties/ClusterProperty.html#EXPENSIVE_IMAP_INVOCATION_REPORTING_THRESHOLD[relevant Javadocs^]
+for details on which `IMap` methods are logged.
+
 == Plan capacity
 Proper capacity planning is crucial for providing
 sufficient system resources to the Hazelcast cluster. This
--- a/docs/modules/cluster-performance/pages/imap-bulk-read-operations.adoc
+++ b/docs/modules/cluster-performance/pages/imap-bulk-read-operations.adoc
@@ -19,12 +19,11 @@ Methods like `IMap#entrySet()` and `IMap#values()` can trigger an OOME, dependin
 on the size of your map and the available memory on each member.
 To mitigate this risk, you should follow these best practices.
 
-NOTE: Client invocations of certain `IMap` methods such as those described above
-are logged on the member side when their results meet or exceed a threshold defined
+NOTE: To help you to monitor this potential problem, client invocations of certain `IMap` methods are logged on the member side when their results meet or exceed a threshold.
+The threshold for logging large results is defined
 by the Hazelcast property `hazelcast.expensive.imap.invocation.reporting.threshold`,
 which has a default value of `100` results.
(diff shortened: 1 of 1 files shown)


docs/modules/cluster-performance/pages/imap-bulk-read-operations.adoc
```diff
@@ -19,6 +19,13 @@ Methods like `IMap#entrySet()` and `IMap#values()` can trigger an OOME, dependin
 on the size of your map and the available memory on each member.
 To mitigate this risk, you should follow these best practices.
 
+NOTE: Client invocations of certain `IMap` methods such as those described above
+are logged on the member side when their results meet or exceed a threshold defined
+by the Hazelcast property `hazelcast.expensive.imap.invocation.reporting.threshold`,
+which has a default value of `100` results.
+See the https://docs.hazelcast.org/docs/{full-version}/javadoc/com/hazelcast/spi/properties/ClusterProperty.html#EXPENSIVE_IMAP_INVOCATION_REPORTING_THRESHOLD[relevant Javadocs^]
+for details on which `IMap` methods are logged.
```
JackPGreen, line 27:
```suggestion
To identify this potential problematic usage, client invocations of `IMap` methods such as those described above returning large results are logged on the member side.

The threshold for logging large results is defined
by the Hazelcast property `hazelcast.expensive.imap.invocation.reporting.threshold`,
which has a default value of `100` results.
See the https://docs.hazelcast.org/docs/{full-version}/javadoc/com/hazelcast/spi/properties/ClusterProperty.html#EXPENSIVE_IMAP_INVOCATION_REPORTING_THRESHOLD[relevant Javadocs^]
for details on which `IMap` methods are logged.
```


Nit - I feel like this could be split into two sections to be clearer. But I'm not sure mine is an improvement!

JackPGreen (approved):

docs/modules/cluster-performance/pages/imap-bulk-read-operations.adoc
Rob-Hazelcast: Agree, but that opening sentence is a little hard to parse, and we need to be clear that it's a defined set of methods _and_ a defined threshold. How about this:

```suggestion
NOTE: To help you to monitor this potential problem, client invocations of certain `IMap` methods are logged on the member side.
The threshold for logging large results is defined
by the Hazelcast property `hazelcast.expensive.imap.invocation.reporting.threshold`,
which has a default value of `100` results.
The applicable methods are listed in the https://docs.hazelcast.org/docs/{full-version}/javadoc/com/hazelcast/spi/properties/ClusterProperty.html#EXPENSIVE_IMAP_INVOCATION_REPORTING_THRESHOLD[Javadocs^].
```

I could go either way on using a note box, but we need it here because this doesn't follow on from "you should follow these best practices".

Should I be able to see that field at that link?

Anything we can say / link to about what to do with the logs?

JamesHazelcast: Updated to match your suggestion, thanks @Rob-Hazelcast - that field will not be present in the javadocs until the update is released (new field), but this docs change should not be published until then either. I've asked Olly in the past about whether we should merge changes that aren't live and the conclusion seemed to be yes, so I assume there is some internal cherry picking of what goes live and what doesn't (if not, please let me know).

> Anything we can say / link to about what to do with the logs?
The logs are informational, just to flag potential unexpected expensive calls from clients. I'm not sure what specific guidance we would give.

Rob-Hazelcast (approved):

docs/modules/cluster-performance/pages/imap-bulk-read-operations.adoc
Rob-Hazelcast: I figured it'd be something like that, thanks.

Merging ultimately means publishing. It's fine to merge to main whenever - that does get published, but it's marked as pre-release - but you shouldn't merge to any earlier branches until a feature or change is available to customers (assuming it's available on earlier versions, which new features typically wouldn't be).

So:

- if this is for the next release - 5.6, I think? - merge to main now.

- if this is part of a 5.5.X maintenance release, it would need to be merged to main and backported to 5.5, and you shouldn't do that until the feature's available (because the 5.5 docs would be updated immediately).

PR #1509, opened 2025-01-30
Describe detailed vector search metrics [AI-296]

k-jamroz:
Documentation of metrics added in https://github.com/hazelcast/hazelcast-mono/pull/3877/

### DIFF
--- a/docs/modules/ROOT/pages/list-of-metrics.adoc
+++ b/docs/modules/ROOT/pages/list-of-metrics.adoc
@@ -533,6 +533,17 @@ and some invocations may be translated to `set` if there is only 1 entry.
 |count
 |Number of search operations coordinated by this member on the vector collection
 
+|`vectorCollection.searchIndexQueryCount`
+|count
+|Number of actual searches on vector index instance (partition) in this vector collection, aggregated for all indexes on the collection.
+
+Includes search retries caused by concurrent modifications and other reasons.
+If retries were not needed it should be equal to `searchCount * non-empty partitions count`, in case of retries it will be larger.
+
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/list-of-metrics.adoc
```diff
@@ -533,6 +533,17 @@ and some invocations may be translated to `set` if there is only 1 entry.
 |count
 |Number of search operations coordinated by this member on the vector collection
 
+|`vectorCollection.searchIndexQueryCount`
+|count
+|Number of actual searches on vector index instance (partition) in this vector collection, aggregated for all indexes on the collection.
+
+Includes search retries caused by concurrent modifications and other reasons.
+If retries were not needed it should be equal to `searchCount * non-empty partitions count`, in case of retries it will be larger.
```
Rob-Hazelcast, line 541:
```suggestion
If retries were not needed it should be equal to `searchCount * non-empty partitions count`. In case of retries it will be larger.
```

docs/modules/ROOT/pages/list-of-metrics.adoc
```diff
@@ -533,6 +533,17 @@ and some invocations may be translated to `set` if there is only 1 entry.
...
+
+|`vectorCollection.searchIndexVisitedNodes`
+|count
+|Number of graph nodes visited during searches. It is equal to number of distance/score calculations during searches
```
Rob-Hazelcast, line 545:
```suggestion
|Number of graph nodes visited during searches. It is equal to the number of distance/score calculations during searches.
```

Rob-Hazelcast (approved):

PR #1510, opened 2025-01-30
Some clarifications about data persistence [HZG-257]

th0masb:
Some clarifications that data persistence is there to augment in-memory partition backups and that planned cluster shutdowns must be done cluster-wide. A rolling restart is not the same as a cluster shutdown, it is a sequence of individual member restarts and so is more appropriate to categorize as a speed improvement instead of a resiliency one.

### DIFF
--- a/docs/modules/storage/pages/persistence.adoc
+++ b/docs/modules/storage/pages/persistence.adoc
@@ -1,5 +1,5 @@
 = Persisting Data on Disk
-:description: Persistence allows individual members and whole clusters to recover data by persisting map entries, JCache data, and streaming job snapshots on disk. Members can use persisted data to recover from planned cluster-wide shutdowns, unplanned cluster-wide failures and speed up individual member restarts by reducing the volume of data sent over the network.
+:description: Persistence allows individual members and whole clusters to recover data by persisting map entries, JCache data, and streaming job snapshots on disk. Members can use persisted data to recover from planned cluster-wide shutdowns, unplanned cluster-wide failures, and to accelerate individual member restarts by reducing the volume of data sent over the network.
 :toc-levels: 3
 :page-enterprise: true
 
@@ -11,7 +11,7 @@ Data in Hazelcast is usually stored in-memory (RAM) so that it's faster to acces
 
 Clusters can use persisted data for the following scenarios:
 
-- **Persisting data across cluster-wide shutdowns**:
+- **Cluster-wide shutdowns**:
 
 ** **Planned**: A whole cluster is shut down and restarted with the same configuration, state, and data as before the restart. 
 +
@@ -21,10 +21,10 @@ In order for persistence to function correctly for planned cluster shutdowns you
 ====
 ** **Unplanned**: A cluster is restarted after all its members crash at the same time due an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
 
-- **Speeding up single member restarts**: 
+- **Accelerating single member restarts**: 
 
 ** **Planned**: During a rolling restart each cluster member is restarted one by one for scenarios such as installing an operating system patch or new hardware. xref:maintain-cluster:rolling-upgrades.adoc[Rolling upgrades] are an example of a rolling restart.
-** **Unplanned**: A member may crash or terminate unexpectedly at any time, using persistence allows faster recovery.
+** **Unplanned**: A single member may crash or terminate unexpectedly at any time, using persistence allows faster recovery by using the data on disk where possible. In some cases a member can recover entirely using the data on disk removing the need to transfer data over the network.
 
 
 == Supported Features
--- a/docs/modules/storage/pages/persistence.adoc
+++ b/docs/modules/storage/pages/persistence.adoc
@@ -1,5 +1,5 @@
 = Persisting Data on Disk
-:description: Persistence allows individual members and whole clusters to recover data by persisting map entries, JCache data, and streaming job snapshots on disk. Members can use persisted data to recover from cluster-wide shutdowns and speed up individual member restarts by reducing the volume of data sent over the network.
+:description: Persistence allows individual members and whole clusters to recover data by persisting map entries, JCache data, and streaming job snapshots on disk. Members can use persisted data to recover from planned cluster-wide shutdowns, unplanned cluster-wide failures and speed up individual member restarts by reducing the volume of data sent over the network.
 :toc-levels: 3
 :page-enterprise: true
 
@@ -7,7 +7,7 @@
 
 == Why Persist Data
 
-Data in Hazelcast is usually stored in-memory (RAM) so that it's faster to access. However, data in RAM is volatile, meaning that when a member shuts down, its data is lost. In-memory partition backups are still the primary method for data resilience in Hazelcast but additionally persisting data to disk can be useful in certain situations. When you persist data on disk, members can load it upon a restart back into memory directly instead of relying on other active members sending it over the network.
+Data in Hazelcast is usually stored in-memory (RAM) so that it's faster to access. However, data in RAM is volatile, meaning that when a member shuts down, its data is lost. In-memory partition backups are the primary method for data resilience in Hazelcast but additionally persisting data to disk can be useful in certain situations. When you persist data on disk, members can load it upon a restart instead of relying on other active members sending it over the network. In the case of cluster-wide failure - e.g. due to data centre power outage - where we cannot rely on receiving the data from other active members, persistence means that we can recover data that would otherwise be irreversibly lost.
 
 Clusters can use persisted data for the following scenarios:
 
@@ -17,7 +17,7 @@ Clusters can use persisted data for the following scenarios:
 +
 [IMPORTANT]
 ====
-In order for persistence to function correctly for planned cluster shutdowns you must xref:maintain-cluster:shutdown.adoc[shutdown the entire cluster at once].
+In order for persistence to function correctly for planned cluster shutdowns you must first put the cluster into `PASSIVE` state before shutting down members. A list of methods for gracefully shutting down a cluster can be found in the xref:maintain-cluster:shutdown.adoc[shutdown] docs.
 ====
 ** **Unplanned**: A cluster is restarted after all its members crash at the same time due an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
 
--- a/docs/modules/storage/pages/persistence.adoc
+++ b/docs/modules/storage/pages/persistence.adoc
@@ -7,7 +7,9 @@
 
 == Why Persist Data
 
-Data in Hazelcast is usually stored in-memory (RAM) so that it's faster to access. However, data in RAM is volatile, meaning that when a member shuts down, its data is lost. In-memory partition backups are the primary method for data resilience in Hazelcast but additionally persisting data to disk can be useful in certain situations. When you persist data on disk, members can load it upon a restart instead of relying on other active members sending it over the network. In the case of cluster-wide failure - e.g. due to data centre power outage - where we cannot rely on receiving the data from other active members, persistence means that we can recover data that would otherwise be irreversibly lost.
+Data in Hazelcast is usually stored in-memory (RAM) so that it's faster to access. However, data in RAM is volatile, meaning that when a member shuts down, its data is lost. In-memory partition backups are the primary method for data resilience in Hazelcast, but additionally persisting data to disk can reduce downtime and data loss in certain scenarios.
+
+When you persist data on disk, members can load it on restart, instead of waiting for other active members to send it over the network. In a cluster-wide failure - e.g. due to a data center power outage - there are no active members to receive data from, so persistence means that we can recover data that would otherwise be lost.
 
 Clusters can use persisted data for the following scenarios:
 
@@ -17,11 +19,11 @@ Clusters can use persisted data for the following scenarios:
 +
 [IMPORTANT]
 ====
-In order for persistence to function correctly for planned cluster shutdowns you must first put the cluster into `PASSIVE` state before shutting down members. A list of methods for gracefully shutting down a cluster can be found in the xref:maintain-cluster:shutdown.adoc[shutdown] docs.
+In order for persistence to function correctly for planned cluster shutdowns you must first put the cluster into `PASSIVE` state before shutting down members. A list of methods for gracefully shutting down a cluster can be found in the xref:maintain-cluster:shutdown.adoc[shutdown] documentation.
 ====
-** **Unplanned**: A cluster is restarted after all its members crash at the same time due an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
+** **Unplanned**: A cluster is restarted after all its members crash at the same time due to an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
 
-- **Accelerating single member restarts**: 
+- **Single member restarts**: 
 
 ** **Planned**: During a rolling restart each cluster member is restarted one by one for scenarios such as installing an operating system patch or new hardware. xref:maintain-cluster:rolling-upgrades.adoc[Rolling upgrades] are an example of a rolling restart.
 ** **Unplanned**: A single member may crash or terminate unexpectedly at any time, using persistence allows faster recovery by using the data on disk where possible. In some cases a member can recover entirely using the data on disk removing the need to transfer data over the network.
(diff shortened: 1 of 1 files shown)


docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
 = Persisting Data on Disk
-:description: Persistence allows individual members and whole clusters to recover data by persisting map entries, JCache data, and streaming job snapshots on disk. Members can use persisted data to recover from a planned shutdown (including rolling upgrades), a sudden cluster-wide crash, or a single member failure.
+:description: Persistence allows individual members and whole clusters to recover data by persisting map entries, JCache data, and streaming job snapshots on disk. Members can use persisted data to recover from cluster-wide shutdowns and speed up individual member restarts by reducing the volume of data sent over the network.
```
JamesHazelcast, line 2: I think your addition is helpful context to have, but I think the 3 scenarios mentioned are important and should not be replaced here. Your context relates to planned shutdowns/rolling upgrades, but sudden cluster-wide crashes and single member failures should be supported.

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
 :toc-levels: 3
 :page-enterprise: true
 
 {description}
 
 == Why Persist Data
 
-Data in Hazelcast is usually stored in-memory (RAM) so that it's faster to access. However, data in RAM is volatile, meaning that when one or more members shut down, their data is lost. When you persist data on disk, members can load it upon a restart and continue to operate as usual.
+Data in Hazelcast is usually stored in-memory (RAM) so that it's faster to access. However, data in RAM is volatile, meaning that when a member shuts down, its data is lost. In-memory partition backups are still the primary method for data resilience in Hazelcast but additionally persisting data to disk can be useful in certain situations. When you persist data on disk, members can load it upon a restart back into memory directly instead of relying on other active members sending it over the network.
```
JamesHazelcast, line 10:
```suggestion
Data in Hazelcast is usually stored in-memory (RAM) so that it's faster to access. However, data in RAM is volatile, meaning that when a member shuts down, its data is lost. In-memory partition backups are the primary method for data resilience in Hazelcast but additionally persisting data to disk can be useful in certain situations. When you persist data on disk, members can load it back into memory upon a restart  instead of relying on other active members sending it over the network.
```

Some wording tweaks - since this is under "why persist data", it's probably worth taking this opportunity to add a bit of context about the benefit of persistence in the event of a cluster-wide failure (i.e. DC loses power), as it allows recovery of data that would otherwise be irreversibly lost.

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
 
-Clusters can use persisted data to recover from the following scenarios:
+Clusters can use persisted data for the following scenarios:
 
-- **Cluster-wide shutdown**:
+- **Persisting data across cluster-wide shutdowns**:
 
-** **Planned**: A whole cluster is shut down and restarted with the same configuration, state, and data as before the restart. A planned shutdown includes rolling restarts, where the cluster is restarted member by member for scenarios such as installing an operating system patch or new hardware. xref:maintain-cluster:rolling-upgrades.adoc[Rolling upgrades] are an example of a rolling restart.
+** **Planned**: A whole cluster is shut down and restarted with the same configuration, state, and data as before the restart.
```
JamesHazelcast, line 16: I'm not sure I agree with the removal of information here - when the cluster state is `PASSIVE`, it is perfectly fine to perform a rolling restart with persistence.

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
-- **Cluster-wide shutdown**:
+- **Persisting data across cluster-wide shutdowns**:
 
-** **Planned**: A whole cluster is shut down and restarted with the same configuration, state, and data as before the restart. A planned shutdown includes rolling restarts, where the cluster is restarted member by member for scenarios such as installing an operating system patch or new hardware. xref:maintain-cluster:rolling-upgrades.adoc[Rolling upgrades] are an example of a rolling restart.
+** **Planned**: A whole cluster is shut down and restarted with the same configuration, state, and data as before the restart. 
++
+[IMPORTANT]
+====
+In order for persistence to function correctly for planned cluster shutdowns you must xref:maintain-cluster:shutdown.adoc[shutdown the entire cluster at once].
+====
```
JamesHazelcast, line 21: This is not accurate, the cluster state must be `PASSIVE` and then it is acceptable to shut down members individually (MC will move the cluster state to `PASSIVE` for restarts, for example).

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
 ** **Unplanned**: A cluster is restarted after all its members crash at the same time due an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
-- **A single member restart**: A single member is restarted for whatever reason.
+
+- **Speeding up single member restarts**: 
+
+** **Planned**: During a rolling restart each cluster member is restarted one by one for scenarios such as installing an operating system patch or new hardware. xref:maintain-cluster:rolling-upgrades.adoc[Rolling upgrades] are an example of a rolling restart.
+** **Unplanned**: A member may crash or terminate unexpectedly at any time, using persistence allows faster recovery.
```
JamesHazelcast, line 27: This also depends on configuration - if rebalancing is delayed, or the cluster is in `PASSIVE` state, it loads all data from disk without receiving any from other members

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
+[IMPORTANT]
+====
+In order for persistence to function correctly for planned cluster shutdowns you must xref:maintain-cluster:shutdown.adoc[shutdown the entire cluster at once].
+====
 ** **Unplanned**: A cluster is restarted after all its members crash at the same time due an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
-- **A single member restart**: A single member is restarted for whatever reason.
+
+- **Speeding up single member restarts**: 
+
+** **Planned**: During a rolling restart each cluster member is restarted one by one for scenarios such as installing an operating system patch or new hardware. xref:maintain-cluster:rolling-upgrades.adoc[Rolling upgrades] are an example of a rolling restart.
```
JamesHazelcast, line 26: This is part of a cluster-wide shutdown, we do not want to create confusion between these topics imo

JamesHazelcast (changes requested):
Thanks for working to improve our documentation on this topic @th0masb - a few comments from me on this round

docs/modules/storage/pages/persistence.adoc
th0masb: Yeah this could probably be worded better, the link I included is about how to shutdown an entire cluster by putting it into passive mode first and then shutting down the members, one of the ways in that doc is using MC. I can reword it to be more explicit about the passive state and then provide the link as a list of ways you might go about shutting down the cluster.

docs/modules/storage/pages/persistence.adoc
th0masb: I was under the impression that even if the member crashed when the cluster was in active state and rebalancing was not delayed the persisted data could still be used to speed up the member rejoining the cluster. Aren't merkle trees compared to reduce network traffic as much as possible?

docs/modules/storage/pages/persistence.adoc
th0masb: I found it confusing to read the existing way, how is a rolling restart a cluster wide shutdown if only one member at any time is stopped? Is it mandatory to put the cluster into passive mode for a rolling restart? It seems like it work fine if the cluster was still in active mode.

docs/modules/storage/pages/persistence.adoc
th0masb: I didn't remove it, I moved it under the single member restart category. Since a rolling restart is a sequence of single member restarts instead of the cluster restarting.

docs/modules/storage/pages/persistence.adoc
th0masb: So I kind of merged the planned and unplanned shutdowns/restarts together so I can separate those again. It's not clear that 

> Members can use persisted data to recover from ... a single member failure

is useful here. A cluster with backups enabled can already recover from a single member failure, persistence doesn't add that feature, it just speeds that process up for large volumes of data. Having this here makes it seem like persistence is an alternative to configuring in-memory backups.

docs/modules/storage/pages/persistence.adoc
th0masb: Reworded this, let me know what you think.

docs/modules/storage/pages/persistence.adoc
JamesHazelcast: Agreed that we want to be clear that persistence _can_ speed up single member failure recovery, but the backup mechanism is the primary recovery method 👍

docs/modules/storage/pages/persistence.adoc
JamesHazelcast: Yes that's correct, I'm saying that in addition to this "faster recovery" case, there are conditions where it recovers entirely using its own data, not just using it as a speedup mechanism.

docs/modules/storage/pages/persistence.adoc
JamesHazelcast: I suppose there's arguments to be made both ways - I can see why you see it as a series of single member restarts, but I can also see that it does _eventually_ result in the entire cluster restarting, so it could be seen as a cluster-wide shutdown (restart). I'm not too fussed, we can let docs team review this aspect and see if it makes sense to them 👍

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
 = Persisting Data on Disk
-:description: Persistence allows individual members and whole clusters to recover data by persisting map entries, JCache data, and streaming job snapshots on disk. Members can use persisted data to recover from a planned shutdown (including rolling upgrades), a sudden cluster-wide crash, or a single member failure.
+:description: Persistence allows individual members and whole clusters to recover data by persisting map entries, JCache data, and streaming job snapshots on disk. Members can use persisted data to recover from planned cluster-wide shutdowns, unplanned cluster-wide failures and speed up individual member restarts by reducing the volume of data sent over the network.
```
JamesHazelcast, line 2:
```suggestion
:description: Persistence allows individual members and whole clusters to recover data by persisting map entries, JCache data, and streaming job snapshots on disk. Members can use persisted data to recover from planned cluster-wide shutdowns, unplanned cluster-wide failures, and to accelerate individual member restarts by reducing the volume of data sent over the network.
```
minor rewording for clarity

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
+Data in Hazelcast is usually stored in-memory (RAM) so that it's faster to access. However, data in RAM is volatile, meaning that when a member shuts down, its data is lost. In-memory partition backups are the primary method for data resilience in Hazelcast but additionally persisting data to disk can be useful in certain situations. When you persist data on disk, members can load it upon a restart instead of relying on other active members sending it over the network. In the case of cluster-wide failure - e.g. due to data centre power outage - where we cannot rely on receiving the data from other active members, persistence means that we can recover data that would otherwise be irreversibly lost.
 
-Clusters can use persisted data to recover from the following scenarios:
+Clusters can use persisted data for the following scenarios:
 
-- **Cluster-wide shutdown**:
+- **Persisting data across cluster-wide shutdowns**:
```
JamesHazelcast, line 14:
```suggestion
- **Cluster-wide shutdowns**:
```
Nit: we already start this list with "clusters can use persisted data for...", so I don't think we need to have "Persisting data across" here

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
+** **Planned**: A whole cluster is shut down and restarted with the same configuration, state, and data as before the restart. 
++
+[IMPORTANT]
+====
+In order for persistence to function correctly for planned cluster shutdowns you must first put the cluster into `PASSIVE` state before shutting down members. A list of methods for gracefully shutting down a cluster can be found in the xref:maintain-cluster:shutdown.adoc[shutdown] docs.
+====
 ** **Unplanned**: A cluster is restarted after all its members crash at the same time due an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
-- **A single member restart**: A single member is restarted for whatever reason.
+
+- **Speeding up single member restarts**:
```
JamesHazelcast, line 24:
```suggestion
- **Accelerating single member restarts**: 
```
Nit: more appealing wording, at least imo

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
+====
+In order for persistence to function correctly for planned cluster shutdowns you must first put the cluster into `PASSIVE` state before shutting down members. A list of methods for gracefully shutting down a cluster can be found in the xref:maintain-cluster:shutdown.adoc[shutdown] docs.
+====
 ** **Unplanned**: A cluster is restarted after all its members crash at the same time due an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
-- **A single member restart**: A single member is restarted for whatever reason.
+
+- **Speeding up single member restarts**: 
+
+** **Planned**: During a rolling restart each cluster member is restarted one by one for scenarios such as installing an operating system patch or new hardware. xref:maintain-cluster:rolling-upgrades.adoc[Rolling upgrades] are an example of a rolling restart.
+** **Unplanned**: A member may crash or terminate unexpectedly at any time, using persistence allows faster recovery.
```
JamesHazelcast, line 27:
```suggestion
** **Unplanned**: A single member may crash or terminate unexpectedly at any time, using persistence allows faster recovery.
```
We only support single member failure in terms of persistence functionality

JamesHazelcast (approved):
A few more comments from me, but mainly nits so approving in advance - thanks for raising this @th0masb and improving our docs 👍

th0masb:
@Rob-Hazelcast I think this is ready now, as for backporting I am unsure of the policy for the docs but since this is a clarification on existing functionality and the page content looks to be the same since 5.3 I would recommend backporting to 5.3.

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
+** **Planned**: A whole cluster is shut down and restarted with the same configuration, state, and data as before the restart. 
++
+[IMPORTANT]
+====
+In order for persistence to function correctly for planned cluster shutdowns you must first put the cluster into `PASSIVE` state before shutting down members. A list of methods for gracefully shutting down a cluster can be found in the xref:maintain-cluster:shutdown.adoc[shutdown] docs.
+====
 ** **Unplanned**: A cluster is restarted after all its members crash at the same time due an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
-- **A single member restart**: A single member is restarted for whatever reason.
+
+- **Accelerating single member restarts**:
```
Rob-Hazelcast, line 24:
```suggestion
- **Single member restarts**: 
```

The two bullets should be consistent. Removing the verb here; could alternatively add a verb ("recovering from cluster-wide shutdowns") above.

docs/modules/storage/pages/persistence.adoc
```diff
@@ -1,21 +1,31 @@
...
+- **Cluster-wide shutdowns**:
 
-** **Planned**: A whole cluster is shut down and restarted with the same configuration, state, and data as before the restart. A planned shutdown includes rolling restarts, where the cluster is restarted member by member for scenarios such as installing an operating system patch or new hardware. xref:maintain-cluster:rolling-upgrades.adoc[Rolling upgrades] are an example of a rolling restart.
+** **Planned**: A whole cluster is shut down and restarted with the same configuration, state, and data as before the restart. 
++
+[IMPORTANT]
+====
+In order for persistence to function correctly for planned cluster shutdowns you must first put the cluster into `PASSIVE` state before shutting down members. A list of methods for gracefully shutting down a cluster can be found in the xref:maintain-cluster:shutdown.adoc[shutdown] docs.
+====
 ** **Unplanned**: A cluster is restarted after all its members crash at the same time due an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
```
Rob-Hazelcast, line 22:
```suggestion
** **Unplanned**: A cluster is restarted after all its members crash at the same time due to an event such as a power outage. Note that some data loss is expected unless `fsync` is set to `true`. For more information, see xref:storage:configuring-persistence.adoc#data-structures[Data structure options].
```

Rob-Hazelcast (approved):
Suggested some small edits for clarify but generally looks good, thanks!

PR #1511, opened 2025-01-30
REL-177: 5.5.3 Release Notes

oliverhowell:
- Create release notes for 5.5.3
- Update template and update previous 5.5 instances


docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,37 @@
...
+* 14775: IllegalStateException - NodeEngine context is not available for Namespaces in Hazelcast 5.5.0 (SUP-572)
+
+== Security
+
+* Security Fix for CVE-2023-45676
+We have resolved https://github.com/advisories/GHSA-gcg6-xv4f-f749[CVE-2023-45676], a vulnerability related to improper authorization checks in certain scenarios. This issue could potentially allow attackers to bypass expected permission restrictions.
+
+// new sections we could consider including
+
+// == Dependency changes
```
oliverhowell, line 33: I'm going to take this out as we can't technically do it rn

docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,37 @@
...
+// == Version changes
```
oliverhowell, line 34: We could highlight e.g. JDK or other 3rd party version changes here, or even changes in support policy if applicable

docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,37 @@
...
+// == Downloads
```
oliverhowell, line 35: We could have an explicit downloads link here (or just keep in intro as above)

docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,37 @@
+= 5.5.2 Release Notes
+:description: These release notes list any new features, enhancements, fixes, and breaking changes implemented between version 5.5.2 and version 5.5.3 of Hazelcast Platform.
+
+{description}
```
oliverhowell, line 4: This was the description 'tip' I was obliquely referring to in your PR today @amandalindsay

amandalindsay: Thanks, yes, and I replied because sometimes (but not always) description is surfaced as a {variable} as above and appears to the user, whereas in other examples it is not and is never displayed. I'm wondering if description is sometimes used internally for some wizardry? One to discuss.

docs/modules/release-notes/pages/5-5-0.adoc
```diff
@@ -3,6 +3,12 @@
 
 {description}
 
+This 5.5.0 release is available for {enterprise-product-name} and {open-source-product-name}.
```
Rob-Hazelcast, line 6: "This release" or "The 5.5.0 release", I think. This implies it's one of many 5.5.0 releases.

docs/modules/release-notes/pages/5-5-3.adoc
oliverhowell: It will appear in source but never in rendered output, unless some strange local build artefact?

Rob-Hazelcast: It's used as the page description in external search engine results. I'd expect to see it used for mouse-over text too but it looks like we haven't implemented that. Might be accessible to screen reader users too.

IMO we should standardize treating it as the first paragraph of the section and include the `{description}` tag.

docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,37 @@
+= 5.5.2 Release Notes
+:description: These release notes list any new features, enhancements, fixes, and breaking changes implemented between version 5.5.2 and version 5.5.3 of Hazelcast Platform.
```
Rob-Hazelcast, line 2: For ease of maintenance and to reduce the risk of errors, can we just say "in this release of" rather than this in between gubbins?

docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,37 @@
+= 5.5.2 Release Notes
```
Rob-Hazelcast, line 1:
```suggestion
= 5.5.3 Release Notes
```

docs/modules/release-notes/pages/5-5-2.adoc
```diff
@@ -3,6 +3,10 @@
 
 {description}
 
+This is a maintenance release for {Enterprise Edition}.
```
Rob-Hazelcast, line 6: This isn't a variable, the curly brackets get rendered in the output

docs/modules/release-notes/pages/5-5-3.adoc
oliverhowell: That was an important one to spot lol

docs/modules/release-notes/pages/5-5-2.adoc
oliverhowell: There's a gotcha if you use the VSC auto-complete where it actually replaces the variable, it's a pain

docs/modules/release-notes/pages/5-5-3.adoc
oliverhowell: Adding this to the TBD list - original motivation was to make explicitly clear that each release was independent (i.e. had to be installed separately) but tbf not sure that this text really achieves that. Also minor but changing to your suggestion would re-introduce the same 'this version' text you advised against above.

docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,30 @@
...
+
+For more details on new features, see xref:ROOT:whats-new.adoc[What's new in 5.5].
+
+== Known issues
+* There is currently a known issue that has potential for causing a memory leak in Vector collections in some scenarios (AI-133). For more info, including a workaround, see xref:data-structures:vector-collections.adoc#known-issue[Vector Collection Known Issue].
+
+== Fixes
+* Fixed task config publishing in Kafka Connect (HZG-220)
+* Fixed migration issues with UCN classes in InMemoryFormat.OBJECT (HZG-227)
+* 14775: IllegalStateException - NodeEngine context is not available for Namespaces in Hazelcast 5.5.0 (SUP-572)
```
JackPGreen, line 25: [The ticket seems to suggest this was fixed in `5.5.2`](https://hazelcast.atlassian.net/browse/SUP-572)?

docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,30 @@
...
+
+Release date - 2025-02-10
+
+This is a maintenance release for {enterprise-product-name}. 
+
+For help downloading Hazelcast {enterprise-product-name}, see xref:getting-started:install-enterprise.adoc[] or https://hazelcast.com/trial-request/?utm_source=docs-website[request a trial license key].
+
+== Enhancements
+* CP: Enhanced error message displayed when the CP system is reset for clients using FencedLock (SUP-117)
+* Updated BouncyCastle to support FIPS compliance (license.extractor.version set to 7.0.2) (SUP-667)
```
JackPGreen, line 14:
```suggestion
* Updated BouncyCastle to support FIPS compliance (`license.extractor.version` set to 7.0.2) (SUP-667)
```

I prefer code formatting.

Also I think `license.extractor.version` makes no sense as it's an internal tool where the source (deliberately) isn't released. Maybe we should just drop it?

docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,30 @@
...
+* Added IPv6 support to the AWS discovery plugin (SUP-653)
+
+For more details on new features, see xref:ROOT:whats-new.adoc[What's new in 5.5].
+
+== Known issues
+* There is currently a known issue that has potential for causing a memory leak in Vector collections in some scenarios (AI-133). For more info, including a workaround, see xref:data-structures:vector-collections.adoc#known-issue[Vector Collection Known Issue].
+
+== Fixes
+* Fixed task config publishing in Kafka Connect (HZG-220)
+* Fixed migration issues with UCN classes in InMemoryFormat.OBJECT (HZG-227)
```
JackPGreen, line 24:
```suggestion
* Fixed migration issues with UCN classes in `InMemoryFormat.OBJECT` (HZG-227)
```

Code formatting.

docs/modules/release-notes/pages/5-5-3.adoc
```diff
@@ -0,0 +1,30 @@
...
+* 14775: IllegalStateException - NodeEngine context is not available for Namespaces in Hazelcast 5.5.0 (SUP-572)
+
+== Security
+
+* Security Fix for CVE-2023-45676
+We have resolved https://github.com/advisories/GHSA-gcg6-xv4f-f749[CVE-2023-45676], a vulnerability related to improper authorization checks in certain scenarios. This issue could potentially allow attackers to bypass expected permission restrictions.
```
JackPGreen, line 30: I think we need to be more explicit that we've fixed a CVE in a _depdency_. At the moment it reads like _Hazelcast_ had an improper authorization check which would be... more noteworthy.

JackPGreen (changes requested):

docs/modules/release-notes/pages/5-5-3.adoc
oliverhowell: Agree, good spot

JackPGreen (approved):

docs/modules/release-notes/pages/5-5-3.adoc
oliverhowell: Yeah, that's exactly the standard I'm recommending above - it's not consistently applied everywhere but it just helps make sure we're writing and re-using good descriptions.

docs/modules/release-notes/pages/5-5-0.adoc
```diff
@@ -3,6 +3,12 @@
 
 {description}
 
+The 5.5.0 release is available for {enterprise-product-name} and {open-source-product-name}. 
+
+For help downloading Hazelcast {enterprise-product-name}, see xref:getting-started:install-enterprise.adoc[] or https://hazelcast.com/trial-request/?utm_source=docs-website[request a trial license key].
+
+For help downloading Hazelcast {open-source-product-name}, see xref:getting-started:install-hazelcast.adoc[].
+
 == New Features
```
amandalindsay, line 12: Are we using sentence case on titles e.g. Release notes and Known issues?

amandalindsay (approved):
Looks good, one tiny little comment, not important

docs/modules/release-notes/pages/5-5-0.adoc
oliverhowell: Yes! I should really fix the rest of these files since I have them open

PR #1523, opened 2025-02-05
Improve UCN migration page + fix indentation [SUP-799]

JamesHazelcast:
Adds more information about the "gotchas" of migration from UCD to UCN, and also fixes incorrect 3 space indentation for YAML configuration examples in UCN documentation.

Fixes https://hazelcast.atlassian.net/browse/SUP-799


gareth-johnston (approved):

JackPGreen (approved):

Rob-Hazelcast (approved):
Pushed some copy edits but generally looks good, thanks!

docs/modules/clusters/pages/ucn-migrate-ucd.adoc
```diff
@@ -5,6 +5,13 @@
...
+* {ucn} does not support loading resources from the classpath, and they should be defined from static locations instead. Resources, including classpath classes, can still be uploaded by clients to the cluster but this will need to be done through the programmatic API. See xref:clusters:ucn-dynamic-client.adoc[] for more information.
+* Resources in {ucn} can be dynamically added or replaced at runtime. This provides various benefits, but care should be taken to ensure that
+resources are not removed from actively running operations as this could cause errors. See xref:clusters:ucn-dynamic-config.adoc[] for more information.
```
oliverhowell:
```suggestion
* {ucn} does not have any configuration elements for Hazelcast Clients as all static configuration is managed in member configurations.
This means that resources deployed statically need to be referenceable by all members of the cluster. For more information, see xref:clusters:ucn-member-side.adoc[].
* Resources deployed with {ucd} were essentially added to the member classpath. In contrast, {ucn} introduces the Namespace concept which allows these resources to be separated. This separation allows conflicting classes to exist in separate namespaces without issue, but data structures need to reference the correct namespace to access the relevant resources. To achieve similar behaviour to {ucd} in {ucn}, configure your resources in the `default` namespace, which means they will be available to all data structures which do not explicitly define a namespace already.
* {ucn} does not support loading resources from the classpath, and they should be defined from static locations instead. Resources, including classpath classes, can still be uploaded by clients to the cluster but you will  need to do this using the programmatic API. For more information, see xref:clusters:ucn-dynamic-client.adoc[].
* Resources in {ucn} can be dynamically added or replaced at runtime. This provides various benefits, but care should be taken to ensure that
resources are not removed from actively running operations as this could cause errors. For more information, see xref:clusters:ucn-dynamic-config.adoc[].
```

oliverhowell (changes requested):
Minor changes but thanks for such a quick turnaround!

docs/modules/clusters/pages/ucn-migrate-ucd.adoc
JamesHazelcast: Thanks for the feedback Oli, applied here: https://github.com/hazelcast/hz-docs/pull/1523/commits/4c407833519d4861ebe21fccd6110c1274db5ead

oliverhowell (approved):

PR #1536, opened 2025-02-12
Correct maximum recommended clients per member

JamesHazelcast:
The information currently presented in the docs is outdated, as client tasks are handled by 3 executors instead of 1 now (since 2018).

This section also did not mention I/O thread counts which are also relevant when determining client connection limits.

This PR updates this section with more details and up-to-date information.


amandalindsay (approved):
Looks good

gbarnett-hz (approved):

PR #1538, opened 2025-02-12
Update Javadoc links to use OS version [DI-425]

JackPGreen:
The OS & EE Javadocs are versioned separately - reflecting the latest version of each product.
However in the documentation, we use the (problematic) `full-version` property when linking to the _OS_ Javadoc which is the latest _EE_ version.
[We no longer publish (incorrectly versioned) OS Javadoc as part of an EE release](https://hazelcast.atlassian.net/browse/DI-328) which is why this problem is only now visible.

Replaced:
- https://docs.hazelcast.org/docs/{full-version}/javadoc
- https://docs.hazelcast.org/docs/{page-latest-supported-java-client}/javadoc
- https://javadoc.io/doc/com.hazelcast/hazelcast/{full-version}

With `https://docs.hazelcast.org/docs/{os-version}/javadoc` as are all OS content.

Tested [by running external link check on branch](https://github.com/hazelcast/hz-docs/actions/runs/13295463298/job/37126237227).

Ideally this _should_ be backported to all maintenance branches, but the PR is unlikely to merge without conflicts. If required, re-applying the change (now it's scoped) is trivial.

Fixes: [DI-425](https://hazelcast.atlassian.net/browse/DI-425)

[DI-425]: https://hazelcast.atlassian.net/browse/DI-425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -11,7 +11,8 @@ asciidoc:
     # The full major.minor.patch version, which is used as a variable in the docs for things like download links
     full-version: '6.0.0-SNAPSHOT'
     os-version: '5.5.0'
-    ee-version: '5.5.2'
+    # TODO REVERT
+    ee-version: '5.5.3'
     # The minor.patch version, which is used as a variable in the docs for things like file versions
     minor-version: '6.0-SNAPSHOT'
     # The snapshot version for installing with brew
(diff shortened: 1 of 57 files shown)


oliverhowell (approved):
Thanks @JackPGreen

PR #1539, opened 2025-02-12
Fix MC Docker tags

JackPGreen:
Update MC Docker tags to `hazelcast/management-center:{page-latest-supported-mc}`.

Updated from:
- `latest-snapshot` - should use released artefacts
- `full-version` - the versions of Hazelcast platform and MC are not tied (e.g. no `5.5.2` MC)

### DIFF
--- a/docs/modules/getting-started/pages/get-started-docker.adoc
+++ b/docs/modules/getting-started/pages/get-started-docker.adoc
@@ -613,7 +613,7 @@ In this step, you start a local instance of Management Center and use it to view
 ----
 docker run \
     --network hazelcast-network \
-    -p 8080:8080 hazelcast/management-center:latest-snapshot
+    -p 8080:8080 hazelcast/management-center:{page-latest-supported-mc}
 ----
 
 . In a web browser, go to localhost:8080 and enable dev mode.
--- a/docs/modules/getting-started/pages/get-started-docker.adoc
+++ b/docs/modules/getting-started/pages/get-started-docker.adoc
@@ -609,7 +609,7 @@ In this step, you start a local instance of Management Center and use it to view
 
 . Start Management Center.
 +
-[source,shell]
+[source,shell,subs="attributes+"]
 ----
 docker run \
     --network hazelcast-network \
--- a/docs/modules/getting-started/pages/persistence.adoc
+++ b/docs/modules/getting-started/pages/persistence.adoc
@@ -160,7 +160,7 @@ Now, check the map in Management Center.
 ----
 docker run \
     --network hazelcast-network \
-    -p 8080:8080 hazelcast/management-center:latest-snapshot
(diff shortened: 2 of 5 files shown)


docs/modules/getting-started/pages/get-started-docker.adoc
```diff
@@ -613,7 +613,7 @@ In this step, you start a local instance of Management Center and use it to view
 ----
 docker run \
     --network hazelcast-network \
-    -p 8080:8080 hazelcast/management-center:latest-snapshot
+    -p 8080:8080 hazelcast/management-center:{page-latest-supported-mc}
```
Rob-Hazelcast, line 616: These don't appear to resolve when used in code samples. The original text wasn't a variable - was that just a mistake, or is `latest-snapshot` a valid input to Docker?

@oliverhowell FYI

JackPGreen: Latest snapshot isn’t a valid tag (presumably any more). I didn’t actually test it in the deployment though - we could use latest instead?

JackPGreen: Looks like I’m wrong and it works - https://hub.docker.com/layers/hazelcast/management-center/latest-snapshot/images/sha256-5182aec7dfc667ade16b61ef2cb7b49e5936950af2b8a6d9fcbd4a564e474bd5

still shouldn’t be a snapshot tho.

JackPGreen: And there is no `latest`… so if we can’t use substitution, maybe this is the best of a bad bunch?

JackPGreen: > As a separate task, it might be worth us searching the docs for variables inside code blocks (regex in the source files?) and/or broken variables on the site (IDK).

Good idea, tried but could only find https://github.com/hazelcast/clc-docs/pull/50

I'm surprised this issue doesn't occur more often tbh.

Here was how I tested it btw - just a simple set of (hacky) commands to build the docs locally, and scrape the output text for anything in curly braces and then log it + where it was found:
```
npm install
npm run-script build-local

find test -name "*.html" | while read -r file; do
    lynx -dump -listonly -nonumbers "${file}" | { grep --extended-regexp "\{[^}]*\}" && echo " in ${file}" || test $? = 1; }
done
```

Rob-Hazelcast: Great, thank you for sorting all that out.

> I think this is a non-issue now the substitution works?

I might still expect someone in the relevant team to sign off this sort of change, but the most important thing is that the updated command is tested and we're not breaking anything.

Rob-Hazelcast (approved):

PR #1574, opened 2025-02-24
Clarify UCN object refresh [HZG-210]

TomaszGaweda:
Classes' bytecode will be refreshed, but old instances will use old bytecode.

### DIFF
--- a/docs/modules/clusters/partials/ucn-obj-refresh-caution.adoc
+++ b/docs/modules/clusters/partials/ucn-obj-refresh-caution.adoc
@@ -1,3 +1,3 @@
 CAUTION: {ucn} can replace resources dynamically, but this does not affect the lifecycle of existing instances using older class versions.
 This means that the creation of new objects will use new resources (latest bytecode), but existing objects won't be removed and recreated automatically. In this case, manual intervention is needed, such as destroying and recreating the relevant objects, or restarting Hazelcast members.
-This is particularly applicable to long-lived objects tied to the lifecycle of Hazelcast data structures, such as `MapLoader`s: deploying a new version won't replace an existing `MapLoader`, and to refresh such an object, a cluster restart is required.
+This is particularly applicable to long-lived objects tied to the lifecycle of Hazelcast data structures, such as a `MapLoader`: deploying a new version won't replace an existing `MapLoader`, and to refresh such an object, a cluster restart is required.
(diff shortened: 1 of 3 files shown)


JamesHazelcast (changes requested):
Thanks for this @TomaszGaweda - I'd like to see more detail provided in this warning, specifically noting the difference between UCN refreshing **resources**, but not destroying and recreating **objects** with the new resources. I'd also like to see some information about manual intervention (restarting cluster, manually destroying recreating objects, etc.) being required unless the user code function's documentation explicitly states that refreshing the resource will refresh the associated objects as well.

Although we don't want to provide an exhaustive list of instances, it might be worth giving the `MapLoader` as an example, noting that it is tied to the lifecycle of an `IMap` and therefore requites a member restart to update - this could be an opportunity to mention that any data-structure specific UCFs are likely also tied to the data structure lifecycle.

TomaszGaweda:
How about now @JamesHazelcast ?

docs/modules/clusters/partials/ucn-obj-refresh-caution.adoc
```diff
@@ -0,0 +1,3 @@
+CAUTION: {ucn} can replace resources dynamically, but that does not affect the lifecycle of already existing instances of old classes.
+That means that creation of new objects will use new resources (new bytecode), but older objects won't be removed and recreated; in this case a manual intervention is needed.
+This caution applies especially to long-living objects, tied to a lifecycle of Hazelcast structures, such as `IMap`'s MapLoader`s: deploying new version won't replace already added MapLoader and if user needs to refresh such object, a cluster restart is needed.
```
JamesHazelcast, line 3:
```suggestion
CAUTION: {ucn} can replace resources dynamically, but that does not affect the lifecycle of existing instances using older class versions.
This means that the creation of new objects will use new resources (latest bytecode), but older objects won't be removed and recreated automatically; in this case a manual intervention is needed, such as destroying and recreating relevant objects, or restarting Hazelcast members.
This is particularly applicable to long-living objects tied to the lifecycle of Hazelcast data structures, such as `IMap`'s `MapLoader`s: deploying a new version won't replace an existing `MapLoader`, and to refresh such an object, a cluster restart would be required.
```

A bit of rewording from me (although the docs team might have something to say about it 😄)

JamesHazelcast (approved):
The caution LGTM, I made some wording suggestions, but approving in advance - thanks Tomasz!

Rob-Hazelcast (approved):
Approving because this is a good addition, but I'm concerned we're being vague about requirements. We say "such as" recreating objects or restarting members, and then bring up the MapLoader example that supposedly requires restarting the whole cluster. How are users supposed to know what intervention is required for a given object?

PR #1576, opened 2025-02-24
Update rolling upgrade MC version info

wyohm:
MC is not in lock step versions with Platform so it doesn't make sense to say "MC is compatible with the previous minor version". I tried to make it clear and accurate.

There should probably be a link on "Hazelcast Cluster Compatibility" (https://docs.hazelcast.com/management-center/5.7/getting-started/overview#hazelcast-cluster-compatibility), but either I'm not understanding the structure, or the "Preview" function doesn't show it.


wyohm:
After discussing with Rob, this should be a PR to 'main'. Closing this PR and creating a new one on 'main'.

PR #1579, opened 2025-02-26
Compress images

JackPGreen:
Using [ImageOptim](https://imageoptim.com) we can *losslessly* compress images which _should_ make the final docs smaller and therefore load faster.

> Saved 7.9MB out of 39.5 MB. 25.6% per file on average (up to 59%)


pawelklasa:
This will not affect the readability or quality.

Issue #1568, opened 2025-02-21
Docs: Feedback for Supported Versions and Compatibilities


ammad-hzcast:
Hi, I have some feedback about [this page](https://github.com/hazelcast/hz-docs/edit/main/docs/modules/deploy/pages/versioning-compatibility.adoc)
Is there a specific reason why the compatibility matrix for [supported operating system section](https://docs.hazelcast.com/hazelcast/5.5/deploy/versioning-compatibility#supported-operating-systems) uses major.minor.LatestPatch instead of major.minor.X? For example, on the [page for version 5.3](https://docs.hazelcast.com/hazelcast/5.3/deploy/versioning-compatibility#supported-operating-systems) , it lists 5.3.8 as the supported version. If the OS support is valid for all patches, why not simply use 5.3.X?

PR #1582, opened 2025-02-27
Clarify version compatibility by removing specific patch versions

ammad-hzcast:
Updated the supported operating systems section to use major.minor or major.minor.X instead of specific patch versions. This resolves confusion about patch version support and ensures clarity that all patch versions within the major.minor range are supported.

Fixes: https://github.com/hazelcast/hz-docs/issues/1568

### DIFF
--- a/docs/modules/deploy/pages/versioning-compatibility.adoc
+++ b/docs/modules/deploy/pages/versioning-compatibility.adoc
@@ -135,7 +135,7 @@ Hazelcast Platform runs on Java, and supports the following LTS JDKs. It may run
 [options="header"]
 .Supported JDKs
 |===
-|JDK | Platform {full-version}
+|JDK | Platform {minor-version}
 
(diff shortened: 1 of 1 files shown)


JackPGreen:
> `display_version` is defined in antora.yml but it's not resolving in the output, presumably because it's not under `asciidoc:attributes`. You probably want to use `minor-version`. @JackPGreen you've done a lot of work in this area; could you make sure we do the right thing here please?

I'm not sure what the right thing to do here is but `minor-version` seems more sensible to me.

Rob-Hazelcast (approved):
Working now. I've set it to backport to 5.5; the variable wasn't implemented in earlier versions.

JackPGreen (approved):

Rob-Hazelcast:
Related MC PR: https://github.com/hazelcast/management-center-docs/pull/449

PR #1586, opened 2025-03-06
Doc 390

amandalindsay:
Improved Reliable topic description, and added a note to the Topic Statistics section about the known issue with it

### DIFF
--- a/docs/modules/data-structures/pages/reliable-topic.adoc
+++ b/docs/modules/data-structures/pages/reliable-topic.adoc
@@ -15,13 +15,7 @@ data structures, e.g., collection listeners,
 
 Use Reliable Topic when you need a robust, durable messaging system that can handle high throughput without losing messages, especially in scenarios where message delivery guarantees are crucial.
 
-Reliable Topic offers the following metrics:
-
-1. Publish operation count: the total number of messages published to the topic.
-2. Receive operation count: the total number of messages received from the topic.
-3. Creation time: when the topic was created on the member.
-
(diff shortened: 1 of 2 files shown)


docs/modules/data-structures/pages/reliable-topic.adoc
```diff
@@ -1,17 +1,27 @@
...
-very fast producer, it will not lead to problems at topics that run at a slower pace.
+very fast producer, it will not lead to problems at topics that run at a slower pace
 * Since the event system behind a regular `ITopic` is shared with other
 data structures, e.g., collection listeners,
-  you can run into isolation problems. This does not happen with the Reliable `ITopic`.
+  you can run into isolation problems. This does not happen with the Reliable `ITopic`
+
+Use Reliable Topic when you need a robust, durable messaging system that can handle high throughput without losing messages, especially in scenarios where message delivery guarantees are crucial.
+
+Reliable Topic offers the following metrics:
```
emreyigit, line 18: A small note about stats. The stats provided on each Hazelcast member locally. It means there is no global stats about published or received message count in one place. Each instance provides its own local stats.

amandalindsay: OK, I've updated it to simply state "For more information about the metrics provided by Reliable Topic" with a xref to the Topic Stats page

oliverhowell (approved):
Might be one where we don't need the capitalization on Reliable Topic but approving as it's probably better to tackle the sentence case changes in a separate PR

PR #1596, opened 2025-03-10
List dependencies for deb/rpm install

TomaszGaweda:
The process should be simple - user should not wonder what tools are needed.

This PR adds commands to install all required dependencies, so user won't get errors while installing the repositories.

### DIFF
--- a/docs/modules/getting-started/pages/install-hazelcast.adoc
+++ b/docs/modules/getting-started/pages/install-hazelcast.adoc
@@ -51,35 +51,31 @@ Linux::
 +
 --
 
-. Check that you have the GNU Wget package installed. You'll use Wget to download the GPG keys for the Hazelcast package.
-+
-[source,bash]
-----
-wget -V
-----
-+
-If you do not see a version number, download and install link:https://www.gnu.org/software/wget/[Wget^].
-+
+Use either of these package managers to install Hazelcast, depending on your Linux distribution:
 
-. Use either of these package managers to install Hazelcast, depending on your Linux distribution:
-+
 .Debian
 
 [source,shell,subs="attributes+"]
 ----
+# install necessary tools
+sudo apt-get install -y --no-upgrade wget gpg coreutils
+
 wget -qO - https://repository.hazelcast.com/api/gpg/key/public | gpg --dearmor | sudo tee /usr/share/keyrings/hazelcast-archive-keyring.gpg > /dev/null
 echo "deb [signed-by=/usr/share/keyrings/hazelcast-archive-keyring.gpg] https://repository.hazelcast.com/debian stable main" | sudo tee -a /etc/apt/sources.list
 sudo apt update && sudo apt install hazelcast={os-version}
 ----
-+
+
 .RPM
 [source,shell,subs="attributes+"]
 ----
+# install necessary tool
+sudo yum -y install wget
+
 wget https://repository.hazelcast.com/rpm/stable/hazelcast-rpm-stable.repo -O hazelcast-rpm-stable.repo
 sudo mv hazelcast-rpm-stable.repo /etc/yum.repos.d/
 sudo yum install hazelcast-{os-version}
 ----
-+
+
 --
 Windows::
 +
--- a/docs/modules/getting-started/pages/install-hazelcast.adoc
+++ b/docs/modules/getting-started/pages/install-hazelcast.adoc
@@ -1,5 +1,5 @@
 = Installing Hazelcast {open-source-product-name}
-:description: You can install the {open-source-product-name} of Hazelcast, using the command line interface (CLI), Docker, the binary, or use as Java library.
+:description: You can install the {open-source-product-name} of Hazelcast using a package manager, Docker, or the binary. Alternatively, you can add Hazelcast as a dependency in your Java project.
 :page-aliases: installing-using-download-archives.adoc, deploy:running-in-modular-java.adoc
 
 {description}
--- a/docs/modules/getting-started/pages/install-hazelcast.adoc
+++ b/docs/modules/getting-started/pages/install-hazelcast.adoc
@@ -51,7 +51,7 @@ Linux::
 +
 --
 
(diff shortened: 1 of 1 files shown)


JackPGreen (commented):
I like the idea but should these installation instructions be duplicated between `hz-docs` and the packaging repo?

My gut feeling is we should defer to the packaging repo's documentation and add it there instead.

TomaszGaweda:
> My gut feeling is we should defer to the packaging repo's documentation and add it there instead.

If I was a user who wants to install Hazelcast, I'd go to docs first, not search if there is a repo which contains specific packaging flow :)

JackPGreen:
> > My gut feeling is we should defer to the packaging repo's documentation and add it there instead.
> 
> If I was a user who wants to install Hazelcast, I'd go to docs first, not search if there is a repo which contains specific packaging flow :)

We can link to it. Just feel like it's better "owned" by the team responsible for maintaining `packaging`.

TomaszGaweda:
>We can link to it. 

In theory? Yes. In practice it's a bad user experience, when you have to jump between unrelated pages (docs & github repo's markdown). For docker we don't just link to hazelcast-docker, but also say what actually is needed.

JackPGreen:
> For docker we don't just link to hazelcast-docker, but also say what actually is needed.

Good point, you’ve changed my mind :)

…but I’d still like to see a change on the packaging side as well, please?

TomaszGaweda:
@JackPGreen https://github.com/hazelcast/hazelcast-packaging/pull/258 :)

docs/modules/getting-started/pages/install-hazelcast.adoc
```diff
@@ -51,35 +51,31 @@ Linux::
...
 wget -qO - https://repository.hazelcast.com/api/gpg/key/public | gpg --dearmor | sudo tee /usr/share/keyrings/hazelcast-archive-keyring.gpg > /dev/null
 echo "deb [signed-by=/usr/share/keyrings/hazelcast-archive-keyring.gpg] https://repository.hazelcast.com/debian stable main" | sudo tee -a /etc/apt/sources.list
 sudo apt update && sudo apt install hazelcast={os-version}
 ----
-+
+
 .RPM
 [source,shell,subs="attributes+"]
 ----
+# install necessary tool
```
JackPGreen, line 71:
```suggestion
# install necessary tools
```

As on the other one.

TomaszGaweda: RPM requires only wget (one thing)

docs/modules/getting-started/pages/install-hazelcast.adoc
```diff
@@ -51,35 +51,31 @@ Linux::
...
-. Use either of these package managers to install Hazelcast, depending on your Linux distribution:
-+
 .Debian
 
 [source,shell,subs="attributes+"]
 ----
+# install necessary tools
+sudo apt-get install -y --no-upgrade wget gpg coreutils
+
 wget -qO - https://repository.hazelcast.com/api/gpg/key/public | gpg --dearmor | sudo tee /usr/share/keyrings/hazelcast-archive-keyring.gpg > /dev/null
```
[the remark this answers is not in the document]
TomaszGaweda: good point, eb89c4e60810dc55a10569ca65b6b89b4ea208b9

docs/modules/getting-started/pages/install-hazelcast.adoc
JackPGreen: 🤦

docs/modules/getting-started/pages/install-hazelcast.adoc
```diff
@@ -51,35 +51,29 @@ Linux::
...
 ----
-+
+
 .RPM
 [source,shell,subs="attributes+"]
 ----
-wget https://repository.hazelcast.com/rpm/stable/hazelcast-rpm-stable.repo -O hazelcast-rpm-stable.repo
-sudo mv hazelcast-rpm-stable.repo /etc/yum.repos.d/
-sudo yum install hazelcast-{os-version}
+# install necessary tool
```
JackPGreen, line 70: I don't think these comments add value in isolation.

docs/modules/getting-started/pages/install-hazelcast.adoc
```diff
@@ -1,5 +1,5 @@
 = Installing Hazelcast {open-source-product-name}
-:description: You can install the {open-source-product-name} of Hazelcast, using the command line interface (CLI), Docker, the binary, or Java.
+:description: You can install the {open-source-product-name} of Hazelcast, using the command line interface (CLI), Docker, the binary, or use as Java library.
```
Rob-Hazelcast, line 2:
```suggestion
:description: You can install the {open-source-product-name} of Hazelcast using a package manager, Docker, or the binary. Alternatively, you can add Hazelcast as a dependency in your Java project.
```

I don't think "the CLI" is quite right (which CLI?) and the docs otherwise refer to a package manager, and it's hard to get the Java library use case into the same sentence and have it read well.

docs/modules/getting-started/pages/install-hazelcast.adoc
```diff
@@ -51,35 +51,27 @@ Linux::
...
-. Check that you have the GNU Wget package installed. You'll use Wget to download the GPG keys for the Hazelcast package.
-+
-[source,bash]
-----
-wget -V
-----
-+
-If you do not see a version number, download and install link:https://www.gnu.org/software/wget/[Wget^].
-+
+Use either of these package managers to install Hazelcast, depending on your Linux distribution:
```
Rob-Hazelcast, line 54:
```suggestion
You can use either of these package managers to install Hazelcast, depending on your Linux distribution:
```

Rob-Hazelcast (approved):
I've left some suggestions, but otherwise this looks good. Always happy to see UX improvements.

Please add a backport label for 5.5 so users can start taking advantage of this straight away. (You can go back further if you think that's appropriate, but this section looks a little broken in earlier versions so it might not merge cleanly.)

PR #1637, opened 2025-03-31
Clarify WAN queue capacity behaviour [HZG-308]

JamesHazelcast:
We only expose the primary events queue when we talk about "outbound queue size", which makes sense from a practical standpoint, but it can lead to confusion when you encounter scenarios such as partition promotions during migration. This can lead to metrics showing the WAN queue size exceeding the configured capacity, but it's only a shift of existing data, not a population of new data.

Without proper explanation, customers can be confused and concerned that the memory footprint of their WAN queues has suddenly increased, even though this is not the case.

This PR adds more information when discussing queue capacity in WAN, and also adds a bit more detail to the outbound queue metric description.

Fixes https://hazelcast.atlassian.net/browse/HZG-308

### DIFF
--- a/docs/modules/ROOT/pages/list-of-metrics.adoc
+++ b/docs/modules/ROOT/pages/list-of-metrics.adoc
@@ -2701,7 +2701,7 @@ Based on your latency tolerance in your business use case, you can define a thre
 
 |`wan.outboundQueueSize`
 |count
-|Outbound WAN queue size on this member
+|Outbound primary WAN queue size on this member
 
 |`wan.removeCount`
 |count
--- a/docs/modules/ROOT/pages/list-of-metrics.adoc
+++ b/docs/modules/ROOT/pages/list-of-metrics.adoc
@@ -2701,7 +2701,7 @@ Based on your latency tolerance in your business use case, you can define a thre
 
 |`wan.outboundQueueSize`
 |count
-|Outbound primary WAN queue size on this member
+|Total number of WAN events currently placed in the WAN queues of primary partitions on this member
 
 |`wan.removeCount`
 |count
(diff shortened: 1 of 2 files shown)


docs/modules/wan/pages/tuning.adoc
```diff
@@ -181,7 +181,16 @@ For clusters with high data mutation rates or with long expected periods of disr
...
+Separate queues are used for each WAN Replication configured for IMap and ICache.
+
+Two queues are used for each WAN Replication: Primary and Backup. The primary queue is offered events from owned
+partitions, while the backup queue is offered events from partitions owned by other members.
+The configured queue capacity applies separately to each of the primary and backup event queues.
+Metrics for WAN outbound queue sizes report the primary event queue size.
+
+NOTE: During partition migrations, particularly when a cluster size shrinks, some elements in the backup events queue
+can be promoted to the primary events queue, resulting in an increased outbound queue size which can exceed the configured
+queue capacity. The total memory footprint of both queues combined will not increase in this case.
```
Rob-Hazelcast, line 193: This is all good, but can you conclude with an "and therefore no bad things happen" type statement just to be completely clear?

Rob-Hazelcast (approved):

docs/modules/wan/pages/tuning.adoc
JamesHazelcast: Updated here: https://github.com/hazelcast/hz-docs/pull/1637/commits/11993b4a4a6fd21a3b7ba6961448bdc3f2d74f68 is that what you're looking for?

docs/modules/ROOT/pages/list-of-metrics.adoc
```diff
@@ -2701,7 +2701,7 @@ Based on your latency tolerance in your business use case, you can define a thre
 
 |`wan.outboundQueueSize`
 |count
-|Outbound WAN queue size on this member
+|Outbound primary WAN queue size on this member
```
ahmetmircik, line 2704:
```suggestion
|Total number of WAN events currently placed in the WAN queues of primary partitions.
```

ahmetmircik (approved):
one minor comment + LGTM

docs/modules/ROOT/pages/list-of-metrics.adoc
JamesHazelcast: Good suggestion, done, thanks Ahmet

PR #1650, opened 2025-04-08
Bugfix/use correct tools version

fantkolja:


oliverhowell (approved):
Sorry for delay - assuming this is still needed @fantkolja

PR #1658, opened 2025-04-11
CORE-223 - Excluding CP leadership from high latency nodes

Rob-Hazelcast:
https://hazelcast.atlassian.net/browse/CORE-223

Metric and phone home details TBC

Note: no updates to https://docs.hazelcast.com/hazelcast/6.0-snapshot/cp-subsystem/management or non-platform docs.


docs/modules/cp-subsystem/pages/configuration.adoc
```diff
@@ -95,8 +87,7 @@ that should participate in each one, using the <<group-size, `group-size`>> opti
 To scale out throughput and memory capacity, you can choose a CP group size that is
 smaller than the <<cp-member-count, CP member count>> to distribute your CP data structures to multiple CP groups.
 
-CP groups usually consist of an odd number of CP members between three and seven, i.e., it should be either three, five, or seven.
-An odd number of CP members in a group is more advantageous to an even number because of the quorum or majority calculations.
+CP groups must consist of three, five, or seven members. An odd number of CP members avoids stalemate scenarios in leadership elections and quorum calculations.
```
gareth-johnston, line 90: Must? I think it's strongly advisable but not a enforced/validated requirement.

docs/modules/cp-subsystem/pages/configuration.adoc
```diff
@@ -119,13 +110,22 @@ you should xref:cp-subsystem:management.adoc#destroying-a-cp-group-by-force[dest
...
+In some deployment scenarios, you may want to prevent a CP group member from becoming the leader; for example, if your CP group is distributed across multiple data centers and there is a high latency for communications with the remote data center(s). If the CP group leader is in the local data center -- and there are enough followers in the local data center to achieve quorum -- the leader will not need to wait for acknowledgements from followers in the remote data center and so the increased latency will have no impact. If the leader is in the remote data center, there will be a higher latency for client operations and this will reduce the maximum throughput achievable. To maximize throughput, you can prevent CP members in the remote data center from becoming leader by enabling <<auto-step-down-when-leader, 'auto-step-down-when-leader'>> on those members. In this case, the increased latency only has an impact if a remote data center member is needed to achieve quorum.
```
gareth-johnston, line 124: > If the CP group leader is in the local data center -- and there are enough followers in the local data center to achieve quorum -- the leader will not need to wait for acknowledgements from followers in the remote data center and so the increased latency will have no impact.

We want to ensure best practices are maintained in the dialogue here, this suggests that a local **single** DC achieving quorum is fine. It is not, because it impacts fault tolerance. We want to ensure we have replication of the log going across multiple DCs. 

In this example we can use a 3 DC set up. 2 DCs close together, one further away (higher latency). In this scenario, the leader ending up on the slow DC will cause higher latency.

docs/modules/cp-subsystem/pages/configuration.adoc
```diff
@@ -119,13 +110,22 @@ you should xref:cp-subsystem:management.adoc#destroying-a-cp-group-by-force[dest
...
 
-NOTE: Only CP members that are part of a CP group can become leaders. CP members that are not in a CP group do not participate in the leadership rebalancing task. For example, you start a cluster with the CP group size set to three and three CP members with priority 0. Later, you promote a new CP member with priority 100. Even though it has a higher priority, the new CP member never becomes a leader because the CP group is full. The new CP member is eligible for leadership only when a CP member leaves the CP group and the new CP member takes the previous member's place.
+A member must become leader if it has the most recent log in the CP group. In this case, a member with <<auto-step-down-when-leader, 'auto-step-down-when-leader'>> enabled will become leader but immediately trigger the leadership rebalancing task, and will reject client operations. If the leadership rebalancing task fails, it will retry automatically.
```
gareth-johnston, line 126: Might be worth referencing this as a Raft property, not a hazelcast specific impl

> member must become leader if it has the most recent log in the CP group.

docs/modules/cp-subsystem/pages/configuration.adoc
```diff
@@ -119,13 +110,22 @@ you should xref:cp-subsystem:management.adoc#destroying-a-cp-group-by-force[dest
...
+
+Reducing the number of leadership candidates reduces the fault tolerance of the CP Subsystem. You should therefore only enable this feature if it is necessary to achieve your throughput requirements, and you should enable it on as few members as possible. This feature does not apply to the `METADATA` group, which is not sensitive to latency.
```
gareth-johnston, line 128: > , which is not sensitive to latency.

I would add,

which is required for membership operations.

gareth-johnston (commented):
Nice work so far, a few thoughts.

docs/modules/cp-subsystem/pages/configuration.adoc
Rob-Hazelcast: Are you sure? Current docs for `group-size` say "must".

docs/modules/cp-subsystem/pages/configuration.adoc
Rob-Hazelcast: The METADATA group is described in cp-subsystem.adoc. Is it relevant to describe its function here?

docs/modules/ROOT/pages/list-of-metrics.adoc
```diff
@@ -1427,6 +1427,10 @@ This is because the cluster has to communicate with more members, which can add
 |`cp.session.version`
 |count
 |Version number of the CP session, basically it shows how many times the session heartbeat is received
+
+|`cp.session.average.adbdication.time`
```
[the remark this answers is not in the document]
Rob-Hazelcast: I'll leave this as a placeholder for now, but please let me know or update this PR when the new name is decided

docs/modules/cp-subsystem/pages/configuration.adoc
gareth-johnston: I mean that the abdicating nodes still must serve the METADATA group, just an extra justification is nice for the user IMO.

docs/modules/cp-subsystem/pages/configuration.adoc
gareth-johnston: Yes sorry you are correct my bad!

docs/modules/ROOT/pages/list-of-metrics.adoc
```diff
@@ -2163,6 +2164,10 @@ Based on your latency tolerance in your business use case, you can define a thre
 |count
 |Last log term of this CP group
 
+|`raft.group.leadership.averageTimeMs`
+|ms
+|Average time for CP group members to hold the leader role
```
JamesHazelcast, line 2169:
```suggestion
|The average time this CP group member has held the leader role for
```

docs/modules/ROOT/pages/phone-homes.adoc
```diff
@@ -63,6 +63,7 @@ The following information is sent in a phone home:
 ** Number of failed configuration reload calls
 * CP Subsystem specific:
 ** Whether CP Subsystem is enabled
+** Whether auto step down is enabled
```
JamesHazelcast, line 66:
```suggestion
** Whether CP leader auto step down is enabled
```

docs/modules/cp-subsystem/pages/best-practices.adoc
```diff
@@ -1,11 +1,14 @@
...
+* Try to reuse already created or existing xref:cp-subsystem:cp-subsystem.adoc#cp-members[CP objects] (AtomicReference, FencedLock, etc.) and try to avoid creating new ones because they will not be garbage collected.
+
+* Because each CP data structure is bounded by a single xref:cp-subsystem:cp-subsystem.adoc#cp-groups[CP group], a client will create a separate session on each CP group that holds the session-based data structure that the client interacts with. When idle, clients send separate heartbeat messages to each CP group to keep the sessions alive. For instance, if five clients try to acquire four locks, where each lock is placed into a different CP group, then there will be 5 x 4 = 20 heartbeat operations committed in total. However, if all locks are put into the same CP group, there will be only five heartbeats in each period. Therefore, it is strongly recommended that you store all session-based CP data structures (locks and semaphores) in a single CP group to avoid that cost.
 
+* Create the minimum number of xref:cp-subsystem:configuration.adoc#choosing-a-group-size[CP groups] that you think are sufficient for your CP subsystem use case.
```
JamesHazelcast, line 10: What is this advice based on? Most customers only use the `default` CP group, but this actually creates quite a few issues as it becomes a bottleneck. This is discussed in [this internal document](https://hazelcast.atlassian.net/wiki/spaces/EN/pages/5881724929/Snapshotting+Mechanics+in+CP+Subsystem):

> Wherever possible leverage multiple CP groups for data needs, spreading CP data structures among those groups, trying to keep an even distribution of load. This allows for more parallel processing across multiple leader nodes, reduces the size of individual group snapshots, and allows for snapshots to be persisted in parallel.

docs/modules/cp-subsystem/pages/configuration.adoc
```diff
@@ -119,13 +110,26 @@ you should xref:cp-subsystem:management.adoc#destroying-a-cp-group-by-force[dest
...
+
+Only CP members that are part of a CP group can become leaders. CP members that are not in a CP group do not participate in the leadership rebalancing task. For example, you start a cluster with the CP group size set to three and all members with priority 0. Later, you promote a new CP member with priority 100. Even though it has a higher priority, the new CP member never becomes a leader because the CP group is full. The new CP member is eligible for leadership only when a CP member leaves the CP group and the new CP member takes the previous member's place.
+
+IMPORTANT: <<cp-member-priority, `cp-member-priority`>> does not override the Raft leadership election process: it is a hint rather than a rule. Consequently, it is possible that a member with a low priority becomes leader for a period until the background leadership rebalancing task transfers the leadership to a member with a higher priority. The leadership rebalancing task runs approximately every 60 seconds; however, a custom frequency can be defined by the system property `hazelcast.raft.leadership.rebalance.period`. If necessary, you can also <<autostepdown, configure members to reject leadership>>.
```
JamesHazelcast, line 119:
```suggestion
IMPORTANT: <<cp-member-priority, `cp-member-priority`>> does not override the Raft leadership election process: it is a hint rather than a rule. Consequently, it is possible that a member with a low priority becomes leader for a period until the background leadership rebalancing task transfers the leadership to a member with a higher priority. The leadership rebalancing task runs approximately every 60 seconds; however, a custom frequency can be defined by the system property `hazelcast.raft.leadership.rebalance.period`. If necessary, you can also <<autostepdown, configure members to automatically step down from leadership>>.
```

docs/modules/cp-subsystem/pages/configuration.adoc
```diff
@@ -119,13 +110,26 @@ you should xref:cp-subsystem:management.adoc#destroying-a-cp-group-by-force[dest
...
+image:autostepdown-deployment.svg[Distributed CP Group deployment]
+
+If the CP group leader is in one of the local data centers -- and there are enough local followers to achieve quorum -- the leader will not need to wait for acknowledgements from followers in the remote data center and so the increased latency will have no impact. If the leader is in the remote data center, there will be a higher latency for client operations and this will reduce the maximum throughput achievable. To maximize throughput, you can prevent CP members in the remote data center from becoming leader by enabling <<auto-step-down-when-leader, `auto-step-down-when-leader`>> on those members. The increased latency will then only have an impact if a remote data center member is needed to achieve quorum.
```
JamesHazelcast, line 128: We say "you can prevent CP members [...] becoming leader" but this isn't accurate - they do become leader for a short period of time (minimum possible) while they transfer leadership to another node.

docs/modules/cp-subsystem/pages/configuration.adoc
```diff
@@ -119,13 +110,26 @@ you should xref:cp-subsystem:management.adoc#destroying-a-cp-group-by-force[dest
...
+
+In the Raft algorithm, a member must become leader if it has the most recent log in the group. In this case, a member with <<auto-step-down-when-leader, `auto-step-down-when-leader`>> enabled will become leader but immediately trigger the leadership rebalancing task, and will reject client operations. If the leadership rebalancing task fails, it will retry automatically.
```
JamesHazelcast, line 130:
```suggestion
In the Raft algorithm, a member must become leader if it has the most recent log in the group. In this case, a member with <<auto-step-down-when-leader, `auto-step-down-when-leader`>> enabled will become leader but immediately trigger the leadership rebalancing task, and will reject client operations in the interim. If the leadership rebalancing task fails, it will retry automatically.
```

docs/modules/cp-subsystem/pages/configuration.adoc
```diff
@@ -1592,3 +1552,101 @@ config.getCPSubsystemConfig().setRaftAlgorithmConfig(raftConfig);
...
+Java::
++
+--
+[source,java]
+----
+Config config = new Config();
+
+config.getCPSubsystemConfig()
+.setCPMemberPriority(0)
+----
```
JamesHazelcast, line 1605: Does it make sense for our configuration example to use the default value `0`?

docs/modules/cp-subsystem/pages/cp-subsystem.adoc
```diff
@@ -70,7 +59,7 @@ In this mode, the following restrictions apply:
 
 - CP data structures use xref:consistency-and-replication:consistency.adoc[partitioning and lazy
 replication mechanisms] instead of the <<consensus, Raft consensus mechanism>>.
-- CP Subsystem Management APIs are not available.
+- CP Subsystem Management APIs are not available.A Hazelcast member or client that interacts with the CP Subsystem through APIs.
```
JamesHazelcast, line 62: This looks like a mistake? It doesn't seem to make sense here and is missing a space from the start

docs/modules/cp-subsystem/pages/cp-subsystem.adoc
```diff
@@ -104,23 +93,21 @@ To configure the number of CP members that participate in a group, use the xref:
...
 groups compared to other CP members, it can turn into a bottleneck. Therefore, the CP Subsystem runs a periodic background task to try to make sure that each CP
 member is a leader for an equal number of CP groups.
-For example, a cluster has three CP members and three CP groups, each CP member becomes the
-Raft leader for only one CP group. If one more CP group is created, then one of
-the CP members becomes the Raft leader for two CP groups.
+For example, a cluster has three CP members and three CP groups, each CP member becomes the leader for only one CP group. If one more CP group is created, then one of the CP members becomes the leader for two CP groups.
 
-You can also give CP members a priority rating to affect their chances of becoming a leader. See xref:configuration.adoc#configuring-leadership-priority[Configuring Leadership Priority].
+You can give CP members a priority rating to affect their chances of becoming a leader, or configure them to reject leadership if elected. See xref:configuration.adoc#configuring-leadership-priority[Configuring Leadership Priority] and xref:configuration.adoc#autostepdown[Configuring Members to Reject Leadership].
```
JamesHazelcast, line 110: Same as before, not sure we should be saying "rejects" or "prevents" - I'd prefer we said "configure them to immediately transfer leadership if elected" etc.

JamesHazelcast (commented):
Thanks for working on this @Rob-Hazelcast - a few comments from me

docs/modules/cp-subsystem/pages/best-practices.adoc
Rob-Hazelcast: Dunno, it's existing text, not something I've added (see line 8 in this diff).

docs/modules/cp-subsystem/pages/cp-subsystem.adoc
Rob-Hazelcast: I'm not sure I agree. Stopping a node becoming leader is the intent of the feature. And given that the node rejects client operations throughout, it doesn't become a functioning leader, so I think it's a reasonable technical description too. I'd say the fact that it avoids becoming leader by transferring it to another node is an implementation detail, and is explained in the feature description.

docs/modules/cp-subsystem/pages/cp-subsystem.adoc
Rob-Hazelcast: Yes, good catch. It's the glossary definition for "caller". I must have accidentally pasted it here too.

docs/modules/cp-subsystem/pages/configuration.adoc
Rob-Hazelcast: I had the same thought. I might update the existing docs here.

docs/modules/cp-subsystem/pages/configuration.adoc
Rob-Hazelcast: Are they the leader in any meaningful sense if they never do any of the things the leader normally does?

docs/modules/cp-subsystem/pages/cp-subsystem.adoc
gareth-johnston: In terms of the `Raft` algorithm, the semantics @JamesHazelcast suggests are more consistent IMO. Because, any node must be capable of becoming a leader according to 'Rules For Servers'. Not that all our customers will know Raft, but it is important we are clear that whilst it does become a leader, the service will be temporarily unavailable until a higher priority node takes over leadership.

gareth-johnston: It is muddy water though, because a Leader should also be able to replicate operations, where our Leader just rejects them. We can go as far to represent this as a new RaftState, however after considering this I think it overcomplicates things.

Rob-Hazelcast: Let's go with "transfer" then. Thanks!

docs/modules/cp-subsystem/pages/best-practices.adoc
```diff
@@ -1,11 +1,16 @@
...
 
+* Wherever possible, use multiple xref:cp-subsystem:configuration.adoc#choosing-a-group-size[CP groups], spread CP data structures among those groups and try to distribute load evenly. This allows for more parallel processing across multiple leader nodes, reduces the size of individual group snapshots, and allows for snapshots to be persisted in parallel. We recommend creating the same number of CP groups as CP members.
```
JamesHazelcast, line 10:
```suggestion
* For maximum performance, use multiple xref:cp-subsystem:configuration.adoc#choosing-a-group-size[CP groups], spread CP data structures among those groups and try to distribute load evenly. This allows for more parallel processing across multiple leader nodes, reduces the size of individual group snapshots, and allows for snapshots to be persisted in parallel. We recommend creating the same number of CP groups as CP members to maximize throughput.
```

docs/modules/cp-subsystem/pages/configuration.adoc
```diff
@@ -107,25 +98,33 @@ For example, in a CP group of five CP members, operations are committed when the
...
 
-[NOTE]
-====
-We recommend that you create the minimum number of CP groups that you think are sufficient for your CP subsystem use case.
-Creating separate CP groups allows clients to distribute the load among the leaders of several CP groups.
-However, a large number of CP groups running on the same members may slow down the system.
-If you have accidentally or intentionally created a large number of CP groups, and don't have plans to use them anymore,
-you should xref:cp-subsystem:management.adoc#destroying-a-cp-group-by-force[destroy them].
-====
+Creating separate CP groups allows clients to distribute the load among the leaders of several CP groups. However, a large number of CP groups running on the same members may slow down the system. A good compromise is to create the same number of CP groups as CP members. That way, each member will usually be the leader of one group.
```
JamesHazelcast, line 101:
```suggestion
Creating separate CP groups allows clients to distribute the load among the leaders of several CP groups. However, a large number of CP groups running on the same members may slow down the system. A good compromise is to create the same number of CP groups as CP members for high throughput use cases. That way, each member will usually be the leader of one group.
```

JamesHazelcast (approved):
A comment of suggestions w.r.t wording for CP group recommendations, but otherwise LGTM - approving in advance

docs/modules/cp-subsystem/pages/best-practices.adoc
```diff
@@ -1,11 +1,16 @@
...
-* Create the minimum number of xref:cp-subsystem:configuration.adoc#choosing-a-group-size[CP groups] that you think are sufficient for your CP subsystem use case.
+* Try to reuse already created or existing xref:cp-subsystem:cp-subsystem.adoc#cp-data-structures[CP objects] (AtomicReference, FencedLock, etc.) and try to avoid creating new ones because they will not be garbage collected.
 
+* Because each CP data structure is bounded by a single xref:cp-subsystem:cp-subsystem.adoc#cp-groups[CP group], a client will create a separate session on each CP group that holds the session-based data structure that the client interacts with. When idle, clients send separate heartbeat messages to each CP group to keep the sessions alive. For instance, if five clients try to acquire four locks, where each lock is placed into a different CP group, then there will be 5 x 4 = 20 heartbeat operations committed in total. However, if all locks are put into the same CP group, there will be only five heartbeats in each period. Therefore, it is strongly recommended that you store all session-based CP data structures (locks and semaphores) in a single CP group to avoid that cost.
```
amandalindsay, line 8:
```suggestion
* Because each CP data structure is bounded by a single xref:cp-subsystem:cp-subsystem.adoc#cp-groups[CP group], a client will create a separate session on each CP group that holds the session-based data structure that the client interacts with. When idle, clients send separate heartbeat messages to each CP group to keep the sessions alive. For instance, if five clients try to acquire four locks, where each lock is placed into a different CP group, then there will be 5 x 4 = 20 heartbeat operations committed in total. However, if all locks are put into the same CP group, there will be only five heartbeats in each period. Therefore, we strongly recommend that you store all session-based CP data structures (locks and semaphores) in a single CP group to avoid that cost.
```

amandalindsay (approved):
Looks good. Made a small suggestion for consistency which you can action or ignore. Approving either way

PR #1669, opened 2025-04-23
Fix partition counts link

JamesHazelcast:


amandalindsay (approved):

docs/modules/cluster-performance/pages/performance-tips.adoc
```diff
@@ -640,7 +640,7 @@ and effort on meticulous and fragile GC tuning.
 
 If you use HDMS with large data sizes,
 we recommend a large increase in partition count, starting with 5009 or higher. See the
-<<partition-count, Partition Count section>> above for more information. Also, if you intend
+<<partitions, Partitions section>> above for more information. Also, if you intend
```
Rob-Hazelcast, line 643:
```suggestion
<<partitions, Partitions section>> for more information. Also, if you intend
```
It's below! But we can just avoid specifying.

Rob-Hazelcast (approved):
Thanks for fixing. I'll add the forwardport label so this gets into main too.

Issue #1091, opened 2024-04-22
Outdated `5.3` XML schema definition used


JackPGreen:
[This page](https://docs.hazelcast.com/hazelcast/latest/data-connections/data-connections-configuration#Hazelcast) uses the `5.3` XML schema even though `5.4` has been released. 

More specifically, the page uses the [`minor-version` property from `antora.yml`](https://github.com/hazelcast/hz-docs/blob/main/docs/antora.yml#L14C5-L14C18), which has not been updated when the version was released in cdc7cd1c464479d1e9593dfe950a4ee7804b69ca.

PR #1677, opened 2025-05-01
Fix config XML / schema outdated versions by removing `minor-version` property

JackPGreen:
The `minor-version` property ~~is~~ was outdated and wasn't being automatically maintained, meaning references that used it were referencing old versions.

Rather that automatically increasing this version, we can simply instead reference an alternative variable that [_is_ being automatically maintained](https://github.com/hazelcast/hazelcast-pipeline-library/blob/766f1d2e302d2bb2aeb7ffe7affb0a35a37743f3/src/com/hazelcast/qe/pipeline/distribution/layout/reference/manual/UpdateHzDocs.groovy#L127).

Fixes https://github.com/hazelcast/hz-docs/issues/1091

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -1,6 +1,6 @@
 name: hazelcast
 title: Hazelcast Platform
-# Version in the URL
+# Version in the URL (minor.patch version)
 version: '6.0-snapshot'
 # Version in the version selector (we display only the latest major.minor version)
 display_version: '6.0-SNAPSHOT'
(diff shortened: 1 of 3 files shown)


docs/antora.yml
```diff
@@ -13,8 +13,8 @@ asciidoc:
     os-version: '6.0.0-SNAPSHOT'
     ee-version: '6.0.0-SNAPSHOT'
     jet-version: '4.5.4'
-    # The minor.patch version, which is used as a variable in the docs for things like file versions
-    minor-version: '6.0-SNAPSHOT'
+    # The outer "version" property duplicated to allow access from inside asciidoc
```
Rob-Hazelcast, line 16:
```suggestion
    # Same as `version` but as an asciidoc attribute.
```

docs/antora.yml
```diff
@@ -13,8 +13,8 @@ asciidoc:
     os-version: '6.0.0-SNAPSHOT'
     ee-version: '6.0.0-SNAPSHOT'
     jet-version: '4.5.4'
-    # The minor.patch version, which is used as a variable in the docs for things like file versions
-    minor-version: '6.0-SNAPSHOT'
+    # Same as `version` but as an asciidoc attribute.
```
Rob-Hazelcast, line 16:
```suggestion
    # Same as `version` above but as an asciidoc attribute
```

Rob-Hazelcast (approved):

PR #1688, opened 2025-05-05
Add warning not to rely on exact vector metric value

k-jamroz:


docs/modules/data-structures/pages/vector-collections.adoc
```diff
@@ -136,6 +136,8 @@ When disabled, each added vector is treated as a distinct vector in the index, e
 TIP: The recommended method for computing cosine similarity is to normalize all vectors to unit length and use the DOT metric instead.
 The COSINE metric can be significantly slower than the DOT metric, especially when the xref:jvm-configuration[Vector API] is not enabled.
 
+WARNING: You should not rely on exact value of the metric.
```
amandalindsay, line 139:
```suggestion
WARNING: Do not rely on the exact value of the metric. 
```

docs/modules/data-structures/pages/vector-collections.adoc
```diff
@@ -136,6 +136,8 @@ When disabled, each added vector is treated as a distinct vector in the index, e
 TIP: The recommended method for computing cosine similarity is to normalize all vectors to unit length and use the DOT metric instead.
 The COSINE metric can be significantly slower than the DOT metric, especially when the xref:jvm-configuration[Vector API] is not enabled.
 
+WARNING: You should not rely on exact value of the metric. 
+It is a floating point number and the exact value may depend on the order of operations and rounding during evaluation which can be different in different implementations (with or without vectorization) and in different Hazelcast versions.
```
amandalindsay, line 140:
```suggestion
It is a floating point number and the exact value may depend on the order of operations and rounding during evaluation which can vary in different implementations (with or without vectorization) and in different Hazelcast versions.
```

amandalindsay (approved):
Approved with minor tweaks

gbarnett-hz (approved):

PR #1710, opened 2025-05-20
Update cp-subsystem.adoc note about enterprise being a requirement

wyohm:
All of CP-Subsystem is now Enterprise only, so saying that CP Persistence being a requirement makes this enterprise only is redundant and confusing.


Rob-Hazelcast (approved):
I've got an open PR that includes this fix, but might as well get this in for now. Thanks!

PR #1734, opened 2025-06-20
Remove Gradle examples

JackPGreen:
[_Some_ examples include example Maven and Gradle configuration samples](https://docs.hazelcast.com/hazelcast/latest/test/testing), but most [don't](https://docs.hazelcast.com/hazelcast/latest/security/integrating-openssl#using-boringssl).

Moreover:
- we don't provide any instructions for using Hazelcast with Gradle, [only Maven](https://docs.hazelcast.com/hazelcast/latest/getting-started/install-hazelcast#using-maven)
- we aren't explicitly testing against Gradle

As such I don't see the value in these examples and they should be removed.

### DIFF
--- a/docs/modules/pipelines/pages/kinesis.adoc
+++ b/docs/modules/pipelines/pages/kinesis.adoc
@@ -38,35 +38,6 @@ NOTE: Keep in mind that you are setting up a stream called `Tweets` instead of `
 . To check that everything is set up correctly, link:https://docs.aws.amazon.com/streams/latest/dev/kinesis-tutorial-cli-installation.html[install the AWS CLI]
 and link:https://docs.aws.amazon.com/streams/latest/dev/fundamental-stream.html[perform some basic operations with it].
 
-== Step 2. Create a New Java Project
-
-We'll assume you're using an IDE. Create a blank Java project named
-`kinesis-tutorial` and copy the Gradle or Maven file into it:
-
-[tabs] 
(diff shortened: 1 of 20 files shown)


docs/modules/pipelines/pages/kinesis.adoc
```diff
@@ -38,35 +38,6 @@ NOTE: Keep in mind that you are setting up a stream called `Tweets` instead of `
 . To check that everything is set up correctly, link:https://docs.aws.amazon.com/streams/latest/dev/kinesis-tutorial-cli-installation.html[install the AWS CLI]
 and link:https://docs.aws.amazon.com/streams/latest/dev/fundamental-stream.html[perform some basic operations with it].
 
-== Step 2. Create a New Java Project
```
Rob-Hazelcast, line 41: Unfortunately this tutorial has hardcoded step numbers, so you need to update the remaining steps

Rob-Hazelcast (approved):
LGTM, just a couple of small fixes. Thanks!

docs/modules/pipelines/pages/kinesis.adoc
JackPGreen: I didn't mean to delete that anyway - https://github.com/hazelcast/hz-docs/pull/1734/commits/fc71cb7dba7ed579a80a7d61b017a24d611f07c1

PR #1735, opened 2025-06-24
Update `cluster-utilities`

JackPGreen:
This documentation seems to be outdated, predating when the command line tool was integrated into platform / MC respectively in https://github.com/hazelcast/hazelcast/pull/20262

Specifically:
- removed references to Management Center as the platform tool does not support that
- Removed Java 8 pre-requisite reference
   - if you try and run with Java 8, it just says  `Unrecognized option: --add-modules Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.`
- removed references on _how_ to install it
   - duplicated elsewhere
   - in the following section we refer to `Hazelcast package` and _assume_ users will know how to obtain that
- removed links to the _old_ home of the `hazelcast-command-line` tool

### DIFF
--- a/docs/modules/management/pages/cluster-utilities.adoc
+++ b/docs/modules/management/pages/cluster-utilities.adoc
@@ -7,93 +7,23 @@ to check whether the cluster and/or members are safe before shutting down a memb
 to define the minimum number of cluster members required for the cluster to remain up and running.
 It also gives information about the Hazelcast Lite Member.
 
+
+These tools are included with the Hazelcast package.
+
 == Hazelcast Command Line Tool
 
-This is a tool using which you can install and run Hazelcast and Management Center
-on your Unix-like local environments. You need https://www.oracle.com/java/technologies/javase-downloads.html[JRE 8 or newer^]
-as a prerequisite.
-
-You can install this tool using its Homebrew, RPM, or Debian packages:
-
-[tabs] 
-==== 
-Homebrew:: 
-+ 
--- 
-
-[source,bash,subs="attributes+"]
-----
-brew tap hazelcast/hz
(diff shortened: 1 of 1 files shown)


docs/modules/management/pages/cluster-utilities.adoc
```diff
@@ -7,93 +7,23 @@ to check whether the cluster and/or members are safe before shutting down a memb
...
+== Hazelcast Command Line Tool
 
-You are now ready to use the tool.
+This is a tool using which you can install and run Hazelcast 
+on your Unix-like local environments.
 
 * **Starting a standalone Hazelcast member with the default configuration:**
 +
 `./hz start`
 *
```
Rob-Hazelcast, line 21:
```suggestion
```

docs/modules/management/pages/cluster-utilities.adoc
```diff
@@ -7,93 +7,23 @@ to check whether the cluster and/or members are safe before shutting down a memb
...
-cd hazelcast-<version>/bin
-./hz
-```
+== Hazelcast Command Line Tool
 
-You are now ready to use the tool.
+This is a tool using which you can install and run Hazelcast 
+on your Unix-like local environments.
 
 * **Starting a standalone Hazelcast member with the default configuration:**
```
Rob-Hazelcast, line 18:
```suggestion
Start a standalone Hazelcast member with the default configuration:
```

Rob-Hazelcast (approved):
OK as far as it goes, but would be good to explain how to install the tools (a link is fine if this is documented elsewhere), where to run commands from, and any other requirements (I think you need the REST API enabled?).

PR #1740, opened 2025-06-27
[HZ-4796] Docs: Allow clients without a matching cluster name to connect with the membersadd skip namecheck property

Rob-Hazelcast:
https://hazelcast.atlassian.net/browse/HZ-4796

### DIFF
--- a/docs/modules/ROOT/pages/system-properties.adoc
+++ b/docs/modules/ROOT/pages/system-properties.adoc
@@ -80,7 +80,7 @@ it cleans up the local resources that are owned by that client.
 |`hazelcast.client.internal.skip.cluster.namecheck.during.connection`
 |false
 |bool
-|Allows clients to connect without matching the cluster name.
+|Allows clients to connect without matching the cluster name. This can help if you have a large number of existing clients and it would be impractical to update them. When enabled, cluster name validation is skipped during client authentication.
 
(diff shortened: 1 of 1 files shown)


oliverhowell (approved):
LGTM @Rob-Hazelcast though not sure about backport label - worth double checking exactly which versions it was ported to

JackPGreen:
Shouldn't the release notes have the same (or less) detail as the main docs? I feel like the usecase (_"ease upgrades in environments with large numbers of existing clients where updating configurations is impractical"_) helps conceptualize.

JamesHazelcast:
IIRC, this property was added for a specific customer's use case - and I believe Avtar intentionally did not add this property to documentation (not all properties are present) because we don't want to encourage its use.

Rob-Hazelcast:
Is that the only possible use case? And is it typical? It isn't obvious to me why an upgrade would necessitate updating the cluster name, unless e.g. it became a requirement in a more recent release and wasn't previously configured.

Rob-Hazelcast:
I've added some more info without mentioning upgrade specifically. What do you think? @JackPGreen @JamesHazelcast.

I don't think this adds all that much and I was conscious of brevity with this being a table entry in a reference doc, but it doesn't add _nothing_.

PR #1746, opened 2025-07-03
Fix broken links

amandalindsay:
Corrected the links to point to the appropriate section that discusses clustered Rest and JMX


oliverhowell (approved):
Thanks @amandalindsay make sure you add appropriate backport labels for this one too

amandalindsay:
Note that this fix should work for v5.5 (which I've backported to). Earlier versions have the 'clustered-jmx-and-rest-via-management-center' section

Rob-Hazelcast:
I'm confused. The original links are present back to 5.1, and don't appear to work properly for that or any other version.

The new links both go to the same top-level section in the platform docs, which don't talk about those features except to link to the MC home page. This is not very helpful!

A far more useful fix would be to replace the links here with links to the appropriate sections of the MC docs that describe the two features:

https://docs.hazelcast.com/management-center/5.8/integrate/clustered-rest
https://docs.hazelcast.com/management-center/5.8/integrate/jmx

amandalindsay:
I think it's fine to link to the Monitoring/MC section for context, from where the reader can jump to the JMX or Rest pages.
I do not know why the earlier versions which have the correct section do not work - that is a mystery

PR #1769, opened 2025-07-16
Update support.adoc with current severity levels

wyohm:
Update severity levels to match current policy.


wyohm:
@oliverhowell Agreed, it could use a more in-depth review at some point.
I took your suggestions.

PR #1772, opened 2025-07-18
ci: add typo check step to validate workflow and its configuration to prevent false positive

atilsensalduz:
This PR adds a typo check to CI using the typos GitHub Action. I also created a .typos.toml config file so we can ignore certain words that are either project specific, technical terms, or intentionally spelled that way. Without this config, the action would flag a lot of false positives, so this should make the checks more useful and less noisy.

If you notice any words in the whitelist that actually are typos and should be fixed, feel free to update the config or let me know. 

https://github.com/crate-ci/typos

image


docs/modules/cp-subsystem/pages/cp-subsystem.adoc
```diff
@@ -169,7 +169,7 @@ The method by which deployment, pause, resume and rolling update are performed w
...
 ====
 
-We recommend setting xref:cp-subsystem:configuration.adoc#data-load-timeout-seconds[data-load-timeout-seconds,window=_blank] to a value that spans the duration from when the first pod is running to when the last pod is running and has completed its CP intialisation procedure. This is particularly important if you intend to perform _resume_ scenarios. The only way to determine when a CP member has completed its initialisation is to consult the logs. Therefore, we recommend the following to determine a reasonable value for `data-load-timeout-seconds`:
+We recommend setting xref:cp-subsystem:configuration.adoc#data-load-timeout-seconds[data-load-timeout-seconds,window=_blank] to a value that spans the duration from when the first pod is running to when the last pod is running and has completed its CP initialisation procedure. This is particularly important if you intend to perform _resume_ scenarios. The only way to determine when a CP member has completed its initialization is to consult the logs. Therefore, we recommend the following to determine a reasonable value for `data-load-timeout-seconds`:
```
Rob-Hazelcast, line 172:
```suggestion
We recommend setting xref:cp-subsystem:configuration.adoc#data-load-timeout-seconds[data-load-timeout-seconds,window=_blank] to a value that spans the duration from when the first pod is running to when the last pod is running and has completed its CP initialization procedure. This is particularly important if you intend to perform _resume_ scenarios. The only way to determine when a CP member has completed its initialization is to consult the logs. Therefore, we recommend the following to determine a reasonable value for `data-load-timeout-seconds`:
```

PR #1787, opened 2025-07-25
Update java.adoc client connection strategy

wyohm:
Clarify that when `async-start` = false, the client waits for all cluster members (defined by routing mode) to be connected. This comes from a discussion with İhsan in SUP-972.


Rob-Hazelcast (approved):
LGTM, thanks!

PR #1804, opened 2025-08-01
Fix `antora` variables in code blocks post #1800

JackPGreen:
@Rob-Hazelcast [noticed](https://github.com/hazelcast/hz-docs/pull/1800#pullrequestreview-3079045252) that some of the variables weren't resolving in code blocks due not being expanded.


Rob-Hazelcast (approved):
Thanks!

PR #1811, opened 2025-08-06
1785-java-docs-maven: add Maven dependency

amandalindsay:
Added the Maven repo dependency to the java docs topic, as agreed as best strategy
Also deleted the empty note

### DIFF
--- a/docs/modules/clients/pages/java.adoc
+++ b/docs/modules/clients/pages/java.adoc
@@ -24,7 +24,7 @@ If you prefer to use Maven, make sure you have added the appropriate `hazelcast`
 {enterprise-product-name}:: 
 + 
 --
-Add the `hazelcast-enterprise` dependencies to your `pom.xml`:
+Add the `hazelcast-enterprise` dependency to your `pom.xml`:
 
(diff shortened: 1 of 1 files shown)


docs/modules/clients/pages/java.adoc
```diff
@@ -24,17 +24,33 @@ If you prefer to use Maven, make sure you have added the appropriate `hazelcast`
 {enterprise-product-name}:: 
 + 
 --
-Add the `hazelcast-enterprise` dependency to your `pom.xml`:
+Add the `hazelcast-enterprise` dependencies to your `pom.xml`:
```
JackPGreen, line 27:
```suggestion
Add the `hazelcast-enterprise` dependency to your `pom.xml`:
```

There's only a single dependency.

JackPGreen (approved):

JackPGreen:
> I think we should explain this additional config - it's not obvious why you'd need it. 

I don’t disagree.


> Also applies in getting-started/install-enterprise.


But because of this I don’t think it directly affects this PR.

Do you want to try to draft something up and I review?

Issue #1785, opened 2025-07-25
Docs: Feedback for Java Client


JackPGreen:
Hi, I have some feedback about [this page](https://github.com/hazelcast/hz-docs/edit/main/docs/modules/clients/pages/java.adoc)

The `Enterprise Edition` instructions contain the Maven co-ordinates, but _don't_ include the required Maven repository as specified in https://docs.hazelcast.com/hazelcast/latest/getting-started/install-enterprise#using-enterprise-edition-java

Without the repository being specified, Maven cannot resolve the artifact:
```
% mvn dependency:get -Dartifact=com.hazelcast:hazelcast-enterprise:5.5.7
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.7.0:get (default-cli) on project standalone-pom: Couldn't download artifact: org.eclipse.aether.resolution.DependencyResolutionException: The following artifacts could not be resolved: com.hazelcast:hazelcast-enterprise:jar:5.5.7 (absent): Could not find artifact com.hazelcast:hazelcast-enterprise:jar:5.5.7 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
```

I think it would probably be better just linking directly to the `Getting Started` page.

PR #1815, opened 2025-08-06
hz-docs issue 1785 - annotate Maven install config

Rob-Hazelcast:
Follow on from https://github.com/hazelcast/hz-docs/pull/1811 to explain the config.

Community Edition doesn't require specifying our repo because it's hosted on Maven Central, but I don't think we need to explain that difference between the two editions because we prompt users to choose one or the other.

Not an entirely accurate use of in-line annotations but seemed the neatest way to do it. Switched to includes to avoid duplication in source.

https://github.com/hazelcast/hz-docs/issues/1785

### DIFF
--- a/docs/modules/getting-started/pages/install-enterprise.adoc
+++ b/docs/modules/getting-started/pages/install-enterprise.adoc
@@ -86,7 +86,7 @@ ifdef::snapshot[]
     <repository>
         <id>private-repository</id>
         <name>Hazelcast Private Repository</name>
-        <url>https://repository.hazelcast.com/snapshot/</url>
+        <url>https://repository.hazelcast.com/snapshot/</url> <1>
         <releases>
             <enabled>false</enabled>
         </releases>
@@ -99,11 +99,13 @@ ifdef::snapshot[]
 <dependencies>
     <dependency>
         <groupId>com.hazelcast</groupId>
-        <artifactId>hazelcast-enterprise</artifactId>
+        <artifactId>hazelcast-enterprise</artifactId> <2>
         <version>{ee-version}</version>
     </dependency>
 </dependencies>
 ----
+<1> Specifies the location for Maven to access Hazelcast artifacts.
+<2> Specifies the Hazelcast {enterprise-product-name} dependency and version.
 endif::[]
 ifndef::snapshot[]
 [source,xml,subs="attributes+"]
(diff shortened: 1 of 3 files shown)


JackPGreen:
> Switched to includes to avoid duplication in source.

❤️

docs/modules/getting-started/pages/install-enterprise.adoc
```diff
@@ -99,11 +99,13 @@ ifdef::snapshot[]
...
     <dependency>
         <groupId>com.hazelcast</groupId>
-        <artifactId>hazelcast-enterprise</artifactId>
+        <artifactId>hazelcast-enterprise</artifactId> <2>
         <version>{ee-version}</version>
     </dependency>
 </dependencies>
 ----
+<1> Specifies the location for Maven to access Hazelcast artifacts.
+<2> Specifies the Hazelcast {enterprise-product-name} dependency and version.
```
JackPGreen, line 108: Nit - actually just specifies `artifactId` - `version` is on the next line.

docs/modules/getting-started/pages/install-enterprise.adoc
```diff
@@ -125,11 +127,13 @@ ifndef::snapshot[]
...
```
JackPGreen, line 136: Could we move these outside of the `ifndef::snapshot[]` to avoid duplication? It's the same comment either way.

docs/modules/getting-started/pages/install-enterprise.adoc
```diff
@@ -99,11 +99,13 @@ ifdef::snapshot[]
 <dependencies>
     <dependency>
         <groupId>com.hazelcast</groupId>
-        <artifactId>hazelcast-enterprise</artifactId>
+        <artifactId>hazelcast-enterprise</artifactId> <2>
         <version>{ee-version}</version>
     </dependency>
 </dependencies>
 ----
+<1> Specifies the location for Maven to access Hazelcast artifacts.
```
JackPGreen, line 107: This definition seems a bit derivative.

|    `id`    |                        private-repository                       |
|:----------:|:---------------------------------------------------------------:|
| `name`     | Hazelcast Private Repository                                    |
| Annotation | Specifies the location for Maven to access Hazelcast artifacts. |

Could we update the `Name` to make it clearer and forgo the description?

It's _also_ documented in https://github.com/hazelcast/hazelcast-mono#maven-repository and perhaps that name is better.

docs/modules/getting-started/pages/install-enterprise.adoc
JackPGreen, line 1: Not directly related to your change but I think we can drop `enabled` for `releases` and `snapshots` in their respective repos as I assume it's the default based on this - https://github.com/hazelcast/hazelcast-code-samples/blob/cd159d7718a20cb63b3bc7ec9a022dc8a17e94be/enterprise/pom.xml#L49-L64

JackPGreen (approved):

amandalindsay (approved):

PR #1823, opened 2025-08-14
Update client-authorization.adoc

erkangurgen:
Management Permission Config (in YAML format) should be JSONArray


Rob-Hazelcast (approved):
Thanks for fixing. If you haven't already, could you please check the other YAML examples in this section for the same error?

erkangurgen:
Other YAML configs look fine, I do not see an issue.

PR #1829, opened 2025-08-15
DOC-467 - add restart procedure

Rob-Hazelcast:
Restarts are routinely required for updating static config but there's no complete procedure in the docs for performing a restart.

TODO:

- [x] Add incoming links from config sections
- [ ] Maybe add K8s instructions - Helm/Operator?

TODO (future PRs):

- [ ] Improve partial procedures in migrate/rolling-restart and maintain-cluster/rolling-upgrades

### DIFF
--- a/docs/modules/maintain-cluster/pages/restart-cluster.adoc
+++ b/docs/modules/maintain-cluster/pages/restart-cluster.adoc
@@ -36,7 +36,7 @@ This section describes how to restart a cluster using a choice of common tools. 
 
 You should perform restarts in a low utilization or maintenance window to reduce the risk of service impacts.
 
-include::partial$rest-deprecation.adoc[]
+include::clients:partial$rest-deprecation.adoc[]
 
 [[rolling]]
 === Perform a rolling restart


docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
++
+--
+[source,bash]
+----
+docker start <container name>
+----
+--
+====
+
+. Wait until the new member joins the cluster and all partitions are completed. Again, you can monitor progress using the xref:{page-latest-supported-mc}@management-center:clusters:dashboard.adoc[cluster dashboard] in Management Center.
```
JackPGreen, line 114:
```suggestion
. Wait until the new member joins the cluster and all partitions migrations are completed. Again, you can monitor progress using the xref:{page-latest-supported-mc}@management-center:clusters:dashboard.adoc[cluster dashboard] in Management Center.
```

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+Some configuration needs to be consistent across all members in a cluster. To update this configuration, you must restart the whole cluster simultaneously. This causes all data in the cluster to be lost. To avoid data loss, you must enable xref:storage:persistence.adoc[persistence]. To avoid downtime, you must divert traffic to another Hazelcast cluster before restarting.
+
+Note that members are not restarted. In all cases, members are destroyed, and new members are created to replace them.
+
+You should perform restarts in a low utilization or maintenance window to reduce the risk of service impacts.
+
+== Persistence
+
+If xref:storage:persistence.adoc[persistence] is enabled, replacement members will adopt the UUIDs of the shut down members and attempt to load persisted data from disk. This avoids the repartitioning process, helping the cluster recover more quickly during a rolling restart.
```
JackPGreen, line 25:
```suggestion
If xref:storage:persistence.adoc[persistence] is enabled, replacement members will adopt the UUIDs of the destroyed members and attempt to load persisted data from disk. This avoids the repartitioning process, helping the cluster recover more quickly during a rolling restart.
```

Based on the earlier statement:
> In all cases, members are destroyed

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
+:description: Static configuration is configuration that cannot be changed while members are running. To apply updates to static configuration, you must restart the cluster.
+
+{description}
+
+There are two options for restarting a cluster:
+
+* <<rolling>>
+* <<whole>>
+
+Where possible, you should perform a rolling restart. In a rolling restart, members are stopped one at a time and each member's data is automatically repartitioned across the remaining cluster members before shutting down. This minimizes the risk of service interruptions and data loss. When the replacement member joins the cluster, data is partitioned across the cluster again.
```
JackPGreen, line 11: Nit - partitioned, or repartitioned? we use both.

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+Depending on the scope of your configuration changes, you may need update a single member, a cluster, or multiple clusters. The process is the same in all cases.
+
+TIP: You can also use the rolling restart process for canarying: testing changes on a single member before updating the rest of the cluster.
+
+Some configuration needs to be consistent across all members in a cluster. To update this configuration, you must restart the whole cluster simultaneously. This causes all data in the cluster to be lost. To avoid data loss, you must enable xref:storage:persistence.adoc[persistence]. To avoid downtime, you must divert traffic to another Hazelcast cluster before restarting.
+
+Note that members are not restarted. In all cases, members are destroyed, and new members are created to replace them.
+
+You should perform restarts in a low utilization or maintenance window to reduce the risk of service impacts.
```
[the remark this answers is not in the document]
Rob-Hazelcast: The details are up to customers but it's always worth highlighting when something is potentially service impacting.

docs/modules/maintain-cluster/pages/restart-cluster.adoc
Rob-Hazelcast: Yeah, this is a bit of a cop out. I didn't want to say "repartitioned again" because it looks tautologous.

docs/modules/configuration/pages/understanding-configuration.adoc
```diff
@@ -26,7 +26,7 @@ The following topics are also relevant to static configuration:
 NOTE: Hazelcast performs schema validation through the `hazelcast-config-{version}.xsd` file,
 which comes with Hazelcast libraries. If an error occurs in declarative or programmatic configuration, Hazelcast throws a meaningful exception.
 
-Static configuration cannot be changed at runtime. However, you can add <<dynamic-configuration, dynamic configuration>> for some features.
+Static configuration cannot be changed at runtime. To update static configuration, you must restart the cluster. However, you can add <<dynamic-configuration, dynamic configuration>> for some features.
```
k-jamroz, line 29: not every part of static configuration can be updated (eg. partition count, probably I could find more examples)

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+Where possible, you should perform a rolling restart. In a rolling restart, members are stopped one at a time and each member's data is automatically repartitioned across the remaining cluster members before shutting down. This minimizes the risk of service interruptions and data loss. When the replacement member joins the cluster, data is partitioned across the cluster again.
+
+Depending on the scope of your configuration changes, you may need update a single member, a cluster, or multiple clusters. The process is the same in all cases.
+
+TIP: You can also use the rolling restart process for canarying: testing changes on a single member before updating the rest of the cluster.
+
+Some configuration needs to be consistent across all members in a cluster. To update this configuration, you must restart the whole cluster simultaneously. This causes all data in the cluster to be lost. To avoid data loss, you must enable xref:storage:persistence.adoc[persistence]. To avoid downtime, you must divert traffic to another Hazelcast cluster before restarting.
+
+Note that members are not restarted. In all cases, members are destroyed, and new members are created to replace them.
```
k-jamroz, line 19: I have an impression that "destroyed" is not frequently used in context of members. Members are shutdown, terminated or leave the cluster.

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
+
+You should perform restarts in a low utilization or maintenance window to reduce the risk of service impacts.
+
+== Persistence
+
+If xref:storage:persistence.adoc[persistence] is enabled, replacement members will adopt the UUIDs of the destroyed members and attempt to load persisted data from disk. This avoids the repartitioning process, helping the cluster recover more quickly during a rolling restart.
```
k-jamroz, line 25: > This avoids the repartitioning process

only if you switch the cluster to `NO_MIGRATION` or `FROZEN` state (eg. https://docs.hazelcast.com/hazelcast/5.5/kubernetes/kubernetes-persistence#configuration)

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
+--
++
+You can specify the IP address of any member in the cluster.
+
+Java::
++
+--
+[source,java]
+----
+HazelcastInstance.getCluster().shutdown()
```
k-jamroz, line 151: note that this does not work from client

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
+--
+[source,bash]
+----
+bin/hz start
+----
+--
+
+Java::
++
+Call `newHazelcastInstance` to create each new member as required. They will form a cluster automatically. The following example creates a new three member cluster.
```
k-jamroz, line 192: the cluster will be running embedded in current process - likely not suitable for production

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
+= Restart a cluster
+:description: Static configuration is configuration that cannot be changed while members are running. To apply updates to static configuration, you must restart the cluster.
+
+{description}
+
+There are two options for restarting a cluster:
```
k-jamroz, line 6: this page is somewhat related to https://docs.hazelcast.com/hazelcast/5.5/maintain-cluster/shutdown - should it be linked?

docs/modules/ROOT/nav.adoc
```diff
@@ -365,6 +365,7 @@
 * xref:configuration:jet-configuration.adoc[]
 // Maintaining a cluster
 * Maintaining a cluster
+** xref:maintain-cluster:restart-cluster.adoc[]
```
k-jamroz, line 368: https://docs.hazelcast.com/hazelcast/5.5/maintain-cluster/shutdown is further down in the nav which is a bit awkward. maybe `shutting down` should be first? Or better: 1. starting cluster 2. shutting down 3. restarting?

oliverhowell: Good point - worth considering the nav order for this section as part of this PR?

docs/modules/maintain-cluster/pages/restart-cluster.adoc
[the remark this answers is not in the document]
oliverhowell: This page is good, better than the rolling restart one linked above - either as part of this PR or as a quick follow-up, I'd recommend reviewing and editing rolling-restart.

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
+Hazelcast.newHazelcastInstance()
+----
+--
+
+Docker::
++
+--
+[source,bash]
+----
+docker start <container name>
```
oliverhowell, line 109: I don't know if we have a set standard for placeholders but I've been favouring underscores to keep it unbroken e.g. <your_license_key> - could we agree a standard on this and keep it?

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,219 @@
...
+[tabs]
+====
+CLI::
++
+IMPORTANT: This command shuts down all running members on the local device.
++
+--
+[source,bash]
+----
+bin/hz-stop
```
oliverhowell, line 51: Is this correct in all cases?
I can run `hz-stop` but not `bin/hz-stop`...and I get `No Hazelcast server found to stop`

Could just be my setup or misunderstanding...

oliverhowell (approved):
This is a really good addition - some minor comments but I think once you address/resolve @k-jamroz comments we should get this merged and backported!

docs/modules/maintain-cluster/pages/restart-cluster.adoc
Rob-Hazelcast: Interesting. Other docs with this command specify /bin. if If you create a new member and access its terminal, you're placed at `/opt/hazelcast`, and `/bin` is a child of that directory, so it's sensible. But it looks like it's not required and hz-stop runs from anywhere, so better to omit the path. Thanks for testing.

However, if you try and run it with /bin from another directory it should just fail to find the script. Not sure how you end up with `No server found`.

docs/modules/maintain-cluster/pages/restart-cluster.adoc
Rob-Hazelcast: I don't think so - the idea is to provide a self-contained procedure for a routine operation (restarting the cluster). Unless there's any useful info in that shutdown topic that you think should be included or linked to here?

docs/modules/maintain-cluster/pages/restart-cluster.adoc
Rob-Hazelcast: @k-jamroz Should we provide instructions for shutting down / starting from a client? (And if so, can you please provide them?)

docs/modules/maintain-cluster/pages/restart-cluster.adoc
Rob-Hazelcast: @k-jamroz Very good point, we want this to work for production deployments. Can you provide instructions for client/server?

docs/modules/configuration/pages/understanding-configuration.adoc
Rob-Hazelcast: I'm avoiding the question of "how do you know which config needs a whole cluster restart vs a rolling restart" here. Such requirements should be documented in the relevant feature/config sections.

docs/modules/maintain-cluster/pages/restart-cluster.adoc
k-jamroz: you cannot start a (standalone=non-embedded) member using client. in order to start a member you need to start a new process on the server

docs/modules/maintain-cluster/pages/restart-cluster.adoc
k-jamroz: Starting an instance is not possible from client. Shutting down a member/cluster is not possible using public API. Management Center has private, management API for that

docs/modules/maintain-cluster/pages/restart-cluster.adoc
k-jamroz: rolling upgrade is already described https://docs.hazelcast.com/hazelcast/5.5/maintain-cluster/rolling-upgrades#rolling-upgrade-procedure. rolling restart would be the same except version changes

whole cluster restart depends if you have persistence/hotrestart (makes sense to be graceful and there are more options during start + additional cluster states) or not (simply kill and start)

Rob-Hazelcast: Thank you, that's useful context. So we can't cover every possibility here. But I don't think it's acceptable to require a restart to change most product config and not tell users how to perform it.

The rolling upgrade procedure is poor because it doesn't tell you _how_ to do most of the steps and links out do other docs mid-flow, which users hate. I agree the method is much the same and they could arguably be combined, but I think it's useful to have dedicated docs for each - there's specific info about versions etc. in there which is useful for upgrading and a distraction for restarting.

k-jamroz: > The rolling upgrade procedure is poor because it doesn't tell you how to do most of the steps and links out do other docs mid-flow, which users hate

I agree it would be better to provide some concrete examples how to perform these steps. But there are so many ways to do it depending on how you deploy the cluster (manual using hz-cli with or without some steps done using MC, systemd, ansible, helm chart, operator, embedded cluster).

Maybe we could provide docs for a basic option (hz-cli?) and other ways have their own documentation so we avoid duplication. Alternatively, maybe we should have a separate documentation belonging to particular component how to perform restart in each of the options (hz-cli, operator, maybe MC etc.) instead of putting them all together.

And actually in current state of the PR we already provide something like this. But parts of restart documentation overlap with and duplicate https://docs.hazelcast.com/hazelcast/5.5/maintain-cluster/shutdown - eg. section about how to shut down a member.

Rob-Hazelcast: Fair point. That's also duplication though, right? Do you really want 10 different topics for one procedure? We have the tabs so you can cover multiple methods in one place, which is surely more convenient for both us and users. Happy to have some level of compromise though, e.g. Operator covers it in its own docs, and if you use an orchestrator then it's reasonable to leave the detailed procedure up to the customer and just outline Hazelcast's requirements. But in principle, the docs need to explain how to use the product.

If you need to perform a shut down as a key part of a procedure, the instructions need to be in that procedure, duplication be damned. You can't send users off to another doc in the middle of a procedure, it's a real pain point. I can potentially use an `include` so it's not duplicated in the source.

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+Depending on the scope of your configuration changes, you may need to update a single member, a cluster, or multiple clusters. The process is the same in all cases.
+
+TIP: You can also use the rolling restart process for canarying: testing changes on a single member before updating the rest of the cluster.
+
+Some configuration needs to be consistent across all members in a cluster. To update this configuration, you must restart the whole cluster simultaneously. By default, this causes all data in the cluster to be lost. To avoid data loss, you must enable xref:storage:persistence.adoc[persistence]. To avoid downtime, you must divert traffic to another Hazelcast cluster before restarting.
+
+Note that members are not restarted. In all cases, members are terminated, and new members are created to replace them.
```
k-jamroz, line 21: I still think that this rather confuses than clarifies because there are different levels of member "identity" - some are preserved (address; UUID with HotRestart)

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,189 @@
...
+[[rolling]]
+=== Perform a rolling restart
+
+You should check the cluster is healthy before and after making changes. The best way to do this is to run a xref:{page-latest-supported-mc}@management-center:clusters:healthcheck.adoc[Healthcheck] in Management Center. For other options, see xref:maintain-cluster:monitoring.adoc[].
+
+To perform a rolling restart:
+
+. Update the static configuration as required.
+
+. Choose a member and make a note of its name or IP address to help you keep track.
```
k-jamroz, line 48: isn't this too obvious? also the procedure does not make any use of it later

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,189 @@
...
+=== Perform a whole cluster restart
+
+You should check the cluster is healthy before and after making changes. The best way to do this is to run a xref:{page-latest-supported-mc}@management-center:clusters:healthcheck.adoc[Healthcheck] in Management Center. For other options, see xref:maintain-cluster:monitoring.adoc[].
+
+To perform a whole cluster restart:
+
+. Update the static configuration as required.
+
+. Divert traffic to another Hazelcast cluster. How to do this depends on how your Hazelcast deployment is architected.
++
```
k-jamroz, line 120: IMO this looks a bit odd, single bullet point should be better:

image

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,189 @@
...
+. Gracefully shut down the cluster:
++
+[tabs]
+====
+CLI::
++
+--
+[source,bash]
+----
+hz-cluster-admin -a <MEMBER_ADDRESS> -c <CLUSTER_NAME> -o shutdown
```
k-jamroz, line 132: you can also invoke `hz-stop` on each member. `hz-cluster-admin` requires deprecated REST API

docs/modules/maintain-cluster/pages/restart-cluster.adoc
```diff
@@ -0,0 +1,189 @@
...
+
+To perform a whole cluster restart:
+
+. Update the static configuration as required.
+
+. Divert traffic to another Hazelcast cluster. How to do this depends on how your Hazelcast deployment is architected.
++
+Management Center provides an easy way to xref:{page-latest-supported-mc}@management-center:clusters:client-filtering.adoc[manage client connections] using filtering rules.
+
+. Gracefully shut down the cluster:
```
k-jamroz, line 123:
```suggestion
. Change cluster state to `NO_MIGRATION` or `FROZEN` state as needed with Persistence
. Gracefully shut down the cluster:
```

+ example of changing the state. [Should be possible with MC and hz-cluster-admin](https://docs.hazelcast.com/hazelcast/5.5/maintain-cluster/cluster-member-states#changing-a-clusters-state).

docs/modules/maintain-cluster/pages/restart-cluster.adoc
Rob-Hazelcast: Thanks for clarifying, I agree that's better.

PR #1845, opened 2025-08-27
Use EE version in hazelcast-spring dependency

k-jamroz:
EE example should use most up-to-date patched version of `hazelcast-spring`.

Currently the discrepancy is visible in 5.5 docs (5.5.0 vs 5.5.7)


TomaszGaweda (approved):
Good catch, thanks!

PR #1879, opened 2025-09-17
Update monitoring.adoc

erkangurgen:
properties yaml format should be 
<property-name>: <property-value>


amandalindsay (approved):
Thanks for spotting!

PR #1881, opened 2025-09-19
Drop patch numbers from supported Spring framework versions

JackPGreen:
- we can trust that if we support _a_ patch version of Spring, we support all in the same minor
- avoids updating the documentation every time a patch is released

### DIFF
--- a/docs/modules/spring/pages/overview.adoc
+++ b/docs/modules/spring/pages/overview.adoc
@@ -26,8 +26,4 @@ You can embed Hazelcast members in the same Java process as your Spring applicat
 
 == Supported versions
 
-Hazelcast supports the following Spring Framework versions:
-
-* **Supported**: 5.3
-
-* **Latest tested**: 6.2
+Hazelcast supports Spring Framework `6.2` and `5.3`.


docs/modules/spring/pages/overview.adoc
```diff
@@ -28,6 +28,6 @@ You can embed Hazelcast members in the same Java process as your Spring applicat
 
 Hazelcast supports the following Spring Framework versions:
 
-* **Supported**: 5.3.39+
+* **Supported**: 5.3
```
Rob-Hazelcast, line 31:
```suggestion
* **Supported**: 5.3 and higher
```

Rob-Hazelcast (approved):

docs/modules/spring/pages/overview.adoc
Rob-Hazelcast: Huh. Unusual to claim support for two versions and nothing in between.

TomaszGaweda: I'd change to something like:

> 5.3.x and 6.2.x
> Tested on latest patch release of each mentioned major version.

JackPGreen: https://github.com/hazelcast/hz-docs/pull/1881/commits/8f0e10ae385f79e9f8a79a8f1e61145b70e3f317

Rob-Hazelcast (approved):
Looks good, thanks. Looking at the version history this is more reasonable than it seemed at first glance.

https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Versions

PR #1894, opened 2025-09-29
Warning that MySQL 5.7 does not support exactly-once processing [CTT-429]

JamesHazelcast:
It was discovered that a bug in MySQL 5.7 can break this guarantee - since it will not be fixed, we are recommending that users upgrade to MySQL 8 or above.

Fixes docs for https://hazelcast.atlassian.net/browse/CTT-429

### DIFF
--- a/docs/modules/integrate/pages/jdbc-connector.adoc
+++ b/docs/modules/integrate/pages/jdbc-connector.adoc
@@ -143,3 +143,5 @@ some records will be missing from the target database. To test your
 broker we provide a tool, please go to link:https://github.com/hazelcast/hazelcast-jet-contrib/tree/master/xa-test[XA tests]
 to get more information. This only applies to the JDBC sink, the source
 doesn't use XA transactions.
+
+WARNING: The exactly-once processing guarantee is not supported in MySQL 5.7 due to a https://bugs.mysql.com/bug.php?id=87836[known bug in that version] which will not be fixed. Please use MySQL 8 or above, as recommended by Oracle.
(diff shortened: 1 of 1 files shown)


docs/modules/integrate/pages/jdbc-connector.adoc
```diff
@@ -143,3 +143,5 @@ some records will be missing from the target database. To test your
 broker we provide a tool, please go to link:https://github.com/hazelcast/hazelcast-jet-contrib/tree/master/xa-test[XA tests]
 to get more information. This only applies to the JDBC sink, the source
 doesn't use XA transactions.
+
+WARNING: The exactly-once processing guarantee is not supported in MySQL 5.7 due to a https://bugs.mysql.com/bug.php?id=87836[known bug in that version] which will not be fixed. Due to this, as well as the age and lack of support for MySQL 5.7, it is no longer supported by Hazelcast. Please use MySQL 8 or above, as recommended by Oracle.
```
Rob-Hazelcast, line 147:
```suggestion
WARNING: The exactly-once processing guarantee is not supported in MySQL 5.7 due to a https://bugs.mysql.com/bug.php?id=87836[known bug in that version]. MySQL 5.7 is not supported by Hazelcast. Please use MySQL 8 or above, as recommended by Oracle.
```
No doubt it's true, but it's not really for us to say that another company won't fix something. Let's keep it snappy.

Rob-Hazelcast (approved):

docs/modules/integrate/pages/jdbc-connector.adoc
JamesHazelcast: It won't be fixed because it's out of support (it will not get any patches) - it's fixed in later majors because they're still supported. Happy to remove though, I was already on the fence about whether it made sense to call it out.

PR #1913, opened 2025-10-07
use build hook from secrets

fantkolja:


JackPGreen:
> > How will this work with https://github.com/hazelcast/hazelcast-mono/blob/eddaeece59b43459ccef22c91da2df765ac8c4f0/.github/workflows/publish-docs.yml#L15? Can you test it?
> 
> Nice catch! I am thinking, to do [this](https://github.com/hazelcast/hazelcast-mono/pull/5381) to make things less complicated

Can you remove `workflow_call` from these action(s) to prevent this being an issue?

PR #1946, opened 2025-10-20
New cluster property - hazelcast.journal.cleanup.threshold [CTT-761, CTT-739]

shultseva:
The cleanup threshold property for the event journal.

Introduced here: 
https://hazelcast.atlassian.net/browse/CTT-739
https://hazelcast.atlassian.net/browse/CTT-761

### DIFF
--- a/docs/modules/ROOT/pages/system-properties.adoc
+++ b/docs/modules/ROOT/pages/system-properties.adoc
@@ -573,7 +573,7 @@ idle CPU usage but may result in decreased latency. Higher values will increase 
 | 1.0
 | float
 | The cleanup threshold for the event journal, expressed as a fraction of its capacity (a value between 0.0 and 1.0).
-When the remaining capacity falls below this threshold, a cleanup may be triggered during put operations to remove expired events (only if a `timeToLiveSeconds` for event journal is configured).
+When the remaining capacity falls below this threshold, a cleanup may be triggered during put operations to remove expired events (only if a xref:data-structures:event-journal.adoc#configuring-event-journal-ttl[time-to-live] for event journal is configured).
 A value of 0.0 means that no cleanup is performed, while a value of 1.0 means cleanup is triggered after each put operation.
 
 |`hazelcast.local.localAddress`


Rob-Hazelcast (changes requested):

docs/modules/ROOT/pages/system-properties.adoc
```diff
@@ -569,6 +569,14 @@ idle CPU usage but may result in decreased latency. Higher values will increase
 |int
 |Some JMX MBeans are cached  to reduce the overhead of calculating statistics. This parameter determines for how long the MBeans can go stale.
 
+|`hazelcast.journal.cleanup.threshold`
+| 1.0f
+| float
+| The cleanup threshold for the event journal, expressed as a fraction of its capacity (a value between 0.0f and 1.0f).
+When the remaining capacity falls below this threshold, a cleanup may be triggered during put operations to remove expired events (only if a TTL is configured).
+A value of 0.0f means cleanup occurs when the journal is full, while a value of 1.0f means cleanup is triggered after each put operation.
```
k-jamroz: > A value of 0.0f means cleanup occurs when the journal is full

It rather means that there is no cleanup, only overwriting the entries when the journal fills

docs/modules/ROOT/pages/system-properties.adoc
```diff
@@ -569,6 +569,14 @@ idle CPU usage but may result in decreased latency. Higher values will increase
 |int
 |Some JMX MBeans are cached  to reduce the overhead of calculating statistics. This parameter determines for how long the MBeans can go stale.
 
+|`hazelcast.journal.cleanup.threshold`
+| 1.0f
+| float
+| The cleanup threshold for the event journal, expressed as a fraction of its capacity (a value between 0.0f and 1.0f).
+When the remaining capacity falls below this threshold, a cleanup may be triggered during put operations to remove expired events (only if a TTL is configured).
```
k-jamroz:
```suggestion
When the remaining capacity falls below this threshold, a cleanup may be triggered during put operations to remove expired events (only if a `timeToLiveSeconds` for event journal is configured).
```

docs/modules/ROOT/pages/system-properties.adoc
```diff
@@ -569,6 +569,14 @@ idle CPU usage but may result in decreased latency. Higher values will increase
...
+A value of 0.0f means cleanup occurs when the journal is full, while a value of 1.0f means cleanup is triggered after each put operation.
+Applies to event journals of both IMap and ICache.
```
k-jamroz:
```suggestion
```

It applies to all data structures having journal (currently IMap and ICache) and the design is generic and with the idea that other structures would behave in the same way. So I would remove this explicit enumeration.

docs/modules/ROOT/pages/system-properties.adoc
k-jamroz: maybe add link to the [[configuring-event-journal-ttl]] section?

docs/modules/ROOT/pages/system-properties.adoc
```diff
@@ -569,6 +569,14 @@ idle CPU usage but may result in decreased latency. Higher values will increase
 |int
 |Some JMX MBeans are cached  to reduce the overhead of calculating statistics. This parameter determines for how long the MBeans can go stale.
 
+|`hazelcast.journal.cleanup.threshold`
+| 1.0f
```
k-jamroz: the `f` suffix is not necessary when configuring the property and may be puzzling for for non-java users

docs/modules/ROOT/pages/system-properties.adoc
```diff
@@ -569,6 +569,14 @@ idle CPU usage but may result in decreased latency. Higher values will increase
...
+| float
+| The cleanup threshold for the event journal, expressed as a fraction of its capacity (a value between 0.0f and 1.0f).
```
k-jamroz:
```suggestion
| The cleanup threshold for the event journal, expressed as a fraction of its capacity (a value between 0.0 and 1.0).
```

docs/modules/ROOT/pages/system-properties.adoc
k-jamroz:
```suggestion
A value of 0.0 means cleanup occurs when the journal is full, while a value of 1.0 means cleanup is triggered after each put operation.
```

shultseva: d6a3e1ff830590d327569846a6d1ddfb31963fd0

docs/modules/ROOT/pages/system-properties.adoc
```diff
@@ -569,6 +569,13 @@ idle CPU usage but may result in decreased latency. Higher values will increase
 |int
 |Some JMX MBeans are cached  to reduce the overhead of calculating statistics. This parameter determines for how long the MBeans can go stale.
 
+|`hazelcast.journal.cleanup.threshold`
+| 1.0
+| float
+| The cleanup threshold for the event journal, expressed as a fraction of its capacity (a value between 0.0 and 1.0).
+When the remaining capacity falls below this threshold, a cleanup may be triggered during put operations to remove expired events (only if a xref:data-structures:event-journal.adoc#configuring-event-journal-ttl[time-to-live] for event journal is configured).
```
k-jamroz: > during put operations

I am wondering (maybe nit-picking) if it is clear that we mean here journal/ringbuffer put operation and not IMap/ICache put operation

k-jamroz (approved):

PR #1953, opened 2025-10-22
Update C++ instructions post-`boost` upgrade

JackPGreen:
Update the instructions after https://github.com/hazelcast/hazelcast-cpp-client/pull/1322, specifically copying the changes from the (partially duplicated) `Reference_Manual`.


JackPGreen:
> Can you also port updates from other sections like [requirements](https://docs.hazelcast.com/hazelcast/latest/clients/cplusplus) as well?

I didn't see that. In which case I don't think copying is the solution - we need to remove the duplication.

@amandalindsay FYI.

PR #1975, opened 2025-10-28
VectorCollection: CPU Capacity Planning [AI-328]

gbarnett-hz:
Notes on `hz:vector-query` and how it relates to overall CPU utilisation.


oliverhowell (approved):
One minor suggestion but approving in advance (might rewrite the passive bits when I get into vector docs properly)

PR #1998, opened 2025-11-11
DOC-300: replace compatibility table

amandalindsay:
Replace ‘Table 5. Client Version Compatibilities’ with a table similar to the ‘V1: Compatibility Range’ table in Burak's Confluence page, but do not include the LTS/STS column.


Rob-Hazelcast:
Superseded by https://github.com/hazelcast/hz-docs/pull/2056

PR #2019, opened 2025-12-01
Update Jet Python versions for 5.4 [CTT-271]

JamesHazelcast:
Backport (somewhat) of: https://github.com/hazelcast/hz-docs/pull/2007

For 5.4 and 5.5, we require Python 3.8+ and officially support 3.8-3.13.

Fixes https://hazelcast.atlassian.net/browse/CTT-271

### DIFF
--- a/docs/modules/pipelines/pages/transforms.adoc
+++ b/docs/modules/pipelines/pages/transforms.adoc
@@ -258,7 +258,7 @@ to efficient batching.
 Hazelcast can call Python code to perform a mapping step in the
 pipeline. The prerequisite is that the Hazelcast servers are Linux or Mac with Python installed and that the `hazelcast-jet-python` module is deployed
 on the classpath, through being present in the `lib` directory. Hazelcast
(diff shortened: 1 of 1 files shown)


docs/modules/pipelines/pages/python.adoc
```diff
@@ -13,7 +13,7 @@ To complete this tutorial, you need the following:
 |===
 |Prerequisites|Useful resources
 
-|Python 3.7+ |link:https://www.python.org/downloads[Python downloads]
+|Python 3.8+ |link:https://www.python.org/downloads[Python downloads]
```
k-jamroz: should we mention that 3.8 is required starting in 5.4.4 for transparency?

Rob-Hazelcast: Generally avoid this kind of thing as the docs are versioned, but maybe reasonable if it's a new requirement in a patch release. @oliverhowell what do you think?

PR #2035, opened 2025-12-11
DOC-300-2: add static compatibility table

amandalindsay:
Add new 'static' compatibility table to show compatibility between Hazelcast Platform, its management and monitoring components, and clients.

Position in nav is TBC - temporarily under main versioning & compatibility topic


amandalindsay:
> https://docs.hazelcast.com/hazelcast/5.6/deploy/versioning-compatibility is the obvious home for this. Has most/all the same info already, just in a worse format.

I'm aware of that - read my comments in Slack please and await the discussion with PM

jgardiner68:
@Rob-Hazelcast Please update the table.

Needs a row for Platform 5.6.x, MC 5.9.0 and Operator 5.16.0 are the latest version

Rob-Hazelcast:
Not sure about this one. We have all the same info in https://github.com/hazelcast/hz-docs/pull/1998/changes in an easier to maintain format. We don't want statically maintained version numbers when we use variables everywhere else. And most of the table is effectively blank anyway.

Better for users to use the existing version info page and if for some reason they want to the info for more than one platform version they can change the version with the drop-down picker.

PR #2139, opened 2026-03-09
DOC-746 - CP auto stepdown

Rob-Hazelcast:


gareth-johnston (changes requested):
First pass, thanks @Rob-Hazelcast

gareth-johnston (changes requested):
Some more changes needed, thanks @Rob-Hazelcast

docs/modules/cp-subsystem/images/autostepdown-deployment.svg
gareth-johnston: Just noticing this diagram - Can we change the 2ms to <1ms and the remote DC to like 4 or 5. Would be best to advertise more conservative values

gareth-johnston (commented):
Some more requests, should be near the end, thanks @Rob-Hazelcast

gareth-johnston (approved):
Thanks @Rob-Hazelcast, solid job

JamesHazelcast (approved):

burakcelebi (approved):
Let's add ADVANCED_CP

PR #2159, opened 2026-04-14
Remove recommendation to use TPC

wyohm:
TPC is experimental and should not be recommended at this time.
On [the TPC page](https://docs.hazelcast.com/hazelcast/5.6/cluster-performance/thread-per-core-tpc):
> This is an experimental feature and is not recommended for production use. It is separately licensed. If you are interested in trialing this feature, contact your Hazelcast representative.


g-donev (approved):
Looks good.

PR #2168, opened 2026-04-20
Relocate documentation `hazelcast.client.io` `thread.count` properties to the client property page

JackPGreen:
Rather than describing the property on the `performance-tips` page, we should link to the existing documentation on the property and expand the definition there.


g-donev (approved):
Looks good.

PR #2169, opened 2026-04-20
Add links to Javadoc to `Client System Properties`

JackPGreen:
The `com.hazelcast.client.properties.ClientProperty` class contains Javadoc for each of the properties.

We _additionally_ link to the Javadoc for the respective property to add additional information.

_Optionally_ in the future we could remove some of the detail from this page to have a single-source-of-truth.

### DIFF
--- a/docs/modules/clients/pages/java.adoc
+++ b/docs/modules/clients/pages/java.adoc
@@ -1889,12 +1889,12 @@ NOTE: You need to restart clients after modifying system properties.
 |===
 |Property Name | Default Value | Type | Description
 
-|`hazelcast.client.cloud.discovery.token`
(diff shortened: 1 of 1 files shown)


docs/modules/clients/pages/java.adoc
```diff
@@ -2093,7 +2093,7 @@ If both are configured, this one is ignored.
 Note that since this is replaced with `hazelcast.client.metrics.enabled`, the default behavior
 is also changed: it is enabled by default.
 
-|`hazelcast.client.statistics.period.seconds`
```
g-donev, line 2096: @JackPGreen This property and the one above [hazelcast.client.statistics.enabled] are deprecated since 4.0. The replacement properties are already in this table. Should we just remove them?

JackPGreen: I noticed the same. I don’t think we should remove them - someone might be using them in already and if it’s not documented that would be very confusing.

We could (and probably should have when we did it…) note they are deprecated in the doc? But not sure if we have an existing style for that.

but I think this change is out of scope for adding links.

g-donev (approved):
Approved.

PR #2259, opened 2026-07-22
Fix OrderKey code snipped

k-jamroz:
OrderKey snipped was prematurely truncated. It is used in https://docs.hazelcast.com/hazelcast/latest/cluster-performance/data-affinity/


mscwilson (approved):
good spot

mscwilson:
does this need to be backported?

k-jamroz:
> does this need to be backported?

yes, added labels

k-jamroz:
compile problems are unrelated to this change, but caused by changes in hazelcast-vector packaging

PR #2264, opened 2026-07-23
Add info about security filters [DOC-768]

mscwilson:
https://hazelcast.atlassian.net/browse/DOC-768

Documents the SQL reflection and zero-config Compact serialisation filters.

### DIFF
--- a/docs/modules/secure-cluster/pages/hardening-recommendations.adoc
+++ b/docs/modules/secure-cluster/pages/hardening-recommendations.adoc
@@ -57,6 +57,7 @@ To avoid deserialization of objects from untrusted sources, Hazelcast offers som
 xref:security:tls-ssl.adoc#mutual-authentication[mutual TLS authentication] and disabling xref:clusters:network-configuration.adoc#multicast-element[multicast join] configuration.
 We recommend using Java serialization filter configuration for whitelisting the set of trusted classes or
 packages which are allowed for deserialization.
+* Hazelcast can fall back to zero-configuration Compact serialization for classes that have no explicit serializer. We recommend restricting the classes eligible for zero-config Compact serialization by defining an allowlist of only the classes your application needs, rather than relying on the default behavior. See xref:serialization:compact-serialization.adoc#zero-config-filter[Restricting Classes for Zero-Config Compact Serialization].
 * Hazelcast uses Java reflection during SQL execution when the object format is set to `java`. We recommend using xref:sql:sql-reflection-configuration.adoc#configuring-reflection[Java reflection filter configuration] to whitelist the set of trusted classes or packages that are allowed to create through reflection.
 * You can disable script executions on the Hazelcast members. Scripts executed from Management center have access
 to system resources (files, etc.) with privileges of user running Hazelcast.
--- a/docs/modules/secure-cluster/pages/hardening-recommendations.adoc
+++ b/docs/modules/secure-cluster/pages/hardening-recommendations.adoc
@@ -56,8 +56,9 @@ and, if clients are used, also in the `hazelcast-client.xml/yaml`. Access to the
 To avoid deserialization of objects from untrusted sources, Hazelcast offers some protection mechanisms; we recommend enabling
 xref:security:tls-ssl.adoc#mutual-authentication[mutual TLS authentication] and disabling xref:clusters:network-configuration.adoc#multicast-element[multicast join] configuration.
 We recommend using Java serialization filter configuration for whitelisting the set of trusted classes or
-packages which are allowed for deserialization.
-* Hazelcast can fall back to zero-configuration Compact serialization for classes that have no explicit serializer. We recommend restricting the classes eligible for zero-config Compact serialization by defining an allowlist of only the classes your application needs, rather than relying on the default behavior. See xref:serialization:compact-serialization.adoc#zero-config-filter[Restricting Classes for Zero-Config Compact Serialization].
+packages which are allowed for deserialization. Configure these restrictions on both members and clients. The configuration does not have to be the same on each. Bear in mind that a class blocked on a given member or client cannot be used there.
+* Hazelcast can use zero-configuration Compact serialization for eligible classes that have no explicit serializer. We recommend restricting the classes allowed to use zero-config Compact serialization by defining an allowlist of only the classes your application needs, rather than relying on the default behavior, and configuring this on both members and clients. See xref:serialization:compact-serialization.adoc#zero-config-filter[Restricting Classes for Zero-Config Compact Serialization].
+If you do not want to use zero-config Compact serialization, disable it by disabling the defaults and configuring an empty allowlist.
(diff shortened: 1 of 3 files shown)


docs/modules/secure-cluster/pages/hardening-recommendations.adoc
```diff
@@ -57,6 +57,7 @@ To avoid deserialization of objects from untrusted sources, Hazelcast offers som
 xref:security:tls-ssl.adoc#mutual-authentication[mutual TLS authentication] and disabling xref:clusters:network-configuration.adoc#multicast-element[multicast join] configuration.
 We recommend using Java serialization filter configuration for whitelisting the set of trusted classes or
 packages which are allowed for deserialization.
+* Hazelcast can fall back to zero-configuration Compact serialization for classes that have no explicit serializer. We recommend restricting the classes eligible for zero-config Compact serialization by defining an allowlist of only the classes your application needs, rather than relying on the default behavior. See xref:serialization:compact-serialization.adoc#zero-config-filter[Restricting Classes for Zero-Config Compact Serialization].
```
k-jamroz:
```suggestion
* Hazelcast can use zero-configuration Compact serialization for eligible classes that have no explicit serializer. We recommend restricting the classes allowed to use zero-config Compact serialization by defining an allowlist of only the classes your application needs, rather than relying on the default behavior. See xref:serialization:compact-serialization.adoc#zero-config-filter[Restricting Classes for Zero-Config Compact Serialization].
If you do not use zero-configuration Compact serialization, disable defaults and configure empty allowlist for best security.
```

@th0masb could you verify this, in particular the requirement about disabling defaults to configure empty allowlist. We might have discussed this previously, but I realized (again?) that this is a bit unfortunate. If you start with empty allowlist and disabled defaults, but then want to add some unreasonable classes to allowlist (eg. `com.hazelcast`) you lose the safety net of default blocklist.

docs/modules/serialization/pages/compact-serialization.adoc
```diff
@@ -1132,6 +1132,86 @@ module org.example.Foo {
 }
 ----
 
+[[zero-config-filter]]
+== Restricting Classes for Zero-Config Compact Serialization
+
+Zero-config Compact serialization uses reflection to serialize any class that does not have another registered serializer. To control which classes are eligible for this, and to prevent classes from being serialized unintentionally, configure a filter that allows or blocks classes.
+
+The filter uses an allowlist and a blocklist. This is the same format as the xref:sql:sql-reflection-configuration.adoc[SQL Java reflection filter]. You can allow or block classes by fully qualified class name, package name, or class-name prefix. The filter can be set on both members and clients.
```
k-jamroz: it is better to link to https://docs.hazelcast.com/hazelcast/5.7/serialization/serialization-configuration#untrusted-classes instead of SQL docs (SQL is a specific feature)

docs/modules/secure-cluster/pages/hardening-recommendations.adoc
k-jamroz: I would also add a recomendation to configure serialization restrictions (Java - previous point) and ZCCS (this point) not only on members but also on clients. And they do not have to be the same, although blocked classes will not be usable in client/member that has them blocked.

docs/modules/serialization/pages/compact-serialization.adoc
```diff
@@ -1132,6 +1132,86 @@ module org.example.Foo {
...
+
+When a filter is configured:
+
+* If a class is on the blocklist, serializing an object of that class throws an exception.
+* If a class is on the blocklist, deserializing its data returns a `GenericRecord` instead of an instance of the class, rather than throwing an exception.
```
k-jamroz: this is actually quite nice feature, possibly worth highlighting a bit more. some features like PredicateAPI/SQL query can process a compact-serialized class even if it is blocked from serialization, because extracting individual fields without object creation is safe.
entry processor or other member-side code can use `GenericRecord` to access the data, which may be sufficient in some cases

docs/modules/serialization/pages/compact-serialization.adoc
```diff
@@ -1132,6 +1132,86 @@ module org.example.Foo {
...
+* If the allowlist is non-empty, only classes on the allowlist (and not on the blocklist) are eligible for zero-config Compact serialization.
```
k-jamroz: > only classes on the allowlist (and not on the blocklist)

both conditions need to be met. partentheses introduce a bit of assymetry in the text. Maybe:

```suggestion
* If the allowlist is non-empty, only classes on the allowlist which are not on the blocklist are eligible for zero-config Compact serialization.
```
?

docs/modules/serialization/pages/compact-serialization.adoc
```diff
@@ -1132,6 +1132,86 @@ module org.example.Foo {
...
 ----
 
+[[zero-config-filter]]
+== Restricting Classes for Zero-Config Compact Serialization
+
+Zero-config Compact serialization uses reflection to serialize any class that does not have another registered serializer. To control which classes are eligible for this, and to prevent classes from being serialized unintentionally, configure a filter that allows or blocks classes.
+
+The filter uses an allowlist and a blocklist. This is the same format as the xref:sql:sql-reflection-configuration.adoc[SQL Java reflection filter]. You can allow or block classes by fully qualified class name, package name, or class-name prefix. The filter can be set on both members and clients.
+
+When a filter is configured:
```
k-jamroz, line 1144: describe also how to [configure empty filter](https://github.com/hazelcast/hz-docs/pull/2264/files#r3646396610) to effectively disable ZCCS

docs/modules/serialization/pages/compact-serialization.adoc
```diff
@@ -1132,6 +1132,86 @@ module org.example.Foo {
 }
 ----
 
+[[zero-config-filter]]
+== Restricting Classes for Zero-Config Compact Serialization
+
+Zero-config Compact serialization uses reflection to serialize any class that does not have another registered serializer. To control which classes are eligible for this, and to prevent classes from being serialized unintentionally, configure a filter that allows or blocks classes.
```
k-jamroz, line 1138: these restrictions apply only to ZCCS. it is possible to register explicit serializers for classes that are not allowed for ZCCS.

docs/modules/secure-cluster/pages/hardening-recommendations.adoc
mscwilson: I've confirmed this behaviour - check L55 of ReflectiveClassFilter

th0masb: > f you start with empty allowlist and disabled defaults, but then want to add some unreasonable classes to allowlist (eg. com.hazelcast) you lose the safety net of default blocklist.

Yes from what I remember this is the expected behaviour. I don't think it is particularly a problem, if a customer disables defaults _and_ explicitly allows the unreasonable classes then it feels like we've done all we can to try and stop it being used. If they have a non-empty allowlist then they shouldn't be disabling defaults.

k-jamroz (approved):

neilmadden-hazelcast (approved):

PR #2265, opened 2026-07-23
Add ZCCS config properties [DOC-768]

mscwilson:
https://hazelcast.atlassian.net/browse/DOC-768

Lists the zero-config Compact serialisation filter properties. 

For v5.7, `CompactSerializationConfig.zeroConfigFilter` is preferred - documented in #2264.

### DIFF
--- a/docs/modules/ROOT/pages/system-properties.adoc
+++ b/docs/modules/ROOT/pages/system-properties.adoc
@@ -1036,6 +1036,41 @@ of setting a result size limit.
 |string
 |When set to any non-null value, security recommendations are logged at the `INFO` level during the member startup.
 
+|`hazelcast.serialization.compact.allowlist.classes`
+|
+|string
+|Comma-separated list of fully qualified class names to allow for xref:serialization:compact-serialization.adoc[zero-config Compact serialization] (ZCCS). If set, only classes on this list (and not on the blocklist) are eligible for ZCCS. Default is empty.
+
+|`hazelcast.serialization.compact.allowlist.packages`
+|
+|string
+|Comma-separated list of package names to allow for ZCCS. Only classes in the exact package (not sub-packages) are matched.
+
+|`hazelcast.serialization.compact.allowlist.prefixes`
+|
+|string
+|Comma-separated list of class-name prefixes to allow for ZCCS.
+
+|`hazelcast.serialization.compact.blocklist.classes`
+|
+|string
+|Comma-separated list of fully qualified class names to block from ZCCS. Default is empty.
+
+|`hazelcast.serialization.compact.blocklist.defaults`
+|true
+|bool
+|Controls whether the default blocklist of known-problematic classes is applied for ZCCS.
+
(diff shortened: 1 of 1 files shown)


k-jamroz:
this branch targets main (5.8) but lacks backport to 5.7. what was the intention?

docs/modules/ROOT/pages/system-properties.adoc
```diff
@@ -1036,6 +1036,41 @@ of setting a result size limit.
...
+
+|`hazelcast.serialization.compact.blocklist.classes`
+|
+|string
+|Comma-separated list of fully qualified class names to block from ZCCS. Default is empty.
+
+|`hazelcast.serialization.compact.blocklist.defaults`
+|true
+|bool
+|Controls whether the default blocklist of known-problematic classes is applied for ZCCS.
```
k-jamroz: unfortunately this property is `defaults` not `defaultsDisabled` (like elsewhere) which is different that in config and possible source of mistakes. IMO worth highlighting this fact

docs/modules/ROOT/pages/system-properties.adoc
```diff
@@ -1036,6 +1036,41 @@ of setting a result size limit.
 |string
 |When set to any non-null value, security recommendations are logged at the `INFO` level during the member startup.
 
+|`hazelcast.serialization.compact.allowlist.classes`
+|
+|string
+|Comma-separated list of fully qualified class names to allow for xref:serialization:compact-serialization.adoc[zero-config Compact serialization] (ZCCS). If set, only classes on this list (and not on the blocklist) are eligible for ZCCS. Default is empty.
```
k-jamroz, line 1042: these properties are available in 5.7 onwards (TBD if we remove them in 6.0) only to allow easier upgrade from <5.7 versions with hardened config.
from 5.7 onwards `zero-config-filter` XML/YAML config should be recommended

neilmadden-hazelcast (approved):

docs/modules/ROOT/pages/system-properties.adoc
mscwilson: this is why I didn't add the backport to 5.7 label initially, because the text needs to be different. now it occurs to me that I can just edit the backport PR to clarify this for 5.7

k-jamroz: [this PR](https://github.com/hazelcast/hz-docs/pull/2265) applies to 5.8/6.0 and, unless we make further changes, the same preference of using config instead of properties is true also in 5.8/6.0. I do not see any different wording in https://github.com/hazelcast/hz-docs/pull/2271

mscwilson: yeah as I said in Jira, I couldn't work out how to make the different changes for different versions in one PR. So I'll do it sequentially.

[stuff like this is one of the reasons I plan to migrate the docs to a monorepo, where that Jira ticket would be 1 PR instead of maybe 9]

PR #2301, opened 2026-09-02
Pulsar Data Connector docs [CTT-1428]

TomaszGaweda:
Includes:
- Basic connector info
- SQL mappings
- Data Connection

I didn't include too much of Java API details, there are easily accessible after the release. Wanted to focus on basic concepts and ideas "why and how".

### DIFF
--- a/docs/modules/data-connections/pages/pulsar-dc-configuration.adoc
+++ b/docs/modules/data-connections/pages/pulsar-dc-configuration.adoc
@@ -1,0 +1,168 @@
+[[Pulsar]]
+=== Example Apache  Pulsar Data Connection
+
+[tabs]
+====
+XML::
++
+--
+[source,xml]
+----
+<hazelcast>
+  <data-connection name="my-pulsar">
+    <type>Pulsar</type>
+    <properties>
+      <property name="brokerUrl">pulsar://pulsar.mydomain.com:6650</property> <1>
+      <property name="httpServiceUrl">http://pulsar.mydomain.com:8080</property> <2>
+    </properties>
+    <shared>true</shared>
+  </data-connection>
+  <data-connection name="my-other-pulsar">
+    <type>Pulsar</type>
+    <properties>
+      <property name="brokerUrl">pulsar://pulsar.mydomain.com:6650</property> <1>
+      <property name="httpServiceUrl">http://pulsar.mydomain.com:8080</property> <2>
+      <property name="tenant">myTenant</property> <3>
+      <property name="namespace">myNamespace</property> <3>
+      <property name="enableTransactions">true</property> <4>
+    </properties>
+    <shared>true</shared>
+  </data-connection>
+</hazelcast>
+----
+<1> (Required) Broker URL, mandatory property.
+<2> (Optional) Broker HTTP Service - necessary for resource listing.
+<3> (Optional) Options to limit which resources are accessible using this DataConnection.
+<4> (Optional) If true, transactions will be enabled on the client. Default is false. Please note, that transactions must be first enabled in the broker configuration.
+--
+
+YAML::
++
+--
+[source,yaml]
+----
+hazelcast:
+  data-connection:
+    my-pulsar:
+      type: Pulsar
+      properties:
+        brokerUrl: pulsar://pulsar.mydomain.com:6650 <1>
+        httpServiceUrl: http://pulsar.mydomain.com:8080 <2>
+      shared: true
+    my-other-mongodb:
+      type: Mongo
+      properties:
+        brokerUrl: pulsar://pulsar.mydomain.com:6650 <1>
+        httpServiceUrl: http://pulsar.mydomain.com:8080 <2>
+        tenant: myTenant <3>
+        namespace: myNamespace <3>
+        enableTransactions: true <4>
+      shared: true
+----
+<1> (Required) Broker URL, mandatory property.
+<2> (Optional) Broker HTTP Service - necessary for resource listing.
+<3> (Optional) Options to limit which resources are accessible using this DataConnection.
+<4> (Optional) If true, transactions will be enabled on the client. Default is false. Please note, that transactions must be first enabled in the broker configuration.
+--
+
+Java::
++
+--
+[source,java]
+----
+config
+  .addDataConnectionConfig(
+    new DataConnectionConfig("my-pulsar")
+      .setType("Pulsar")
+      .setProperty("brokerUrl", "pulsar://pulsar.mydomain.com:6650") <1>
+      .setProperty("httpServiceUrl", "http://pulsar.mydomain.com:8080") <2>
+      .setShared(true)
+  )
+  .addDataConnectionConfig(
+    new DataConnectionConfig("my-other-pulsar")
+      .setType("Pulsar")
+      .setProperty("brokerUrl", "pulsar://pulsar.mydomain.com:6650") <1>
+      .setProperty("httpServiceUrl", "http://pulsar.mydomain.com:8080") <2>
+      .setProperty("tenant", "myTenant") <3>
+      .setProperty("namespace", "myNamespace") <3>
+      .setProperty("enableTransactions", "true") <4>
+      .setShared(true)
+  );
+----
+<1> (Required) Broker URL, mandatory property.
+<2> (Optional) Broker HTTP Service - necessary for resource listing.
+<3> (Optional) Options to limit which resources are accessible using this DataConnection.
+<4> (Optional) If true, transactions will be enabled on the client. Default is false. Please note, that transactions must be first enabled in the broker configuration.
+--
+SQL::
++
+--
+Data connections created in SQL behave differently from those defined in members' configuration files or in Java.
+
+- To retain SQL-defined data connections after a cluster restart, you must enable xref:storage:configuring-persistence.adoc#sql[SQL metadata persistence]. This feature is available in the {enterprise-product-name}.
+- You can create or drop a data connection using SQL commands. To update a data connection, you need to drop and then recreate it. 
+
+[source,SQL]
+----
+CREATE DATA CONNECTION my_mongodb
+TYPE Mongo
+SHARED
+OPTIONS (
+    'brokerUrl' = 'pulsar://pulsar.mydomain.com:6650' <1>
+    'httpServiceUrl' = 'http://pulsar.mydomain.com:8080 '<2>
+)
+----
+<1> (Required) Broker URL, mandatory property.
+<2> (Optional) Broker HTTP Service - necessary for resource listing.
+
+[source,SQL]
+----
+CREATE DATA CONNECTION my_mongodb
+TYPE Mongo
+SHARED
+OPTIONS (
+    'brokerUrl' = 'pulsar://pulsar.mydomain.com:6650' <1>
+    'httpServiceUrl' = 'http://pulsar.mydomain.com:8080 '<2>
+    'tenant' = 'myTenant' <3>
+    'namespace' = 'myNamespace' <3>
+    'enableTransactions' = 'true' <4>
+);
+----
+
+<1> (Required) Broker URL, mandatory property.
+<2> (Optional) Broker HTTP Service - necessary for resource listing.
+<3> (Optional) Options to limit which resources are accessible using this DataConnection.
+<4> (Optional) If true, transactions will be enabled on the client. Default is false. Please note, that transactions must be first enabled in the broker configuration.
+--
+====
+
+[cols="1,1,3,1"]
+|===
+| Property | Default value | Description | When to use
+
+|`brokerUrl`
+|
+| Pulsar Broker URL, in a form of `pulsar://myhost:port/`
+| Mandatory
+
+| `httpServiceUrl`
+|
+| URL of Pulsar's instance HTTP service, typically on port 8080. Example: `http://myhost:8080/`
+| If you want to be able to list resources via Data Connections (not just treat like a connection pool).
+
+| `tenant`
+|
+| If set, it limits which resources (topics) will be listed to only those of given tenant.
+| Only if you want to restrict access to subset of topics.
+
+| `namespace`
+|
+| If set, it limits which resources (topics) will be listed to only those of given namespace.
+| Only if you want to restrict access to subset of topics.
+
+| `enableTransactions`
+| false
+| Enables transaction for returned Pulsar Clients.
+| If you want to use transactions with clients returned by this Data Connection. *Must* be set to true if used with Pulsar sink / SQL "INSERT INTO" Pulsar topic with `EXACTLY_ONCE` processing guarantee.
+
+|===
--- a/docs/modules/data-connections/pages/pulsar-dc-configuration.adoc
+++ b/docs/modules/data-connections/pages/pulsar-dc-configuration.adoc
@@ -49,8 +49,8 @@ hazelcast:
         brokerUrl: pulsar://pulsar.mydomain.com:6650 <1>
         httpServiceUrl: http://pulsar.mydomain.com:8080 <2>
       shared: true
-    my-other-mongodb:
-      type: Mongo
+    my-other-pulsar:
+      type: Pulsar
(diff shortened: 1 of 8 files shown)


docs/modules/data-connections/pages/pulsar-dc-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
++
+--
+Data connections created in SQL behave differently from those defined in members' configuration files or in Java.
+
+- To retain SQL-defined data connections after a cluster restart, you must enable xref:storage:configuring-persistence.adoc#sql[SQL metadata persistence]. This feature is available in the {enterprise-product-name}.
+- You can create or drop a data connection using SQL commands. To update a data connection, you need to drop and then recreate it. 
+
+[source,SQL]
+----
+CREATE DATA CONNECTION my_mongodb
```
JamesHazelcast: The second YAML example and both SQL examples use Mongo connections - both SQL blocks also omit commas between `OPTIONS` entries. Switch all three to Pulsar and fix the SQL.

docs/modules/integrate/pages/pulsar-connector.adoc
```diff
@@ -1,3 +1,83 @@
...
+
+If xref:security:enable-security.adoc[security] is enabled, your clients may need permissions to use this connector. For details, see xref:pipelines:job-security.adoc[].
+
+== Configuration Options
+
+To read from Pulsar as a source you must provide at least Schema, which Pulsar uses for given topic:
+
+[source,java]
+----
+Pipeline p = Pipeline.reate();
```
JamesHazelcast: `Pipeline.reate()` does not compile - neither builder sets the required `connectionSupplier` or `dataConnectionRef`, and the sink ends with a stray backtick. Fix the typos and set a connection on both builders.

docs/modules/sql/pages/mapping-to-pulsar.adoc
```diff
@@ -0,0 +1,214 @@
...
+== Creating a Pulsar Mapping
+
+To create a mapping to a Pulsar topic in SQL, you must use the `CREATE MAPPING` statement to tell the Pulsar connector the following:
+
+- The name of the topic
+- The data connection name
+
+[source,sql]
+----
+CREATE MAPPING my_topic <1>
```
JamesHazelcast: This introductory `CREATE MAPPING` omits `valueFormat`, so `KvMetadataResolvers` rejects it with `Missing 'valueFormat' option`. Add a supported `valueFormat`.

docs/modules/sql/pages/mapping-to-pulsar.adoc
```diff
@@ -0,0 +1,214 @@
...
+For keys and values that are primitives, set the format as the SQL data type that corresponds to the primitive. For example, if the message key is an integer and the value is a
+string, use the `int` and `varchar` formats:
+
+```sql
+CREATE MAPPING my_topic
+TYPE Pulsar
+OPTIONS (
+    'keyFormat'='int',
+    'valueFormat'='varchar',
+    'bootstrap.servers' = '127.0.0.1:9092'
```
JamesHazelcast: This example omits `DATA CONNECTION` and uses Kafka's `bootstrap.servers`, which Pulsar ignores - the first read or write reaches `Utils.getClient()` with neither a supplier nor `DataConnectionRef` and fails. Reference `myPulsar` here.

docs/modules/sql/pages/mapping-to-pulsar.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+[source,sql]
+----
+CREATE MAPPING my_topic
+DATA CONNECTION myPulsar
+OPTIONS (
+    'keyFormat' = 'java',
+    'keyJavaClass' = 'java.lang.Long',
+    'valueFormat' = 'java',
+    'valueJavaClass' = 'com.example.Person'
```
JamesHazelcast: `valueFormat='java'` cannot handle `com.example.Person` here - `PulsarTable.resolveSchema()` only supports the built-in types and throws `UnsupportedOperationException` for this class. The SQL also misses its closing `);`. Use a supported format/class, or add arbitrary Java object support.

docs/modules/sql/pages/mapping-to-pulsar.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+- The name of the topic
+- The data connection name
+
+[source,sql]
+----
+CREATE MAPPING my_topic <1>
+DATA CONNECTION myPulsar <2>
+OPTIONS (
+    'preferredLocalParallelism' = '2' <3>
```
JamesHazelcast: `PulsarTable` never reads `preferredLocalParallelism` - it always returns `LOCAL_PARALLELISM_USE_DEFAULT`, so the documented value `2` has no effect. The partition guidance is also Kafka-specific - this connector uses a Shared subscription rather than one-to-one partition assignment. Remove this option and guidance or implement them for Pulsar.

docs/modules/integrate/pages/connectors.adoc
```diff
@@ -176,14 +186,14 @@ The Jet API supports more connectors than SQL.
...
 |streaming
-|N/A
+|exactly-once
 
 |xref:integrate:pulsar-connector.adoc[PulsarSources.pulsarReader]
-|hazelcast-jet-contrib-pulsar
+|hazelcast-jet-pulsar
 |streaming
-|exactly-once
+|N/A
```
JamesHazelcast: The PRD specifies an exactly-once reader, and `PulsarReaderBuilder` still describes itself as fault tolerant and snapshots/restores message IDs - `N/A` contradicts both. Keep the documented guarantee aligned with the implementation.

docs/modules/sql/pages/mapping-to-pulsar.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+The Pulsar connector acts as a consumer or a producer so that you can create streaming queries that continuously query a given Pulsar topic, and also write to a Pulsar topic.
+
+== Installing the Connector
+
+This connector is included in the full distribution of Hazelcast.
+
+If you're using the slim distribution, you must add the link:https://search.maven.org/artifact/com.hazelcast.jet/hazelcast-jet-pulsar/{os-version}/jar[`hazelcast-jet-pulsar` module] to your member's classpath.
+
+The Pulsar connector is compatible with Pulsar version equal to
```
JamesHazelcast: PRD R10 requires a Pulsar/Spring Boot/Java compatibility matrix. `>= 4.0.0` omits Spring Boot and Java and claims support beyond the broker versions exercised here (`4.2.3` and `5.0.0-M1`). Replace it with the supported and tested matrix.

docs/modules/data-connections/pages/pulsar-dc-configuration.adoc
JamesHazelcast: The second YAML example and both SQL examples use Mongo connections - both SQL blocks are also missing commas between `OPTIONS` entries.

docs/modules/integrate/pages/pulsar-connector.adoc
JamesHazelcast, line 27: `Pipeline.reate()` does not compile - neither builder sets the required `connectionSupplier` or `dataConnectionRef`, and the sink ends with a stray backtick.

docs/modules/sql/pages/mapping-to-pulsar.adoc
JamesHazelcast, line 48: This `CREATE MAPPING` omits `valueFormat`, so `KvMetadataResolvers` rejects it

docs/modules/sql/pages/mapping-to-pulsar.adoc
JamesHazelcast: This example omits `DATA CONNECTION` and uses Kafka's `bootstrap.servers`, which Pulsar ignores - the first read or write reaches `Utils.getClient()` with neither a supplier nor `DataConnectionRef` and fails.

docs/modules/sql/pages/mapping-to-pulsar.adoc
JamesHazelcast, line 194: `valueFormat='java'` cannot handle `com.example.Person` - `PulsarTable.resolveSchema()` only supports the built-in types and throws `UnsupportedOperationException` for this class. The SQL also misses its closing `);`

docs/modules/sql/pages/mapping-to-pulsar.adoc
JamesHazelcast, line 53: `PulsarTable` never reads `preferredLocalParallelism` - it always returns `LOCAL_PARALLELISM_USE_DEFAULT`, so the documented value `2` has no effect. The partition guidance is also Kafka-specific - this connector uses a Shared subscription rather than one-to-one partition assignment. We need to remove this option and guidance or implement them for Pulsar

docs/modules/integrate/pages/connectors.adoc
JamesHazelcast: The PRD specifies an exactly-once reader, and `PulsarReaderBuilder` still describes itself as fault tolerant and snapshots/restores message IDs - `N/A` contradicts both.

docs/modules/integrate/pages/pulsar-connector.adoc
```diff
@@ -1,3 +1,83 @@
...
+the records only after they are committed.
+
+If you use at-least-once guarantee, records are visible immediately, but
+in the case of a failure some records could be duplicated. You
+can also configure jobs in exactly-once mode and decrease the guarantee
+just for a particular Pulsar sink.
+
+== Version Compatibility
+
+The Pulsar sink and source are based on version 5.x while keeping support of 4.x.
```
JamesHazelcast: PRD R10 requires a Pulsar server/Spring Boot/Java compatibility matrix, but this only says `5.x` supports `4.x` and `mapping-to-pulsar.adoc` claims every version from `4.0.0` onwards. Neither identifies client versus server or gives the supported Spring Boot and Java versions.

docs/modules/integrate/pages/connectors.adoc
```diff
@@ -32,6 +32,12 @@ You can use the following connectors in SQL by configuring them in the xref:sql:
 |streaming
 |exactly-once
 
+
+|xref:sql:mapping-to-pulsar.adoc[Apache Pulsar]
```
JamesHazelcast, line 36: `mapping-to-pulsar.adoc` is missing from `ROOT/nav.adoc`, while `querying-streams.adoc` still says Kafka is the only SQL streaming source - the only links to the new page are in this connector table.

docs/modules/sql/pages/mapping-to-pulsar.adoc
TomaszGaweda: not sure what you mean by first part. There are tests for that - pulsarAvro is in fact "java" : https://github.com/TomaszGaweda/hazelcast-mono/blob/ed82a215237d14023c96cec39838f56631edbaca/hazelcast/extensions/pulsar/src/test/java/com/hazelcast/jet/pulsar/sql/SqlAvroTest.java#L59

docs/modules/sql/pages/mapping-to-pulsar.adoc
TomaszGaweda: at never reading preferredLocalParallelism - gotcha, bug.

at guidance being Kafka specific - no, it's not, Pulsar just moves assignment to the broker, but incorrect number may result in some processors getting more work

docs/modules/integrate/pages/connectors.adoc
TomaszGaweda: true, some brain fart

docs/modules/integrate/pages/pulsar-connector.adoc
TomaszGaweda: I mean, it compiles, but fails at runtime ;p fixed

docs/modules/data-connections/pages/pulsar-dc-configuration.adoc
TomaszGaweda: 🙈

docs/modules/data-connections/pages/pulsar-dc-configuration.adoc
```diff
@@ -0,0 +1,168 @@
...
+      <property name="tenant">myTenant</property> <3>
+      <property name="namespace">myNamespace</property> <3>
+      <property name="enableTransactions">true</property> <4>
+    </properties>
+    <shared>true</shared>
+  </data-connection>
+</hazelcast>
+----
+<1> (Required) Broker URL, mandatory property.
+<2> (Optional) Broker HTTP Service - necessary for resource listing.
```
JamesHazelcast, line 34: `tenant` and `namespace` only filter `listResources()` - they do not restrict which topics the returned client can read or write. A fully qualified topic outside that scope is still accessible if the broker allows it.

docs/modules/sql/pages/mapping-to-pulsar.adoc
```diff
@@ -0,0 +1,217 @@
...
+
+For keys and values that are primitives, set the format as the SQL data type that corresponds to the primitive. For example, if the message key is an integer and the value is a
+string, use the `int` and `varchar` formats:
+
+[source,sql]
+----
+CREATE MAPPING my_topic
+DATA CONNECTION myPulsar
+OPTIONS (
+    'keyFormat'='int',
```
JamesHazelcast, line 89: An insert with an integer key fails with `ClassCastException` - `WriteRowProcessorSupplier` passes the projected key straight to `WritePulsarP`, which expects a `String`. The `bigint` keys in the JSON examples have the same problem. Can we use `varchar` keys until the connector supports numeric keys?

docs/modules/pipelines/pages/pulsar.adoc
```diff
@@ -256,16 +240,16 @@ public class JetJob {
     public static void main(String[] args) {
         String topicName = "hz-jet-topic";
 
-        StreamSource<Event> source = PulsarSources.pulsarReaderBuilder(
-                topicName,
-                () -> PulsarClient.builder().serviceUrl("pulsar://localhost:6650").build(),
-                () -> Schema.JSON(Event.class),
-                Message::getValue).build();
+        StreamSource<Event> source = PulsarSources.pulsarConsumerBuilder(json(Event.class))
```
JamesHazelcast, line 243: Switching to the consumer changes the restart behaviour described below - the current consumer acks messages even with snapshots disabled and reconnects to the same subscription. Restarting does not replay acknowledged history - `Earliest` only sets the position for a new subscription.

docs/modules/integrate/pages/connectors.adoc
TomaszGaweda: 5d09335475b295718e7738c927e1620454d83708

docs/modules/sql/pages/mapping-to-pulsar.adoc
TomaszGaweda: 008c6adad8b51e211287404d7ac6b9c9d49f5ca6

docs/modules/sql/pages/mapping-to-pulsar.adoc
TomaszGaweda: 4d43e062138cad2602c64f72718b502b2b1a31c4

docs/modules/data-connections/pages/pulsar-dc-configuration.adoc
TomaszGaweda: true, ef2f8abf42d7d37c7d221dfbc6bb809cea60cfaa

docs/modules/pipelines/pages/pulsar.adoc
TomaszGaweda: What changes? I don't see anything in the text below that changes, maybe I am overlooking something obvious after staring at it for too long :D

mscwilson (approved):

PR #2304, opened 2026-09-08
Remove outdated information that Solaris is supported

k-jamroz:
Makes performance tips page consistent with https://docs.hazelcast.com/hazelcast/5.7/deploy/versioning-compatibility which does not state that SPARC/Solaris is supported since at least 5.3.


JamesHazelcast (approved):

Patras3 (approved):
Thanks!

mscwilson (approved):

